PR #175, opened 2019-05-30
Feature/psq dascale plotting enhancements

t-b:
Created from: http://stash.corp.alleninstitute.org/projects/ENG/repos/mies-igor/pull-requests/1083/overview


t-b:
- [ ]  frequency calculation seems to be off (should be 10hz):

grafik

t-b:
@timjarsky Regarding the spike frequencies, I'm using the number of spikes *found*, see https://github.com/AllenInstitute/MIES/blob/feature/psq-dascale-plotting-enhancements/Packages/MIES/MIES_AnalysisFunctions_PatchSeq.ipf#L1304, whereas you say I should use the number of spikes in the DA wave from the wavebuilder? If yes, how should I handle the case of multiple epochs with different number of spikes?

timjarsky:
>.... whereas you say I should use the number of spikes in the DA wave from the wavebuilder? If yes, how should I handle the case of multiple epochs with different number of spikes?

@t-b No, it is just that in this case they should be equal.

timjarsky:
@t-b spike frequency calculation fails in the test case where I use a pulse train to simulate spikes. This is in part because I think it uses the pulse duration, rather than the epoch duration to calculate frequency. I also noticed that`spikeDetection` has a minimum of 1.  Just want to discuss with you first to make sure that this is limited to my test config.

t-b:
@t-b: Does it handle the case with no spikes at all properly?

timjarsky:
> @t-b: Does it handle the case with no spikes at all properly?

@t-b No; it reliably returns a non zero frequency value.

t-b:
@timjarsky The fix was a bit complicated. But it's fixed now.

t-b:
- [ ] polish local patch with search range for spikes

t-b:
@timjarsky Spike searching is now done only in the range of the pulse and the frequency is rounded.

timjarsky:
@t-b looking good. Last remaining item is the resolution of the spike frequency:

image

t-b:
@timjarsky Next try with the frequency rounding.

timjarsky:
@t-b 

has the rounding issue become an axis range issue?

Also, why is the fit off?

image

t-b:
@timjarsky Can you upload the file?

t-b:
@timjarsky I've pushed some changes. But I think they are not really related. What you are seeing are most likely floating point rounding issues. Both `a` and `b` are close to the double precision of `1e-15`. And the slope output is not wrong with `258a%` as that means `258` atto (10^-18).

Do you expect spike frequencies which are flat like that?

PR #177, opened 2019-05-30
Create IVSCC standalone panel and remove tango

t-b:


t-b:
@bergjim Yeahh that is a biggish PR in terms of removal.

For testing you cang et the installer from the "All Checks have passed"->"
Compilation of release package", i.e. http://bamboo.corp.alleninstitute.org/browse/MIES-CG5-CRP-2/artifact. And then you will have the new IVSCC panel available in "MIES Panels"->"Automation"->"IVSCC control panel".

t-b:
@bergjim No this PR still works on IP7 and 8.

t-b:
Requested changes:

1) For the baseline QC, the current sequence is for the QC check to run first, then the pipette offset. The pipette offset should always run first (such that the QC check after the offset is applied should almost always pass).
2) The order of the 2nd and 3rd buttons is reversed. The gigaohm seal is the second protocol in the sequence, with the access check being the last.

t-b:
> The order of the 2nd and 3rd buttons is reversed. The gigaohm seal is the second protocol in the sequence, with the access check being the last.

Fixed.

t-b:
@bergjim 
> For the baseline QC, the current sequence is for the QC check to run first, then the pipette offset. The pipette offset should always run first (such that the QC check after the offset is applied should almost always pass).

I'm not 100% sure I understand the problem, but I've fixed two related bugs which could explain the behaviour.

Another issue is how test pulse counting is done in general. Tim and I think that the current approach is not optimal. A better approach would be to count the number of TPs after the background functions ran the first time. @timjarsky has the details and offered to chat with @bergjim.

t-b:
@bergjim Rebased onto latest master. Now requires IP8.

t-b:
@bergjim @timjarsky I'm merging this PR right now in order to keep maintenance low.

@bergjim I'm happy to make any changes you need in a new PR then.

PR #178, opened 2019-05-30
Add epoch information

t-b:
@MichaelHuth:
- [x] Add amplitude (DAScale * amplitude) for TP and stimset epochs, check in tests as well
- [x] Sort by start time after finishing writing it
- [x] Remove "Sweep=.*;" prefix from epoch strings
- [x] Write about epoch labels in the documentation

Close #178.


t-b:
Comment from TimJarsky:

I see overlapping epochs and the ordering doesn't appear to follow the timing of epoch delivery

grafik

t-b:
@MichaelHuth Rebased to incorporate the sync-with-private-repo branch correctly.

MichaelHuth:
I have added now an Amplitude numeric key into the information for epochs and TP. The amplitude is DAScale * amplitude and TestpulseAmplitude * amplitude respectively, without gain. It equals the epochs maximum values in the corresponding saved sweep wave as on saving the DA channels are divided by gain again (at least for NI).

t-b:
@MichaelHuth 

```
Errors building the documentation
Doxygen says:
/home/thomasb/bamboo-agent-home/xml-data/build-dir/MIES-CG7-DOC/Packages/MIES/MIES_DataConfiguratorITC.ipf:1890: warning: The following parameters of DC_AddEpochsFromTP(string panelTitle, variable channel, variable baselinefrac, variable testPulseLength, variable offset, string name, variable amplitude) are not documented:
  parameter 'amplitude'
/home/thomasb/bamboo-agent-home/xml-data/build-dir/MIES-CG7-DOC/Packages/MIES/MIES_DataConfiguratorITC.ipf:1952: warning: The following parameters of DC_AddEpochsFromStimSetNote(string panelTitle, variable channel, wave stimset, variable stimsetBegin, variable setLength, variable sweep, variable scale) are not documented:
  parameter 'scale'
```

timjarsky:
@t-b It looks like inserted TP epoch info is being sorted separately: 

image

t-b:
@timjarsky Good point.

@MichaelHuth Looks like the sorting in https://github.com/AllenInstitute/MIES/pull/178/commits/bdb24c014b83f8c777bcd8cf9212ebb8bd2ba4b5#diff-8311da9a9a5675cb652187f91a7ebecdR1183 line  1183 does not do the right thing. Why not just sort the first column of the epochChannel wave?

MichaelHuth:
@timjarsky @t-b I can not reproduce that from data out of the LNB. From the wave name `EpochsWave` in the screenshot it looks like it was taken before the sorting happens.

Though, I will change it.

t-b:
@MichaelHuth I think we want to have the `epcochsWave` sorted and not only the LBN entries.

MichaelHuth:
@t-b So the idea is that after each call to `DC_AddEpoch` each layer of `EpochsWave` is for itself in a sorted by the start times, such that EpochsWave is at any point sorted?

I think the current data layout with a 3D wave as `EpochsWave` is not very suited for column based row sorting in each layer independently. If that is the goal we should think about changing it to a wave reference wave that points to 2D Epochs waves for each channel.

t-b:
@MichaelHuth I'll think about that and reply.

MichaelHuth:
@t-b I committed a solution that keeps the 3D epoch wave.

t-b:
@MichaelHuth Sorting is correct. Nice!

One bug popped up, see [Experiment.zip](https://github.com/AllenInstitute/MIES/files/3298635/Experiment.zip). There we have a negative epoch end time. After fixing the bug please also ensure that we have assertions to catch these cases in the future.

MichaelHuth:
@t-b Ok this should fix now the bug that appeared with flipped distributed pulse trains in combination with oodDAQ.
The tests included already a dDAQ and oodDAQ test. However I added a oodDAQ test with a flipped distributed stimset that gets cut off in oodDAQ mode.

timjarsky:
@t-b nitpick alert...... Would it be possible to have the stimset epoch listed before the first epoch of the stimset? It is currently the other way around:

image

timjarsky:
> @t-b nitpick alert...... Would it be possible to have the stimset epoch listed before the first epoch of the stimset? It is currently the other way around:
> 
> image

@t-b I guess what I am really asking is that for epocha that start at the same time, that the one that finishes last be placed first.

MichaelHuth:
The reason is that if one adds a epoch that starts at the same time as a present epoch, then it is inserted before this entry.

t-b:
@timjarsky So basically:
Primary sorting key: - Ascending start time
Secondary sort key: - Descending stop time

MichaelHuth:
@t-b Ok, I change it back to how it was before, just with the secondary key and save it also back to the EpochsWave (not only to the LNB).

MichaelHuth:
@t-b I would like to note that if a stimset has only a single epoch then the stimset start and length equals the epoch length for uncut non-oodDAQ. Then the order whether the stimset or epoch is first is undefined with regular sorting.

MichaelHuth:
@t-b rebased on current master

MichaelHuth:
@t-b I added the tree level as tertiary sorting key (ascending). I added as well a OODDAQ_PRECISION constant in the tests according to the precision of the saved oodDAQ regions, currently 1 ms.
Please review if the sorting is now as intended.

t-b:
@MichaelHuth One test needs fixing.

```
Collapse 	Failed 	Epochs#EP_EpochTest5:ITC Epochs#EP_EpochTest5:ITC in UTF_Epochs.ipf (1) History
	8 secs

FAIL: 0: is false   Assertion "CHECK(endT[epochNr] <= endInt)" failed in line 184 (361), procedure "UTF_Epochs.ipf" 

```

MichaelHuth:
The tests run fine with NI hardware here. EpochTest5 is the unflipped oodDAQ poisson distribution. The failing CHECK could have several origins. It usually indicates that the Epochs might cover the whole DA wave range. We should have a look at this as soon as you are able to run this on ITC again.

MichaelHuth:
@timjarsky Yesterday I visited @t-b and we took the opportunity to check why the tests failed with ITC. After some digging it turned out that the default conversion from start/endtimes to strings in the epochsWave with num2str() induced a precision limit of 5 digits. This relates to 10 microseconds on a seconds scale. The ITC hardware test used a DA frequency that when rounding the epoch borders to 5 digits led to non-perfectly consecutive epochs.
We increased now the precision to 7 digits (aka 0.1 microseconds). This allows precise epoch times up to 10 MHz DA rate.

t-b:
I found a bug with flipped stimsets. 
[Experiment-epochs-bug.zip](https://github.com/AllenInstitute/MIES/files/3326625/Experiment-epochs-bug.zip)

The pulse starting times are wrong. Is it easily possible to compare the pulse starting times in the tests with that FindLevel returns?

I've used

```
Function SetCursors(panelTitle, trace, DAC, epoch)
	string panelTitle, trace
	variable epoch, DAC

	variable first, last
	WAVE/T epochsWave = GetEpochsWave(panelTitle)

	string win = "DB_ITC18USB_Dev_0"

	first = str2num(epochsWave[epoch][%StartTime][DAC]) * 1E3
	last  = str2num(epochsWave[epoch][%EndTime][DAC]) * 1E3

	Cursor/W=$win/A=1 A $trace first
	Cursor/W=$win/A=1 B $trace last
End
```

for visualization of the epochs.

MichaelHuth:
On a brief phone conference with Thomas we agreed to change the notation of the flipped pulse trains in the epochs that way that the pulse itself is first in each level 2 epoch. Currently due to the flipping the level 2 pulse time consists of baseline + pulse. For the later pulse averaging this seems not suitable.

t-b:
@MichaelHuth Output of hardware tests: 
[JU_HardwareTests_2019-06-27_12-00-04.zip](https://github.com/AllenInstitute/MIES/files/3337923/JU_HardwareTests_2019-06-27_12-00-04.zip)

t-b:
- I've renamed "baseline" to "Baseline" and pushed a fixup commit
- The pulse length for the last pulse with poisson and flipped is wrong I think. I would expect it to include the trailing baseline of the epoch as well. See 
grafik

Packages/doc/epoch_information.rst
```diff
@@ -0,0 +1,119 @@
+.. _epoch_information_doc:
+
+=================
+Epoch Information
+=================
+
+Description
+-----------
+
+Whenever a stimset is used to acquire data MIES generates also information about epochs.
```
timjarsky: How about: At acquisition time, sweep epoch metadata is generated.

MichaelHuth: I changed it.

t-b:
@timjarsky Documentation looks good. 

@MichaelHuth 
I've fixed one wrong level range and also changed to our convention where `DA` is the *input* signal.
I've also readded labnotebook version increment and sweep settings increment. These were probably lost in the rebase.

Please rebase and push if the fixups are good.

MichaelHuth:
@t-b ok I missed that. It is quite unusual that the sample is the reference point for the signal handling.

t-b:
@timjarsky Looks good from my POV.

Issue #186, opened 2019-06-03
Installer: Does not warn about all user manual installation in C:\Program Files\WaveMetrics\...


t-b:
IIRC we warn the user when a manuall installation is detected. But that seems to not include all user manual installations into `%ProgramFiles%` or?

Bug report from @timjarsky:
> Here is what happened: MIES was previously manually installed (by an administrator) for 
> all users. Then the installer was used to install for a user/account that did not have admin 
> privileges. The installer did not tell the user to first remove the manually installed copy of 
> MIES. (bearbeitet)

@MichaelHuth Is that correct?

MichaelHuth:
I have looked into the installer code and currently different privileges are not mixed. 

So if you install as user (specifically this means 'not admin' !) then it checks if a uninstaller from a previous user mode installation of this user is present.

If you install as admin then it checks in the registry ifthe mies software package registered an uninstaller.


Generally a user can not install into %ProgramFiles% unless he has elevated rights.

What probably can be done is:

If a user installs I can try to check if an admin installation is present. If the user has the rights to read the appropriate registry keys. In non-silent mode I can call the uninstaller which most likely will popup an privilege elevation request. Another option would be to show a message box and abort installation.

In silent mode I abort installation silently.

If an admin installs, I can not see if one or more of the users might have MIES installed as MIES and the shortcuts to include everything in Igor Pro is all located inside the users personal folder.

MichaelHuth:
@t-b @timjarsky See https://github.com/AllenInstitute/MIES/pull/191

In my tests it showed that there is no default functionality to popup UAC elevation request automatically if a user tries to run the uninstaller that requires admin rights.

MichaelHuth:
I am still not 100% sure how the originally described error can come up. As thomas wrote, we have a second check before we start copying files. If in the target Igor procedures folder a file links e.g. to MIES_Include.ipf then we abort the installation with a message that there is a manual installation that has to be removed first. This check is done for the global Igor path and the local user folder for Igor Procedures. 

If there was a manual installed copy at the targetted Igor Pro the abort should have come up.

I will change the uninstall logic by the installer now to a try + recheck approach:

- The installer will register MIES in Add/Remove Programs depending on shell context now in HKLM or HKCU. The uninstaller is created with the same shell context.

- On startup the installer will check in HKLM and HKCU for uninstall information (and for compatibility in default user folder)

- The installer will call found uninstallers regardless of context.

[...]

PR #191, opened 2019-06-05
fix for uninstallation of admin installation as user.

MichaelHuth:
should fix: https://github.com/AllenInstitute/MIES/issues/186

Previously the uninstallation check onyl checked vs same privilege
installation. 
Now always user and admin installations are checked, which means that in the worst case the uninstaller is run twice (for user and admin).

If an admin installation is present and a user tries to install, he gets
a suitable message prior aborting installation.

In silent mode the installation is silently aborted in that case.

As already before: If an admin installs for All Users it is NOT checked if any user on the system has an active user installation of MIES.


timjarsky:
@MichaelHuth Is there a windows folder that is accessible to all users? Can we create one? If yes, could we place a file in the folder that records the type of MIES installation?

t-b:
@timjarsky The "windows way" of doing that kind of thing would be do write something in the registry and then the user could see that e.g. in `SystemControl/Programs/Add/Remove programs`. Does that sound useful?

timjarsky:
@t-b Yes; I guess I might have misunderstood how registry access limitations @MichaelHuth mentioned[1] were going to limit our ability to detect all installation types.

[1] If a user installs I can try to check if an admin installation is present. If the user has the rights to read the appropriate registry keys.

t-b:
@timjarsky Good point. Well investigate the registry key access conditions and fallback to a file solution then.

t-b:
ShellLink plugin: https://nsis.sourceforge.io/ShellLink_plug-in

MichaelHuth:
I have patched the Shelllink plugin used by NSIS to allow retrieving link targets from admin written lnk files.

timjarsky:
@t-b How does one go about getting the installer for this (or any) PR?

t-b:
@timjarsky 

Scroll down to the "Checks" element. And then you see

grafik

and there as soon as "Compilation of release package" has a details element, this will bring you to bamboo which will have the usual artefacts.

PR #199, opened 2019-06-11
Bugfix: Use floating point conversion %g for building the region list

t-b:
There is no reason why we only want integers.

Broken since d102c07d (Add new data acquisition mode: Optimized overlap
distributed acquisition, 2016-08-10).

The oodDAQ regression test data was also regenerated.

Bug pointed out by Michael Huth.


t-b:
@MichaelHuth Merged.

PR #214, opened 2019-06-21
fix for cases where DA_Ephys panel is not in sync with GUIState waves

MichaelHuth:
There still existed a few situations where it happens that existing
GUIState waves are not in sync with the available controls of the
DA_Ephys panel. e.g. if a dev has changed controls without increasing
the PANEL VERSION constant. Or if an experiment was loaded with an open
panel with a different control state.

The fix is to reset the GUIState waves if
- The panel is recreated after an experiment load with the name
  BASE_WINDOW_TITLE. The reset is done in AfterFileOpenHook.
- The panel is locked as this changes the panel title and thus the
  the path to the GUIState waves
- The panel is unlocked as this changes the panel title and thus the
  the path to the GUIState waves
- The panel is created

### DIFF
--- a/Packages/MIES/MIES_DAEphys.ipf
+++ b/Packages/MIES/MIES_DAEphys.ipf
@@ -4539,6 +4539,8 @@ Function DAP_LockDevice(panelTitle)
 
 	DoWindow/W=$panelTitle/C $panelTitleLocked
 
+	KillOrMoveToTrash(wv = GetDA_EphysGuiStateNum(panelTitleLocked))
+	KillOrMoveToTrash(wv = GetDA_EphysGuiStateTxT(panelTitleLocked))
 	// initial fill of the GUI state wave
 	// all other changes are propagated immediately to the GUI state waves
 	DAG_RecordGuiStateNum(panelTitleLocked)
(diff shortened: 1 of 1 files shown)


Packages/MIES/MIES_IgorHooks.ipf
```diff
@@ -191,3 +191,15 @@ static Function AfterCompiledHook()
 End
 
 #endif
+
```
timjarsky: @MichaelHuth Please document this function.

Packages/MIES/MIES_IgorHooks.ipf
```diff
@@ -191,3 +191,26 @@ static Function AfterCompiledHook()
...
+	variable refNum
+	string fileNameStr, pathNameStr, fileTypeStr, fileCreatorStr
+	variable fileKind
+
+	string list = WinList(BASE_WINDOW_TITLE, ";", "WIN:64")
+	if(!isEmpty(list))
+		KillOrMoveToTrash(wv = GetDA_EphysGuiStateNum(BASE_WINDOW_TITLE))
+		KillOrMoveToTrash(wv = GetDA_EphysGuiStateTxT(BASE_WINDOW_TITLE))
+	endif
+End
```
t-b: @MichaelHuth I think this is not required. We call `DAP_UnlockDevice` for all locked devices via `IH_Cleanup`, and that is called before closing and experiment. This should catch the same case you want to catch here or?

MichaelHuth: @t-b I don't agree, consider this case: 

Dev1 is locked, you save the experiment and reload it, what happens is:
- IH_Cleanup unlocks "Dev1", thus deletes the GUIState waves for Dev1.
- On loading the saved experiment "DA_Ephys" is created from Macro and these GUIState waves were not deleted on Unlock.

We could delete the GUIState waves as well in `DAP_UnlockDevice` for `panelTitleUnlocked` which is `BASE_WINDOW_TITLE`. This would work in the way you propose. What do you think?

t-b: @MichaelHuth 

> We could delete the GUIState waves as well in DAP_UnlockDevice for panelTitleUnlocked which is BASE_WINDOW_TITLE. This would work in the way you propose. What do you think?

Sounds good.

PR #219, opened 2019-06-25
Utility function num2strHighPrec to extend num2str with defined precision

MichaelHuth:
The num2str standard function uses a precision of 5 digits by default.
In some cases this is not sufficient. num2strHighPrec solves this by adding an
optional precision parameter to be able to use a defined number of digits.

- also added tests to UTF_Utils.ipf for num2strHighPrec.

### DIFF
--- a/Packages/Testing-MIES/UTF_Utils.ipf
+++ b/Packages/Testing-MIES/UTF_Utils.ipf
@@ -2282,3 +2282,103 @@ Function TextWaveToListWorks5()
 	CHECK_EQUAL_STR(list, refList)
 End
 /// @}
+
+/// num2strHighPrec
+/// @{
+
+/// @brief Fail due to negative precision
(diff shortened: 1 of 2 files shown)


Packages/Testing-MIES/UTF_Utils.ipf
```diff
@@ -2282,3 +2282,103 @@ Function TextWaveToListWorks5()
 	CHECK_EQUAL_STR(list, refList)
 End
 /// @}
+
```
timjarsky: Is an 8:1 test function to "feature" function typical?

t-b: @timjarsky I'm not sure "typical" is a good word for it. I would say though that it does test the function behaviour completely enough. It is good practive to have short test cases, so that you know when a test fails what the problem is.

PR #220, opened 2019-06-25
Add utility function ListToTextWaveMD

MichaelHuth:
This utility function converts a list with different separators to a multi dimensional text wave using row major order. As the list length can be dynamic the text wave will be filled sparse from the top. Remaining text wave elements contain empty strings.

- added tests for ListToTextWaveMD to UTF_Utils.ipf


t-b:
@MichaelHuth rebased to solve conflicts.

PR #223, opened 2019-06-25
Change location/timing of POST_DAQ_EVENT call to later in DAP_OneTime…

timjarsky:
Old behavior: Analysis function POST_DAQ_EVENT called after setting the dataAcqRunMode 

New behavior: Analysis function POST_DAQ_EVENT immediately after TP_teardown for TP during DAQ

### DIFF
--- a/Packages/MIES/MIES_DAEphys.ipf
+++ b/Packages/MIES/MIES_DAEphys.ipf
@@ -1296,12 +1296,12 @@ Function DAP_OneTimeCallAfterDAQ(panelTitle, [forcedStop, startTPAfterDAQ])
 
 	DAP_ResetGUIAfterDAQ(panelTitle)
 
+	NVAR dataAcqRunMode = $GetDataAcqRunMode(panelTitle)
+	dataAcqRunMode = DAQ_NOT_RUNNING
+
 	if(!forcedStop)
 		AFM_CallAnalysisFunctions(panelTitle, POST_DAQ_EVENT)
 	endif
-
-	NVAR dataAcqRunMode = $GetDataAcqRunMode(panelTitle)
-	dataAcqRunMode = DAQ_NOT_RUNNING
 
 #if (IgorVersion() < 8.00)
 	StopAsyncIfDone()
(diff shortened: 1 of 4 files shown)


Packages/MIES/MIES_DAEphys.ipf
```diff
@@ -1347,6 +1343,10 @@ Function DAP_OneTimeCallAfterDAQ(panelTitle, [forcedStop, startTPAfterDAQ])
 		TP_TeardownCommon(panelTitle)
 	endif
 
+	if(!forcedStop)
+		AFM_CallAnalysisFunctions(panelTitle, POST_DAQ_EVENT)
+	endif
+
 	DAP_ApplyDelayedClampModeChange(panelTitle)
```
t-b: @timjarsky Can we move it even further after `DAP_ApplyDelayedClampModeChange(panelTitle)`?

t-b:
@timjarsky How about the following documentation change:

```diff
diff --git a/Packages/MIES/MIES_AnalysisFunctions.ipf b/Packages/MIES/MIES_AnalysisFunctions.ipf
index 47258f06..6bc6936c 100644
--- a/Packages/MIES/MIES_AnalysisFunctions.ipf
+++ b/Packages/MIES/MIES_AnalysisFunctions.ipf
@@ -35,7 +35,8 @@
 /// Post Sweep   After each sweep (before possible ITI pause)   None
 /// Post Set     After a *full* set has been acquired           This event is not always reached as the user might not acquire all steps
 ///                                                             of a set or indexing on multiple headstages is used.
-/// Post DAQ     After all DAQ has been finished                None
+/// Post DAQ     After all DAQ has finished and                 None
+///              immediately before a potential "TP after DAQ"
 /// =========== ============================================== ===============================================================
 ///
 /// \endrst
```

Packages/MIES/MIES_DAEphys.ipf
t-b: @timjarsky Ditch that. This does not make sense.

t-b:
@timjarsky The failing tests are due to the fact that we first reset the stimset to the original ones if indexing is used and then call the POST_DAQ_EVENT. And this is then done for the wrong stimset.

I think we need to do a bit more here, as the "setting stimsets back before indexing" collides with your wish of "changing the stimset in POST_DAQ_EVENT". I'll think about something.

t-b:
- [ ] Document that we don't allow changing the stimset with indexing in post daq
- [ ] Move post-daq-event call after `DAQ_NOT_RUNNING` setting and before indexing reset

t-b:
@timjarsky Ready for testing.

PR #224, opened 2019-06-25
Bugfix/search string breakage

t-b:


timjarsky:
@t-b 

(1) The search string in All, V-Clamp, and I-Clamp isn't being applied to the Indexing End Set.
(2) Indexing (with repeat sets) is not indexing at the correct sweep. I can demo.

timjarsky:
> @t-b
> 
> (1) The search string in All, V-Clamp, and I-Clamp isn't being applied to the Indexing End Set.
> (2) Indexing (with repeat sets) is not indexing at the correct sweep. I can demo.

@t-b ignore (1); user error.

timjarsky:
> > @t-b

> > (2) Indexing (with repeat sets) is not indexing at the correct sweep. I can demo.

Actually, it repeats the sets after completing an indexing cycle. This was likely a design choice I may want to change.

t-b:
The search string not being applied to the all controls is fixed. 

And yes the indexing is tied to the locked/unlocked logic. The cycles are shown at https://alleninstitute.github.io/MIES/.

And of course I'm happy to implement a "semi"-locked indexing mode.

timjarsky:
> The search string not being applied to the all controls is fixed.
> 
> And yes the indexing is tied to the locked/unlocked logic. The cycles are shown at https://alleninstitute.github.io/MIES/.
> 
> And of course I'm happy to implement a "semi"-locked indexing mode.

@t-b That's okay; we'll increase the number of sweeps in the stimset instead.

PR #231, opened 2019-06-28
SetControlInEvent: Make it possible to set the same control in multiple events

### LINKED ISSUE #226: Ehance SetControlInEvent so that the same control can be set during different analysis function events
e.g.,
Check_TTL_01 Pre DAQ|1|
Check_TTL_01 Post DAQ|0|

t-b:
We now allow the textwave to have multiple tuples of event/data. This
approach is backwards compatible.

Close #226.

### DIFF
--- a/Packages/MIES/MIES_AnalysisFunctions.ipf
+++ b/Packages/MIES/MIES_AnalysisFunctions.ipf
@@ -1022,8 +1022,9 @@ End
 ///
 /// Usage:
 /// - Add analysis parameters named like the control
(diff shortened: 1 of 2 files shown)


timjarsky:
@t-b I'm finding that the parameters panel always overwrites instead of adding a control of the same name.

t-b:
@timjarsky The parameter panel always assumes that you want to write a full wave. So there is currently no concept of just adding new wave entries.

PR #233, opened 2019-06-28
Feature/speedup hardware tests

### LINKED ISSUE #171: Speedup Hardware Tests by not running into the timeout for TP tests
Make it an error if we run into the timeout. See also https://github.com/byte-physics/igor-unit-testing-framework/issues/88.

t-b:
Close #171.


timjarsky:
@t-b 

Wow:
image

t-b:
@timj These tests were always fast.

"Testing with hardware" went down from 82min to 35 min.

PR #234, opened 2019-07-03
IsValidSweepAndConfig: Only require version 0 config waves in special…

t-b:
… cases

In 9caf4680 (Added test for version 2 of ITCChanConfigWave, updated
IsLatestConfigWaveVersion, 2018-12-21) IsValidConfigWave learned to
check for a specific config wave version using the version 2 as default.

But it was forgotten to be able to allow passing in the minimum required config
version into IsValidSweepAndConfig.

This is required in places like the DataBrowser/SweepBrowser and
functions used from there where we want to be able to browse very old
data as well.


timjarsky:
@t-b Sweeps can now be loaded into the dataBrowser, but when selecting overlay sweeps I get:

image

t-b:
@timjarsky Thanks. Gnahh! Fixed!!

PR #236, opened 2019-07-05
Feature/search string in wavebuilder

### LINKED ISSUE #213: Add search string to loading controls in waveBuilder
Feature: add a setVar control used for searching, like in the DA_ephys panel, DA tab for example [1], to the waveBuilder load controls [2].

[1] 
image

[2]
image

t-b:
Close #213.


timjarsky:
@t-b Let's call it `filter` instead of `string'. Drop the `string'. and increase the width of the control.

t-b:
@timjarsky  I've used the same names as in the DAEphys panel. I would propose to have the same names everywhere. Should I rename the DAEphys ones to `filter` as well?

timjarsky:
> @timjarsky I've used the same names as in the DAEphys panel. I would propose to have the same names everywhere. Should I rename the DAEphys ones to `filter` as well?

@t-b ah, okay. I still think `filter` is the right choice (for both). You can rename the DAEphys ones now or when we are doing panel updates at some other time.

t-b:
Will do it now.

t-b:
@timjarsky Done.

PR #243, opened 2019-07-18
Scale ttl waves properly

### LINKED ISSUE #241: Scale splitted TTL waves from ITC hardware properly
- [ ] User expects a TTL wave to be from 0 to 1, rescale it as the `2^` logic is only required due to hardware specialities of the ITC hardware
- [ ] Teach DB_SplitSweepsIfReq that we now have new splitted TTL waves

t-b:
Close #241.


t-b:
@timjarsky I've not included any logic for updating existing splitted TTLs waves. Should I? This would be a bit more involved.

Another potential problem is that this changes how we store the data into NWB as well. I'm just mentioning that so that we are aware of it.

timjarsky:
> @timjarsky I've not included any logic for updating existing splitted TTLs waves. Should I? This would be a bit more involved.


No; at this point backwards compatibility would not be beneficial as we are still in the piloting stage of experiments.

t-b:
@timjarsky Done. I've not changed the NWB export, only the display.

PR #256, opened 2019-08-31
Speedup some functions which are called often during DAQ/TP

t-b:
- [x] cleanup changes
- [x] Store the result of  `TP_GetTestPulseLengthInPoints` for both TP and DAQ flags in the TP/DAQ folders and add global var getters so that we can query that really cheap during DAQ

We do call TP_GetTestPulseLengthInPoints for every oscilloscope update when "TP during DAQ channels" are on.

### DIFF
--- a/Packages/MIES/MIES_Cache.ipf
+++ b/Packages/MIES/MIES_Cache.ipf
@@ -206,6 +206,27 @@ static Function/S CA_WaveCRCs(waveRefs, [crcMode])
 	return NumericWaveToList(crc, ";", format = "%d")
 End
 
+/// @brief Calculate the cache key for SI_FindMatchingTableEntry.
+///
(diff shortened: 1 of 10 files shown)


t-b:
@timjarsky Does 4c09d91bd86f09cf3bb8f7df927bc2be60bdcd28 look useful enough to have it?

timjarsky:
> @timjarsky Does [4c09d91](https://github.com/AllenInstitute/MIES/commit/4c09d91bd86f09cf3bb8f7df927bc2be60bdcd28) look useful enough to have it?

@t-b it does. 

I tested on my set up and found no issues. Is there a particular test you had in mind? I'll approve now and you can merge if there isn't a specific test.

t-b:
@timjarsky Nope, nothing specific to test.

PR #279, opened 2019-09-21
Restore always all cursors

### LINKED ISSUE #277: Databrowser/Sweepbrowser: Restore all available cursors
We only restore cursors A and B, but we now want to restore all cursors.

t-b:
Since forever we only restored the cursors A and B.

For a future application we want to use all cursors, so it is quite
inconvenient that only A and B are restored.

Introduce GetCursorInfos/RestoreCursors and remove RestoreCursor to
handle that.

Close #277.


timjarsky:
@t-b it seems like cursor restoration works unless the trace with the cursors is removed. Is it possible to move the cursors onto a different trace, or?

t-b:
@timjarsky Yes I could take another trace as well. Can I do that in a seprate PR?

PR #292, opened 2019-09-24
Restructure documentation Building Process

t-b:
Use a log function during building process to differentiate between
verbose and non-verbose modes.

Let building process output `--version` information in debug mode.
Standard to debug mode.

all output to stdout.

Update docker
* use python3
* use current debian:stable which is buster.
* use pip3
* install breathe, sphinx via pip
* add latex support


t-b:
Will resurrect once we use docker for the documentation building. We can't do that right now as bamboo is too old to support that.

t-b:
Branch is postponed/enhance-documentation-building.

PR #293, opened 2019-09-24
Restructure Building Process

ukos-git:
Use a log function during building process to differentiate between
verbose and non-verbose modes.

Let building process output `--version` information in debug mode.
Standard to debug mode.

all output to stdout.

Update docker
* use python3
* use current debian:stable which is buster.
* use pip3
* install breathe, sphinx via pip
* add latex support


ukos-git:
duplicate of #292

PR #305, opened 2019-10-02
RunAllTests.pxp: Update Wavebuilder stimsets using FilterIIR

t-b:
The latest bugfix to FilterIIR changed the output in our regression
tests.

- [x] Wait for repsonse from WM wether this is expected or not.


t-b:
@timjarsky AFAIR we wanted to merge this PR. In case you disagree we can always revert it.

PR #306, opened 2019-10-02
Add notebook text wrappers

### LINKED ISSUE #298: PGC_SetAndActivateControl: Add support for Notebooks
- For now we need to replace all text only
- Add tests

t-b:
Close #298.

PGC_SetAndActivateControl is the wrong place to add it as that only works with real controls which have a GUI procedure attached.

I've added plain functions only therefore.

### DIFF
--- a/Packages/MIES/MIES_GuiUtilities.ipf
+++ b/Packages/MIES/MIES_GuiUtilities.ipf
@@ -1535,3 +1535,26 @@ Function GetControlCoordinates(win, ctrl, s)
 	s.left   = V_left
 	s.right  = V_left + V_width
 End
+
+/// @brief Get the text (plain or formatted) from the notebook
+Function/S GetNotebookText(win)
+	string win
+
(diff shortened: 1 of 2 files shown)


Packages/MIES/MIES_GuiUtilities.ipf
```diff
@@ -1535,3 +1535,26 @@ Function GetControlCoordinates(win, ctrl, s)
...
+/// @brief Get the text (plain or formatted) from the notebook
+Function/S GetNotebookText(win)
+	string win
+
+	ASSERT(WinType(win) == 5, "Passed win is not a notebook")
+
+	Notebook $win getData=1
+
+	return S_Value
+End
```
timjarsky: @t-b Do we not use multiple return syntax for functions that only return 1 object?

t-b: I've opened https://www.wavemetrics.com/forum/general/programming-question-regarding-multiple-return-value-syntax to track some of our earlier questions regarding MR functions.

PR #307, opened 2019-10-04
Feature/ovs allow clearing on new cycles

### LINKED ISSUE #264: Optionally clear data browser sweeps with new RA cycle and/or new stim set
In cases where the user is acquiring a single stimset at a time and using display last sweep acquired in combination with overlay sweeps, they often clear the data browser plot on each RA cycle manually by selecting the 'none'  stimulus set in overlay sweeps tab. Let's give the user the option to do this automatically on each new RA cycle.

### LINKED ISSUE #303: auto clear databrowser/sweepbrowser of overlayed sweeps
It's common for users to overlay all the sweeps from a stimulus set or an RA cycle during data acquisition. At the start of a new RA cycle, users have to select the `none` stimulus set to clear the overlayed sweeps. It would be helpful to have an option to autoclear previously overlayed sweeps when a new stimulus or RA cycle begins. I'm imagining two checkboxes in the OS panel of the sweep/data browser.

(1) Clear sweeps on new RA cycle
(2) Clear sweeps on new stimset (used when indexing through several stimulus sets)

t-b:
Close #303.
Close #264.

### DIFF
--- a/Packages/MIES/MIES_BrowserSettingsPanel.ipf
+++ b/Packages/MIES/MIES_BrowserSettingsPanel.ipf
@@ -475,7 +475,9 @@ End
 Function BSP_SetOVSControlStatus(win)
 	string win
 
-	string controlList = "group_properties_sweeps;popup_overlaySweeps_select;setvar_overlaySweeps_offset;setvar_overlaySweeps_step;check_overlaySweeps_disableHS;check_overlaySweeps_non_commula;list_of_ranges"
+	string controlList = "group_properties_sweeps;popup_overlaySweeps_select;setvar_overlaySweeps_offset;"            \
+						 + "setvar_overlaySweeps_step;check_overlaySweeps_disableHS;check_overlaySweeps_non_commula;" \
+						 + "list_of_ranges;check_ovs_clear_on_new_ra_cycle;check_ovs_clear_on_new_stimset_cycle"
 
 	BSP_SetControlStatus(win, controlList, OVS_IsActive(win))
 End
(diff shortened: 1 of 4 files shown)


timjarsky:
@t-b selecting the ITC1600 device type, from the popup menu, leaves the last ITC1600 attached locked.

t-b:
@timjarsky I fixed two related issues. Does that fix it already?

t-b:
@timjarsky Checkout 6a5fc460d078d6adeba6a9a4060b2ab5cf2ea7ea ;)

timjarsky:
@t-b functionality is already helpful. Minor (potential) logic issue: with RAC and SCI checked I was expecting clearing on SCI (and RAC) to occur. but instead, SCI is skipped/ignored. I can demo if this isn't what you were expecting and I apologize if what I'm saying here isn't what I said when we discussed the issue originally :/

t-b:
@timjarsky It is currently not clearing on both occasions see https://github.com/AllenInstitute/MIES/pull/307/files#diff-71126c076df2b66a0e0319da24298d0eR179. Your original design did not specify what to in that case so I made something up.

My thinking was that you are not able to trigger a new RA without also triggering a new SCI. Is that flawed?

timjarsky:
> My thinking was that you are not able to trigger a new RA without also triggering a new SCI. Is that flawed?

@t-b That's correct but it is possible to trigger a new SCI without a new RAC. If I'm reading the code correctly, line 182 should be an independent `If` statement.

Issue #297, opened 2019-09-25
DataBrowser/Sweepbrowser: Make the left external subwindow resizable


t-b:
This subwindow ("BrowserSettingsPanel") is currently fixed in size.

- Make it resizable
- Add minimum size
- Add ResizeControlsHook support and define userdata
- It might make also sense to rethink the organization of the four window macros (`DataBrowser()`, `BrowserSettingsPanel()`, `SweepControlPanel()`, `BrowserSettingsPanel`) . We can currently not just regenerate these as we do with `DAP_EphysPanelStartUpSettings` and that is quite inconvenient. Chat with Matthias why it is like that. Using only one macro makes it easier to regenerate. Add a menu entry for regeneration and function for setting the default values.

t-b:
- Keep simple controls attached to left upper corner.
- Resize inner e.g. edit boxes.
- Use for setup resize package, resize info is saved in user data, then recreate macro.

PR #245, opened 2019-07-24
Add SweepFormula

ukos-git:
Add formula serialization. The formula style follows [json logic](http://jsonlogic.com/)

closes #238

PR #315, opened 2019-10-09
Enable resizable DataBrowser BrowserSettings subwindow

MichaelHuth:
DataBrowser panel macro joined in single recreation macro
This allows to use the recreation workflow as with the DAEphys panel

Added Menu entry "Reset and store current DataBrowser panel", it works
the same way as the already present DAEphys panel recreation.

The SweepBrowser panel that used partially the same window creation macros
uses now fully the DataBrowser macro as well.

Further changes:
- Open functions for SweepBrowser and DataBrowser use both DataBrowser macro
- Init routines for DataBrowser SubWindows only setup dynamic properties.
  The SubWindow creation macros have been removed.
- The DataBrowser window creation macro was moved to an own ipf file:
  MIES_DataBrowser_Macro.ipf
- new: DB_ResetAndStoreCurrentDBPanel() utility function for generating
  DataBrowser recreation.
- All Dynamic property setup functions for SubWindows got an associated
  Unset function that is used to cleanup the GUI state prior generation.
- Previous Open functions for SubWindows were renamed to UnHide[SubWindowName]
  to reflect the simplified core functionality.
- DB_LockDevice renames the current Window only if the name actually changes.
- DB_CloseSettingsHistoryHook kills the SubWindow if the main panel is not of
  DataBrowser type (e.g. if it is SweepBrowser), previously it was always hidden
- Added Constants for SubWindow titles
- Added Constants for ControlLists for BrowserSettings and SweepControl Subwindow
- SweepBrowser Title is set when browser type is set.

Allow resizing of BrowserSettings and SweepHistory panel.

closes: https://github.com/AllenInstitute/MIES/issues/297

### DIFF
--- a/Packages/MIES/MIES_AnalysisBrowser_SweepBrowser.ipf
+++ b/Packages/MIES/MIES_AnalysisBrowser_SweepBrowser.ipf
@@ -533,7 +533,7 @@ Function/DF SB_OpenSweepBrowser()
 
 	string mainWin, renameWin
 
-	Execute "SweepBrowser()"
+	Execute "DataBrowser()"
 
 	mainWin = GetMainWindow(GetCurrentWindow())
 
@@ -547,25 +547,16 @@ Function/DF SB_OpenSweepBrowser()
 	DFREF sweepBrowserDFR = BSP_GetFolder(mainWin, MIES_BSP_PANEL_FOLDER)
 	SB_GetSweepBrowserMap(sweepBrowserDFR)
 
-	renameWin = UniqueName("SweepBrowser", 9, 1)
+	renameWin = UniqueName(SWEEPBROWSER_WINDOW_TITLE, 9, 1)
 	DoWindow/W=$mainWin/C $renameWin
 	mainWin = renameWin
 
 	string/G sweepBrowserDFR:graph = mainWin
 
-	BSP_OpenPanel(mainWin)
+	BSP_InitPanel(mainWin)
 	SB_PanelUpdate(mainWin)
 	return sweepBrowserDFR
 End
-
-Window SweepBrowser() : Graph
-	PauseUpdate; Silent 1		// building window...
-	Display /W=(850.5,168.5,1284,473.75)/K=1  as "SweepBrowser"
-	SetWindow kwTopWin,userdata(panelVersion)=  "2"
-	SetWindow kwTopWin,hook(TA_CURSOR_MOVED)=TimeAlignCursorMovedHook
-	Button button_BSP_open,pos={5.00,5.00},size={25.00,25.00},proc=BSP_ButtonProc_Panel,title="<<"
-	Button button_BSP_open,help={"Open Side Panel"}
-EndMacro
 
 Function/S SB_GetSweepList(win)
 	string win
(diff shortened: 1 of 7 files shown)


timjarsky:
@MichaelHuth should the DB subwindow be resizeable at this point?

t-b:
@timjarsky not yet. Will be ready soon.

MichaelHuth:
Please test if the whole move to the new all-in-one creation macro did not break anything,

t-b:
@MichaelHuth Will do.

t-b:
Review:

Looks good in general. Please note that a current IP bug in the latest nightly makes the subwindow resizing impossible. Until we got a fix you need to stay with e.g. https://www.byte-physics.de/Downloads/WinIgor8_20MAY2019.zip.

Comments:
- Don't define static constants in the middle of the file (MIES_BrowserSettingsPanel.ipf)
- I've added  a fixup commit with a different minimum size for the left browser settings panel
- BSP_UnsetDynamicStartupSettingsOfDataBrowser/BSP_UnsetDynamicSweepControlOfDataBrowser why are these functions Databrowser specific? We do initialize the GUI on startup for either Sweepbrowser or Databrowser so we don't need to enable/disable controls here.
- The function DB_ResetAndStoreCurrentDBPanel() misses to unset user data MIES_BSP_AR_SWEEPFOLDER and friends. It also needs to set the initial value of *all* controls. You see that this is not done because if you enable the DA checkbox and then recreate the marco, the DA checkbox is on in the macro. We use that approach already in DAP_EphysPanelStartUpSettings(). It should also remove the graphs in the settings history panel (DB_ClearGraph).
- Could you remove the horizontal scrollbar in the dashboard?
- I've also added a commit which uses a plaintext notebook for the wave note as this is much faster.
- Settings history panel needs a minimum size

MichaelHuth:
@t-b Did you pushed your commits?

The idea of the Unset* functions is to allow the generation of a rather 'clean' macro where the specific control procedures that are initialized programmatically are removed. The reason is exactly that parts are initialized through functions.
The Unset* functions are only used by DB_ResetAndStoreCurrentDBPanel which works only for DataBrowser windows. Thus the Unset* functions only remove control procedures from the programmatic initilization path for DataBrowser type browser windows.

The panel is recreated in unlocked state, thus MIES_BSP_AR_SWEEPFOLDER is not relevant and overwritten on locking/auto locking to a device. I think however, that the userdata relevant on locking should be removed on unlocking.

t-b:
@MichaelHuth Sorry forgotten done.

t-b:
> I think however, that the userdata relevant on locking should be removed on unlocking.
Good as well.

I'm fine with leaving the Unset* functions as it is.

t-b:
@MichaelHuth Please add a corresponding "close #XXX" to the PR. tHIS IS always a good thing.

t-b:
@MichaelHuth Thx. Nice!

I've pushed one small fix so that the notebook contents are not saved in the recreation macrio.

@timjarsky Ready to test.

t-b:
IP bug is fixed in the next nightly, available from tomorrow.

t-b:
The resizing bug is fixed with the latest nightly, which we now also require.

t-b:
This should be merged before #245.

t-b:
Thanks for the review. I've updated IP8 on the CI machines as well.

PR #317, opened 2019-10-11
Allow tp during daq with indexing

### LINKED ISSUE #217: Allow start and end indexing stimulus set to be the same when TestPulse stimset is selected
When setting the `(first)` and `Indexing End set` to `TestPulse` the following error is returned:

`  (ITC1600_Dev_1) Please select a different indexing end setimset for DA channel 1 referenced by headstage 1`

t-b:
- [x] Find all places where we now have to use `DC_ChannelIsActive` with the new DAQChannelType parameter
- [x] Add tests with locked/unlocked indexing and TP During DAQ channels
- [x] Teach DAP_CheckSettings to not enfore indexing requirements for TP during DAQ channels

Close #217.


t-b:
@timjarsky First ready should be ready.

t-b:
And now with tests as well.

timjarsky:
@t-b I played around and it seems to be working great:

image


Working with two devices may have uncovered a device locking issue that we can look at the next time we chat.

t-b:
@timjarsky All tests passed :)

PR #321, opened 2019-10-16
ACL UserData Editor modification with two column key selection

MichaelHuth:
Added two popup menus that offer all available userdata keys to
be selected independently on each column.

- Currently column sorting resets the popup menu to the first entry.
- The key selection now also has an entry '- empty -' to select the global userdata key.
- When choosing a control type, the key list is reduced accordingly except when choosing SubWindow only (*).

(*) This has to be resolved differently by allowing the Panel selection to select SubWindows as well from the list.

### DIFF
--- a/Packages/MIES/ACL_UserdataEditor.ipf
+++ b/Packages/MIES/ACL_UserdataEditor.ipf
@@ -33,7 +33,13 @@
 // *************************
 //	Date			Version #				Changes
 //	Jan 20, 2007		Version 1.0.0.0			Initial release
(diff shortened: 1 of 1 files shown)


t-b:
@timjarsky We are running more and more into GUI controls which have multiple user data entries. And it was not possible to get an easy look at these. Therefore I figured we can extend that panel to also list  other user data entries.

t-b:
@MichaelHuth Looks good. I've added a small change which allows us to also list graphs. This is required for the databrowser to show up. Does that makse sense?

MichaelHuth:
Yes. Maybe for some future adaption it would make sense to just add any window type that could contain controls with possible userdata entries. The selection list should then have all subwindows as separate entry.

Issue #261, opened 2019-09-03
Rework configuration file handling


t-b:
The current configuration files work for existing users within the orignal scope.

But they are difficult to setup properly for new users and quite tedious to adapt to new controls. See 6acedc25 (ExperimentConfig: Add "Respect ITI for manual initialization", 2018-08-16) for an example which has to touch four places in order to work.

Current limitations:
- Only supports the DAEphys panel
- Only works for explicitly supported controls
- Custom serialization means it is difficult to adapt by hand
- No support for creating a config file from the current settings

A rework would solve all these issues. In addition we can nowadays use the [JSON XOP](https://docs.byte-physics.de/json-xop/) to get good serialization.

The last issue is also an alternative solution for creating a converter from the existing config file to the new config file.

- Implement for DAEphys and Databrowser/SweepBrowser as a first step
- Generic code to save control values from panel, add option via user data to ignore control. Handle controls like headstage clamp modes in a separate function
- Allow to set json path for storage via user data
- Allow human readable names (use the current ones!) as alias for control names. Human readable names are stored in user data. Check for uniqueness.
- Warn if a control does not exist anymore
- As we need to add a lot of new user data, it makes sense to adapt the "Userdata Editor for Controls" (ACL_UserDataEditor.ipf). Get the list of user data names/keys of each control by parsing `S_recreation` from ControlInfo.
 Draft layout: 
![grafik](https://user-images.githubusercontent.com/1185677/66318623-f5e00980-e91b-11e9-9c5c-a960e82b57d9.png)

PR #337, opened 2019-10-25
Generic GUI Load/Save and DA_Ephys Experiment Config Load/Save using customized generic way

MichaelHuth:
- New file MIES_Configuration.ipf with functions to serialize the state of a
  GUI window to a file with generic function CONF_SaveWindow/CONF_RestoreWindow

- Specialized function for DA_Ephys panel save/restore that additionally stores
  the configuration information of the old ExpConfig approach.
  CONF_SaveDAEphys/CONF_RestoreDAEphys

- public functions to save a control, window or window with all its subwindows
  to json

- Restore of controls is configurable by setting userdata properties
  - human readable nice names for each control
  - flag to ignore control on save
  - flag to ignore control on restore
  - priority value that allows to define the order of control restore
  - flag for buttons if they should be activated on restore
  - flag that buttons get only saved if they should be activated on restore
  - flag to hide window while controls get restored
  - flag to disable saving of the control type
  - flag to select value/string/value+string for saving popupmenustate

- Special handling of coupled CheckBoxes (aka RadioButtons) requires a optional
  utility function. This function can be set in userdata of window with these
  radiobuttons.

- Grouping of controls of various types in ControlArrays

- Each grouped element can have its own type

- Disable fields are now ignored if not present on restore

- New Menu entries for save/restore of window states

- Excluded ValDisplay controls for DAEphys panel for save/restore

- Reduced non-GUI parameters for DA_Ephys panel

- Window can be minimized on restore.

closes https://github.com/AllenInstitute/MIES/issues/261

### DIFF
--- a/Packages/MIES/MIES_Configuration.ipf
+++ b/Packages/MIES/MIES_Configuration.ipf
@@ -122,7 +122,6 @@ static StrConstant EXPCONFIG_UDATA_WINHANDLE = "Config_WindowHandle"
 static StrConstant EXPCONFIG_UDATA_RADIOCOUPLING = "Config_RadioCouplingFunc"
 static StrConstant EXPCONFIG_UDATA_CTRLARRAY = "ControlArray"
 static StrConstant EXPCONFIG_UDATA_CTRLARRAYINDEX = "ControlArrayIndex"
-static StrConstant EXPCONFIG_UDATA_SOURCEFILE = "Config_FileName"
 
 static Constant EXPCONFIG_UDATA_MAXCTRLARRAYINDEX = 100
 static Constant EXPCONFIG_JSON_INDENT = 4
@@ -301,7 +300,7 @@ Function CONF_RestoreWindow(fName[, usePanelTypeFromFile])
 				wName = GetMainWindow(GetCurrentWindow())
 				SetWindow $wName, userData($EXPCONFIG_UDATA_SOURCEFILE)=""
 				wName = CONF_JSONToWindow(wName, restoreMask, jsonID)
-				SetWindow $wName, userData($EXPCONFIG_UDATA_SOURCEFILE)=fullFilePath
+				CONF_AddConfigFileUserData(wName, fullFilePath)
 				print "Configuration restored for " + wName
 			else
 				ASSERT(0, "Configuration file entry for panel type has an unknown panel tag (" + panelType + ").")
@@ -324,7 +323,7 @@ Function CONF_RestoreWindow(fName[, usePanelTypeFromFile])
 				jsonID = CONF_ParseJSON(input)
 				SetWindow $wName, userData($EXPCONFIG_UDATA_SOURCEFILE)=""
 				wName = CONF_JSONToWindow(wName, restoreMask, jsonID)
-				SetWindow $wName, userData($EXPCONFIG_UDATA_SOURCEFILE)=fullFilePath
+				CONF_AddConfigFileUserData(wName, fullFilePath)
 				print "Configuration restored for " + wName
 			endif
 		endif
@@ -471,8 +470,7 @@ Function CONF_RestoreDAEphys(jsonID, fullFilePath, [middleOfExperiment, forceNew
 
 		CONF_RestoreHeadstageAssociation(panelTitle, jsonID, middleOfExperiment)
 		CONF_RestoreUserPressure(panelTitle, jsonID)
-
-		SetWindow $panelTitle, userData($EXPCONFIG_UDATA_SOURCEFILE)=fullFilePath
+		CONF_AddConfigFileUserData(panelTitle, fullFilePath)
 
 		filename = GetTimeStamp() + PACKED_FILE_EXPERIMENT_SUFFIX
 		path = CONF_GetStringFromSettings(jsonID, SAVE_PATH)
@@ -510,6 +508,13 @@ Function CONF_RestoreDAEphys(jsonID, fullFilePath, [middleOfExperiment, forceNew
 			Abort
 		endif
 	endtry
+End
+
+/// @brief Add the config file path to the panel as user data
+static Function CONF_AddConfigFileUserData(win, fullFilePath)
+	string win, fullFilePath
+
+	SetWindow $win, userData($EXPCONFIG_UDATA_SOURCEFILE)=fullFilePath
 End
 
 /// @brief Parses a json formatted string to a json object. This function shows a helpful error message if the parse fails
--- a/Packages/MIES/MIES_Configuration.ipf
+++ b/Packages/MIES/MIES_Configuration.ipf
@@ -1,0 +1,2071 @@
+#pragma TextEncoding = "UTF-8"
+#pragma rtGlobals=3 // Use modern global access method and strict wave access.
+#pragma rtFunctionErrors=1
+
+#ifdef AUTOMATED_TESTING
+#pragma ModuleName=MIES_CONF
+#endif
+
+/// @file MIES_Configuration.ipf
+///
+/// @brief __CONF__ Import user settings to configure paramters for Ephys experiments
+///
+/// @anchor Configuration Module brief description
+///******************************************************************************************************************************
+///
+/// The Configuration module allows to save and load the GUI state of complete windows including its subwindows
+///
+/// The main functions are CONF_SaveWindow and CONF_RestoreWindow. CONF_SaveWindow automatically detects if the current window is
+/// a DA_Ephys panel and uses the DA_Ephys GUI state savig configuration settings and saves additional information.
+/// CONF_RestoreWindow also detects if the current window is of DA_Ephys type and handles that specific for this case.
+/// CONF_RestoreDAEphys can be called directly as well if there is no panel open. It will open a new DA_Ephys panel automatically
+/// if no other viable candidate panel was found.
+///
+/// The GUI state is saved in a text file in json format. In the basic structure of the file the main blocks
+/// are the windows and subwindows and within are the control blocks.
+/// A bitmask parameter allows to configure which control information is saved. See WindowControlSavingMask
+/// Currently supported is Value, Position/Size, Userdata, Disabled state and Type of a control.
+/// Not implemented is e.g. Color/Background Color
+/// As Value only the V_Value and/or S_Value and/or S_DataFolder is saved. Special control properties for e.g.
+/// ListBox like V_SelCol are currently not implemented.
+///
+/// How GUI controls are handled can be controlled through userdata settings of each control. The following
+/// userdata settings are supported:
+/// - Config_NiceName: sets a human readable name of the control used in the configuration file
+///                    This name may not be a duplicate of another controls name or ControlArray name.
+///                    Also suffixing with " ControlGroup" is not allowed.
+/// - Config_GroupPath: Sets a group path in which the control appears in the configuration file.
+///                     If not set then "Generic" is used. The group path elements can not contain the '/' character.
+///                     Several levels can be separated by ";", e.g. "level1;level2;".
+///                     In the configuration file the group path elements are suffixed by " ControlGroup" which identifies them
+///                     as path compared to actual control names.
+/// - Config_DontSave: When set to 1 the control is ignored on saving the window.
+/// - Config_DontRestore: When set to 1 the control is ignored on restoring the window.
+/// - Config_PushButtonOnRestore: Only valid with Button controls. When the button is restored a button press
+///                               is issued.
+/// - Config_RestorePriority: Sets a double precision number that defines the order in which controls get restored.
+///                           Lower numbers correspond to earlier restore. If not set a priority of Inf is used.
+///                           If multiple controls have the same priority the order of restore is not defined.
+/// - Config_RadioCouplingFunc: A function name of a function that returns a text wave with lists of coupled
+///                             CheckBoxes (aka RadioButtons). It is assumed that only one CheckBox of the group
+///                             is in enabled state at the same time. See DAP_GetRadioButtonCoupling for example.
+/// - ControlArray: Stores a name of a group of controls. All controls with the same ControlArray name are saved
+///                 as group with that name in the configuration file. ControlArrays save only values and nothing
+///                 is saved if the configured WindowControlSavingMask does not include values.
+/// - ControlArrayIndex: Only valid if also ControlArray is set. Stores the index number of the control in its
+///                      ControlArray. All member of the same ControlArray must have different indices.
+///                      Currently not more than 100 controls are allowed for each ControlArray.
+///
+/// The function WindowtoJSON allows to give optional a list of control types that should be excluded.
+///
+/// If the type of the control was saved, it is checked against the GUI controls type on restore.
+///
+/// CheckBoxes are saved with true/false values.
+///
+/// For the DA_Ephys panel: The function CONF_DefaultSettings sets default values for parameters not present in the GUI.
+///
+/// Special mask flags:
+/// - EXPCONFIG_SAVE_ONLY_RELEVANT: Saves only most relevant value data of a control. It is configured in EXPCONFIG_GUI_PREFERRED
+/// - EXPCONFIG_SAVE_POPUPMENU_AS_STRING_ONLY: Saves value for PopupMenu only as string
+/// - EXPCONFIG_SAVE_POPUPMENU_AS_INDEX_ONLY: Saves value for PopupMenu only as index value
+/// - EXPCONFIG_MINIMIZE_ON_RESTORE: Minimize window while restoring it.
+/// - EXPCONFIG_SAVE_BUTTONS_ONLY_PRESSED: Saves Buttons only when its userdata Config_PushButtonOnRestore is "1"
+///
+/// Handling of window names on restore:
+/// The currently selected windows main window is used as reference. The main window name in the configuration file is
+/// internally replaced with the current main window. This allows to save from "DataBrowser" and load to "DataBrowser2".
+///
+/// Handling of errors in input data:
+/// If input data is of unexpected format an ASSERTion or RTE is issued. The wrapping function converts the RTE to an ASSERT.
+/// The panel restore is aborted and the panel stays in its state. It gets unhided.
+///
+/// Default configuration for generic windows:
+/// By default only the Value property is saved and controls from EXPCONFIG_EXCLUDE_CTRLTYPES excluded.
+///
+/// Default configuration for DA_Ephys panel:
+/// - By default only the Value property is saved
+/// - Controls from DAEPHYS_EXCLUDE_CTRLTYPES are excluded.
+/// - The window is minimized while restoring
+/// - PopupMenu values are saved as string only
+/// - Buttons are only saved if its userdata Config_PushButtonOnRestore is "1"
+/// - Only most relevant data of a control is saved.
+///******************************************************************************************************************************
+
+#if exists("MCC_GetMode") && exists("AxonTelegraphGetDataStruct")
+#define AMPLIFIER_XOPS_PRESENT
+#endif
+
+static StrConstant EXPCONFIG_FIELD_CTRLTYPE = "Type"
+static StrConstant EXPCONFIG_FIELD_CTRLVVALUE = "NumValue"
+static StrConstant EXPCONFIG_FIELD_CTRLSVALUE = "StrValue"
+static StrConstant EXPCONFIG_FIELD_CTRLSDF = "DataSource"
+static StrConstant EXPCONFIG_FIELD_CTRLDISABLED = "Disabled"
+static StrConstant EXPCONFIG_FIELD_CTRLPOSHEIGHT = "Height"
+static StrConstant EXPCONFIG_FIELD_CTRLPOSWIDTH = "Width"
+static StrConstant EXPCONFIG_FIELD_CTRLPOSTOP = "Top"
+static StrConstant EXPCONFIG_FIELD_CTRLPOSLEFT = "Left"
+static StrConstant EXPCONFIG_FIELD_CTRLPOSRIGHT = "Right"
+static StrConstant EXPCONFIG_FIELD_CTRLPOSPOS = "Pos"
+static StrConstant EXPCONFIG_FIELD_CTRLPOSALIGN = "Align"
+static StrConstant EXPCONFIG_FIELD_CTRLUSERDATA = "Userdata"
+static StrConstant EXPCONFIG_FIELD_BASE64PREFIX = "Base64 "
+static StrConstant EXPCONFIG_FIELD_CTRLARRAYVALUES = "Values"
+
+static StrConstant EXPCONFIG_UDATA_NICENAME = "Config_NiceName"
+static StrConstant EXPCONFIG_UDATA_JSONPATH = "Config_GroupPath"
+static StrConstant EXPCONFIG_UDATA_EXCLUDE_SAVE = "Config_DontSave"
+static StrConstant EXPCONFIG_UDATA_EXCLUDE_RESTORE = "Config_DontRestore"
+static StrConstant EXPCONFIG_UDATA_BUTTONPRESS = "Config_PushButtonOnRestore"
+// Lower means higher priority
+static StrConstant EXPCONFIG_UDATA_RESTORE_PRIORITY = "Config_RestorePriority"
+static StrConstant EXPCONFIG_UDATA_WINHANDLE = "Config_WindowHandle"
+static StrConstant EXPCONFIG_UDATA_RADIOCOUPLING = "Config_RadioCouplingFunc"
+static StrConstant EXPCONFIG_UDATA_CTRLARRAY = "ControlArray"
+static StrConstant EXPCONFIG_UDATA_CTRLARRAYINDEX = "ControlArrayIndex"
+static StrConstant EXPCONFIG_UDATA_SOURCEFILE = "Config_FileName"
+
+static Constant EXPCONFIG_UDATA_MAXCTRLARRAYINDEX = 100
+static Constant EXPCONFIG_JSON_INDENT = 4
+static StrConstant EXPCONFIG_FILEFILTER = "Configuration Files (*.json):.json;All Files:.*;"
+static StrConstant EXPCONFIG_CTRLGROUP_SUFFIX = " ControlGroup"
+static StrConstant EXPCONFIG_SETTINGS_FOLDER = "Settings"
+
+// DA_Ephys specific constants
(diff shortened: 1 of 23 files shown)


MichaelHuth:
Should be good to get some feedback.

If this works, should the old ExpConfig approach removed or kept for compatibility?

timjarsky:
> Should be good to get some feedback.
> 
> If this works, should the old ExpConfig approach removed or kept for compatibility?
@t-b @MichaelHuth 
Ultimately, I'd like the old version removed. It's up to you and Thomas whether it is best to remove it in this PR or a separate one.

Packages/MIES/json_functions.ipf
```diff
@@ -428,7 +428,7 @@ Function JSON_AddVariable(jsonID, jsonPath, value, [significance, ignoreErr])
 
 	ignoreErr = ParamIsDefault(ignoreErr) ? JSON_ZFLAG_DEFAULT : !!ignoreErr
 
-	if(trunc(value) == value)
+	if(trunc(value) == value && !numtype(value))
```
t-b: Good find. Let's use the latest JSON XOP and procedure file now that this changes is upstream.

t-b:
I'm not diving into the code yet.

Before I do that we need to make the created JSON file *much* smaller. It is way too verbose. I've pushed  commit which disables user data and disabled state.

- Do we really need to store valdisplay commands? As these are not user editable I don't see a reason.
- Repeated ctrls should be in an array as in

```
      "Check_AD": {
        "NumValue": [0, 0, 0, 1, 1, 1, 1, 1],
        "Type": "CheckBox"
      },
```
- Do you use both `numvalue` and `strvalue` for the popupmenues? What happens if they differ?
- Don't save button state by default.

With these changes we can get to something like 500 lines which is a good start.

How much would it complicate things if you would not store the control type?

timjarsky:
@t-b would it be possible to hide the GUI being configured until the config is complete?

t-b:
> @t-b would it be possible to hide the GUI being configured until the config is complete?

That should be possible.

t-b:
@MichaelHuth please rebase as you still have some XOPs in this branch which are already removed in master. And that makes using IP hard ;)

UserConfigTemplate.txt
```diff
@@ -0,0 +1,773 @@
...
+  "GUI independent configuration data": {
+    "Amplifier Names": "0,1;2,3;4,5;6,7",
+    "Amplifier Serial Numbers": "0;",
+    "Name of Pressure Devices": "Pressure1and2;Pressure3and4;Pressure5and6;Pressure7and8",
+    "Position MCCs": "- none -",
+    "Positive Pressure Calibration Constants": "0.4;0.14;0.05;0.14;0.29;0.03;0.05;0.045",
+    "Save data to": "C:MiesSave",
+    "Stim set file name": ""
+  }
+}
```
timjarsky: @t-b can these be moved to the top?

Packages/MIES/MIES_Configuration.ipf
```diff
@@ -0,0 +1,1785 @@
...
+//                      ControlArray. All member of the same ControlArray must have different indices.
+//                      Currently not more than 100 controls are allowed for each ControlArray.
+//
+// The function WindowtoJSON allows to give optional a list of control types that should be excluded.
+//
+// If the type of the control was saved, it is checked against the GUI controls type on restore.
+//
+// CheckBoxes are saved with true/false values. When grouped in ControlArrays the state is saved as 1, 0.
+//
+// For the DA_Ephy spnael: The function CONF_DefaultSettings sets default values for parameters not present in the GUI.
```
timjarsky: @t-b typo: panel

Packages/MIES/MIES_Configuration.ipf
```diff
@@ -0,0 +1,1785 @@
...
+static Function CONF_DefaultSettings()
+
+	variable jsonID
+
+	jsonID = JSON_New()
+
+	JSON_AddString(jsonID, AMP_TITLE, "0,1;2,3;4,5;6,7")
+	JSON_AddString(jsonID, AMP_SERIAL, "0;")
+	JSON_AddString(jsonID, PRESSURE_DEV, "Pressure1and2;Pressure3and4;Pressure5and6;Pressure7and8")
+	JSON_AddString(jsonID, PRESSURE_CONST, "0.4;0.14;0.05;0.14;0.29;0.03;0.05;0.045")
```
timjarsky: @t-b default PRESSURE_CONST would probably be best set to zero for all headstages.

Packages/MIES/MIES_Configuration.ipf
```diff
@@ -0,0 +1,1785 @@
...
+// userdata settings are supported:
+// - Config_NiceName: sets a human readable name of the control used in the configuration file
+//                    This name may not be a duplicate of another controls name or ControlArray name
+// - Config_JSONPath: Sets a block name in which the control appears in the configuration file.
+//                    If not set then "Generic" is used. The name can not contain the '/' character.
+// - Config_DontSave: When set to 1 the control is ignored on saving the window.
+// - Config_DontRestore: When set to 1 the control is ignored on restoring the window.
+// - Config_PushButtonOnRestore: Only valid with Button controls. When the button is restored a button press
+//                               is issued.
+// - Config_RestorePriority: Sets a number that defines the order in which controls get restored. Lower numbers
```
timjarsky: @t-b can decimals be used for `Config_RestorePriority`? I'm thinking of the case where I forget about a control and later want to weave it in between other controls without needing to change existing priority settings.

UserConfigTemplate.txt
MichaelHuth: When exporting the json information to a string the members are sorted alphabetically. I can give the name of the block "GUI independent configuration data" a prefix that moves it to the top.
The other members at the same level are window names. As long as there are no windows with names that appear alphabetically before the prefix it would stay on top.

Packages/MIES/MIES_Configuration.ipf
MichaelHuth: ok

Packages/MIES/MIES_Configuration.ipf
MichaelHuth: Config_RestorePriority is double precision, I will add that to the brief docu.
I also organized the current priorities roughly in steps of 10.

UserConfigTemplate.txt
t-b: How about something like "Common configuration data"? That would sort it earlier as "ITC*" and "Dev*" and not look too ulgy.

The technical resason for the sorting is the use of the default std::map in our C++ json library. See https://nlohmann.github.io/json/classnlohmann_1_1basic__json_a5e48a7893520e1314bf0c9723e26ea2a.html#a5e48a7893520e1314bf0c9723e26ea2a for some in-depth technical discussion. It is technical possible to add support for a custom sorting order, but it would be some effort.

MichaelHuth: Ok I first had a name starting with "!", but I set it to your recommended value now.

MichaelHuth:
In 6346e2344760ff0f21aa51d0dce8cabb87ed191d I fixed a bug where only the last element of a ControlArray group was restored.

Packages/MIES/MIES_Configuration.ipf
timjarsky: @MichaelHuth I figured and should have caught it long before it reached you.

timjarsky:
@t-b would you layout the user interaction for the new config method?

 In the past, the only method was editing the template file.  Will there be a new MIES menu item for creating a template file? How does one update waveBuilder and not the other MIES panels?

 For individual controls, I like the idea of being able to right-click on a particular control to update its config settings. Related: are there ways we can make it easy for a user to find a particular control in the config file?

Can we handle having multiple configuration files?

t-b:
@timjarsky @MichaelHuth I've chatted with Tim about his questions.

One more thing to implement: The user needs a way to store the settings of one control in a JSON file. The use case being that you setup the GUI, start doing a measurement and then later on realize that you forgot one setting.

t-b:
```igorpro
Window Panel0() : Panel
	PauseUpdate; Silent 1		// building window...
	NewPanel /W=(174,995,503,1118)
	SetDrawLayer UserBack
	DrawText 35,53,"Filename: "
	Button button0,pos={31.00,7.00},size={72.00,25.00},title="Select file..."
	CheckBox check0,pos={34.00,64.00},size={142.00,15.00},title="Enable control selection"
	CheckBox check0,value= 0
	SetVariable setvar0,pos={33.00,87.00},size={284.00,18.00},title="Control name: "
	SetVariable setvar0,value= _STR:""
EndMacro

Function SelectionHook(s)
	STRUCT WMWinHookStruct &s

	switch(s.eventCode)
		case 11: // keyboard
			print s.keyCode
			break
		case 4: // mousemoved
			break
	endswitch

	return 0
End
```

MichaelHuth:
@timjarsky We discussed the way how saving/updating a single control in a present config file could work. This involves two parts:

- How the user selects a control in a window/subwindow
- updating the file once the window + control is known

Our idea for the first step is to introduce a new panel that allows to select a existing configuration file to update. It has a checkbox to enable control selection mode and a setvar that shows the name of the selected control. In control selection mode other windows get an additional window hook that reacts to mouse move and keyboard events. Now the user can move the mouse over a window and (non-hidden) controls under the mouse get marked by a colored frame and the control name is copied to the setvar field in our new panel. The user can now press e.g. "s" to update the control that is currently marked in the selected config file.

Updating an existing configuration file with a control is possible. We currently have not tagged the configuration files for specific panel types. So a saved DA_Ephys panel could be updated with a control from a DataBrowser panel (however it would be ignored on restoring it).

timjarsky:
> @timjarsky We discussed the way how saving/updating a single control in a present config file could work. This involves two parts:
> 
> * How the user selects a control in a window/subwindow
> * updating the file once the window + control is known
> 
> Our idea for the first step is to introduce a new panel that allows to select a existing configuration file to update. It has a checkbox to enable control selection mode and a setvar that shows the name of the selected control. In control selection mode other windows get an additional window hook that reacts to mouse move and keyboard events. Now the user can move the mouse over a window and (non-hidden) controls under the mouse get marked by a colored frame and the control name is copied to the setvar field in our new panel. The user can now press e.g. "s" to update the control that is currently marked in the selected config file.
> 
> Updating an existing configuration file with a control is possible. We currently have not tagged the configuration files for specific panel types. So a saved DA_Ephys panel could be updated with a control from a DataBrowser panel (however it would be ignored on restoring it).

@MichaelHuth Would it be possible for the user to right-click on a control and select a `save control config` menu item. After selection, a save dialog would open that would allow the user to select the appropriate config file. If it is necessary to have the GUI in `show tools` mode, I think I'm okay with that.

MichaelHuth:
> 
> 
> > @timjarsky We discussed the way how saving/updating a single control in a present config file could work. This involves two parts:
> > 
> > * How the user selects a control in a window/subwindow
> > * updating the file once the window + control is known
> > 
> > Our idea for the first step is to introduce a new panel that allows to select a existing configuration file to update. It has a checkbox to enable control selection mode and a setvar that shows the name of the selected control. In control selection mode other windows get an additional window hook that reacts to mouse move and keyboard events. Now the user can move the mouse over a window and (non-hidden) controls under the mouse get marked by a colored frame and the control name is copied to the setvar field in our new panel. The user can now press e.g. "s" to update the control that is currently marked in the selected config file.
> > Updating an existing configuration file with a control is possible. We currently have not tagged the configuration files for specific panel types. So a saved DA_Ephys panel could be updated with a control from a DataBrowser panel (however it would be ignored on restoring it).
> 
> @MichaelHuth Would it be possible for the user to right-click on a control and select a `save control config` menu item. After selection, a save dialog would open that would allow the user to select the appropriate config file. If it is necessary to have the GUI in `show tools` mode, I think I'm okay with that.

@timjarsky We looked into this and the main blocker for this approach is that Igor Pro does not support 'right click'. It knows only one mouse button, i.e. left and right click is the same action. This is probably related to Igor Pros macintosh origins.

grafik

timjarsky:
@t-b 

>Igor Pro does not support 'right click'

I thought the contextual menu was equivalent to right click on a PC and cntrl-click on a mac. Is that something else?

```
Control Structure eventMod Field
The eventMod field appears in the built-in structure for each type of control. It is a bitfield defined as follows:
Int32 eventMod	Bit 0:	A mouse button is down.
	Bit 1:	Shift key is down.
	Bit 2:	Option (Macintosh ) or Alt (Windows ) is down.
	Bit 3:	Command (Macintosh ) or Ctrl (Windows ) is down.
	Bit 4:	Contextual menu click occurred.

```

t-b:
- [ ] indent with 4 spaces to improve readability when user hand edit the settings files
- [ ] Can we have true/false for checkboxes in array controls as well? If yes, documentation needs adaptation as well.
- [ ] The controls popup_Hardware_AvailITC1600s, setvar_Hardware_YokeList,
      setvar_Hardware_Status and popup_Hardware_YokedDACs don't need to be saved
      and restored
- [ ] The config files should have .json as file suffix to distinguish them
      from the old config files and also enable syntax highlightning. The open
      dialogs should then also default to .json as file suffix.
- [ ] Would it be possible to create a hierarchy with a list in
      Config_JSONPath? E.g. Config_JSONPath="Level1;Level2" would result in

```json
"Level1" :
{
  "Level2" :
  {
    "ControlA" : ...
  }
}
```

This would help in grouping the controls and reflect the grouping in the GUI.

- [ ] The panel should store the config file name which was loaded into it in
      its user data. This allows us to store the config file name in the
      labnotebook for the DAEphys panel. (DC_DocumentChannelProperty in
      DC_PlaceDataInHardwareDataWave)
- [ ] Loading the settings from the file is currently not easy enough. We want
      to have a shortcut (/5 for now) for the DAEphys panel which handles opening
      the panel and loading a default config file. In addition we can have a menu
      entry for loading non-default config files. For the DAEphys panel we want to
      avoid forcing the user to open the panel before.
- [ ] Could you check that if the loaded json has a parse error that we
      give a good error message to the user? This will be quite common I fear.

Code:
- First impression is good!

Some small comments though:
- [ ] Loading and saving file contents via FStatus/FBinRead should be moved into a generic
      helper function. Similiarly writing contents to a file with save dialog. You have something similiar already in BHT_ReadTextFile
      and BHT_SaveJSON.
- [ ] Please use `GetUniqueInteger` instead of manual enoise calls
- [ ] GetListOfUnlockedDevices instead of hand-coding it
- [ ] CONF_RequireConfigBlockExists still uses the old name for the common block
- [ ] CONF_GetAllWindows: This function should not use a pass-by-ref parameter.
      And why do you need AbortOnRTE? This pops up the debugger here.
- [ ] CONF_AllWindowsToJSON is missing docu or needs to be static. Testing static functions is possible as it has the ModuleName in this case.
- [ ] Please make the embedded documentation available to the end user. MIES_AnalysisFunctions.ipf does that already.
- [ ] CONF_GetUserdataKeys should live in MIES_GUIUtilities.ipf and be
      available for all. It would also read better if you used variables for the
      magic numbers. Or even calculate the magic numbers from the prefix/suffix strings.
- [ ] CONF_ControlTypeToName/CONF_Name2ControlType should also be available for all.
- [ ] CONF_Amplifiers is the "not implemented" comment still accurate? If yes please add a @todo in the code as well.
- [ ] How about renaming "Config_JSONPAth"? I think the name is a bit misleading nowadays. Something like "Config_BlockNameList" maybe?
- [ ] EXPCONFIG_SAVE_POPUPMENU_AS_INDEX_ONLY is duplicated in line 70
- [ ] MIES_Constants.ipf still has some BASE_* PRECISION_* variables from the num2strEffective branch

```text
  git diff --check master..
  Packages/MIES/MIES_Constants.ipf:1183: leftover conflict marker
  Packages/MIES/MIES_Constants.ipf:1192: leftover conflict marker
  Packages/MIES/MIES_Constants.ipf:1193: leftover conflict marker
  Packages/MIES/MIES_Constants.ipf:1247: leftover conflict marker
  Packages/Testing-MIES/UTF_Configuration.ipf:74: trailing whitespace.
  +    NewPanel/W=(11,203,273,387)/HOST=#
  Packages/Testing-MIES/UTF_Configuration.ipf:108: trailing whitespace.
  +
  Packages/Testing-MIES/UTF_Configuration.ipf:190: trailing whitespace.
  +
  Packages/Testing-MIES/UTF_Configuration.ipf:200: trailing whitespace.
  +
  Packages/Testing-MIES/UTF_Configuration.ipf:202: trailing whitespace.
  +
  Packages/Testing-MIES/UTF_Configuration.ipf:204: trailing whitespace.
  +
```

Code style:
- `input = ""`: variables with intial values should be on their own line

And it needs to be rebased against master.

t-b:
@MichaelHuth For marking control groups we will use a suffix as in "Testpulse ControlGroup".

t-b:
@MichaelHuth One reason for the errored out documentation is that the doxygen filter script had a bug. I've fixed that in https://github.com/AllenInstitute/MIES/pull/383. Feel free to use that branch in here as well.

t-b:
@timjarsky Ready for a first test.

timjarsky:
@t-b let's screen share next week; I'm seeing MCC an opening bug where MIES continues to try to open the MCC after it has already been opened. I saw this when using the `load window configuration` with the DA_Ephys panel.

Where should the DA_Ephys config JSON get saved, and how is it named so that `load standard configuration will work?

t-b:
@timjarsky Jeep, let's do that on tuesday.

> Where should the DA_Ephys config JSON get saved, and how is it named so that `load standard 
> configuration will work?

All json settings files in `Packages/Settings` are loaded for the standard configuration. The files are not yet sorted before loading, but that is one the todo list.

MichaelHuth:
> @timjarsky Jeep, let's do that on tuesday.
> 
> > Where should the DA_Ephys config JSON get saved, and how is it named so that `load standard
> > configuration will work?
> 
> All json settings files in `Packages/Settings` are loaded for the standard configuration. The files are not yet sorted before loading, but that is one the todo list.

I sort the file list now alphabetically. I also prefixed the files with 1_* and 2_* for DA_Ephys and Databrowser.

timjarsky:
@MichaelHuth when configuring the DA_ephys the MCC isn't opening and i get the following assertion (despite disabling Evil mode):

  !!! Assertion FAILED !!!
  Message: "Evil kittens prevented MultiClamp from opening - FULL STOP"
  Stacktrace:
CONF_AutoLoader(...)#L180 [MIES_Configuration.ipf]
CONF_RestoreWindow(...)#L272 [MIES_Configuration.ipf]
CONF_RestoreDAEphys(...)#L379 [MIES_Configuration.ipf]
  Time: 2019-11-25T13:48:47-08:00
  Locked device: [- none -]
  Current sweep: [- none -]
  DAQ: [- none -]
  Testpulse: [- none -]
  Experiment: Untitled.pxp
  Igor Pro version: 8.0.4.2 (34722)
  MIES version:

t-b:
@MichaelHuth 
```
$ ./tools/build-documentation.sh
Start building the documentation
Errors building the documentation
Doxygen says:
E:/projekte/mies-igor/Packages/MIES/MIES_Configuration.ipf:600: warning: argument 'basePAth' of command @param is not found in the argument list of CONF_TraversalFinder(variable jsonID, string basePath, string niceName)
E:/projekte/mies-igor/Packages/MIES/MIES_Configuration.ipf:607: warning: The following parameters of CONF_TraversalFinder(variable jsonID, string basePath, string niceName) are not documented:
  parameter 'basePath'
(
```

@timjarsky We can look into that today.

t-b:
@MichaelHuth 

- [ ] Tests are not passing
- [ ] Store amplifier settings and pressure settings in common data block (Read ChanAmpSettings and PressureDataWave). Remove left over hard coded settings.

timjarsky:
@t-b when attempting to save the databrowser panel I get:

CONF_SaveWindow(...)#L226 [MIES_Configuration.ipf]
CONF_AllWindowsToJSON(...)#L1194 [MIES_Configuration.ipf]
CONF_WindowToJSON(...)#L1365 [MIES_Configuration.ipf]
  Time: 2019-11-27T10:30:19-08:00
  Locked device: [ITC1600_Dev_1]
  Current sweep: [NaN]
  DAQ: [DAQ_NOT_RUNNING]
  Testpulse: [TEST_PULSE_NOT_RUNNING]
  Experiment: 2019_11_27_102839.pxp
  Igor Pro version: 8.0.4.2 (34722)
  MIES version:
  Release_2.0_20190602-505-gc84a85a2
Date and time of last commit: 2019-11-26T10:59:06+01:00
Submodule status: 
160000 795f5b27d8a60e4da5909a0145b357e0fefa86ff 0	Packages/IPNWB
160000 7b6d2115733523463bb9d0250cee38a41a88d8b3 0	Packages/ITCXOP2
160000 aa1858d915ee28ebf5974f59f0bc19a421e5c28e 0	Packages/ZeroMQ
160000 48bb7e158c86cbe387dfd9fbfeecba5c7a614d47 0	Packages/doc/doxygen-filter-ipf
160000 a097151fbc9a8a8f2e504c9e46adcb327a19c8f3 0	Packages/unit-testing

t-b:
@timjarsky Thanks, we look into it.

MichaelHuth:
@t-b design suggestion:

```
"Headstage Association" :
  [ (note: fixed size NUM_HEADSTAGES)
    {
      "Amplifier Name" : "0",
      "Amplifier Serial" : 0,
      "VC DA" : 0,
      "VC AD" : 0,
      "VC DA Gain" : 20, (note: old function presses auto fill button)
      "VC AD Gain" : 0.01,
      "VC DA Unit" : "mV",
      "VC AA Unit" : "pA",
      "IC DA" : 0,
      "IC AD" : 0,
      "IC DA Gain" : 400,
      "IC AD Gain" : 0.01,
      "IC DA Unit" : "pA",
      "IC AA Unit" : "mV",
      "Pressure Device" : "Pressure1and2", (note: If this is not - none - then button Enable is pressed at the end)
      "Pressure DA" : 0,
      "Pressure AD" : 0,
      "Pressure DA Gain" : 2,
      "Pressure AD Gain" : 0.5,
      "Pressure DA Unit" : "psi",
      "Pressure AA Unit" : "psi",
      "Pressure TTLA" : 1,
      "Pressure TTLB" : 2,
      "Pressure Constant Negative" : 0,
      "Pressure Constant Positive" : 0
    },
    {
      (note: if object has no keys then headstage is off)
    },
    {
      ...
    },
    ... for all 8 headstages
  ],

"User Pressure Devices" :
  [
    {
      "DAC Device" : "Dev1", (note: Enable button is pressed after setting)
      "AD" : 0
    },
    {
    ....
    }
  ]

```

t-b:
```json

"Headstage Association" :
{
  "0" : {
      "Amplifier Name" : "0",
      "Amplifier Serial" : 0,
      "VC DA" : 0,
      "VC AD" : 0,
      "IC DA" : 0,
      "IC AD" : 0,
      "Pressure Device" : "Pressure1and2",
      "Pressure DA" : 0,
      "Pressure AD" : 0,
      "Pressure DA Gain" : 2,
      "Pressure AD Gain" : 0.5,
      "Pressure DA Unit" : "psi",
      "Pressure AA Unit" : "psi",
      "Pressure TTLA" : 1,
      "Pressure TTLB" : 2,
      "Pressure Constant Negative" : 0,
      "Pressure Constant Positive" : 0
    }, (note: auto fill is always pressed when an amplifier serial is given)
    "1" : null
      (note: null means headstage is not active)
    }
},
"User Pressure Devices" :
    {
      "DAC Device" : "Dev1", (note: Enable button is pressed after setting)
      "AD" : 0
    }
```

@MichaelHuth I went through your draft with Tim. And we had a couple of changes. First we always do "Auto fill gains" and therefore don't need Units/Gains settable for the user. You need to press enable only once for the pressure device and also only once for user pressure device (And not once per headstage as I erroneously told you). ANd for better readability I think we should have the zero-based headstage number as key for each entry. Not-set headstages have null as value. And one can only have one user pressure device.

t-b:
@MichaelHuth Needs rebase as well.

MichaelHuth:
@t-b Is there a convenient way to retrieve the MCC window title (aka Amplifier Title) depending on the Amplifier Serial?

t-b:
@MichaelHuth 

test output:
```
MIES_include.ipf:60:0: error: include file not found

MIES_include.ipf:60:0: error: include file not found

MIES_include.ipf:60:0: error: include file not found

MIES_include.ipf:60:0: error: include file not found

MIES_include.ipf:60:0: error: include file not found

MIES_include.ipf:60:0: error: include file not found

MIES_include.ipf:60:0: error: include file not found

MIES_include.ipf:60:0: error: include file not found

MIES_include.ipf:60:0: error: include file not found

MIES_include.ipf:60:0: error: include file not found

MIES_include.ipf:60:0: error: include file not found

MIES_include.ipf:60:0: error: include file not found

MIES_include.ipf:60:0: error: include file not found

MIES_include.ipf:60:0: error: include file not found

MIES_include.ipf:60:0: error: include file not found

MIES_include.ipf:60:0: error: include file not found

MIES_include.ipf:60:0: error: include file not found

MIES_include.ipf:60:0: error: include file not found
```

MichaelHuth:
@t-b Could I get another test report hint, why the tests are failing?

t-b:
@MichaelHuth Sure.
[Desktop.zip](https://github.com/AllenInstitute/MIES/files/3960976/Desktop.zip)

MichaelHuth:
I see that starting the TP fails in DAP_CheckHeadStage with

(Dev1) The configured gain for the AD channel 1 differs from the one in the "DAC Channel and Device Associations" menu (0.01 vs 0.05).
(Dev1) Please restart DAQ or TP to use the automatically imported gains from MCC.

I currently have problem reproducing this error here.
The programmatical press of "Auto-Fill" syncs these values between MCC and MIES here properly.

t-b:
@MichaelHuth That test passes for ITC18USB_DEV_0 so test in general should be fine. Could you just start TP again?

t-b:
- [ ] investigate duplicated calls to `CONF_JSONToControl`
- [ ] Use unused AD channels for async
- [ ] forceNewPanel for tagged panels when using load settings
- [ ] trim whitespace for json file comparison tests

t-b:
@MichaelHuth Changes are pushed.

timjarsky:
nice!

image

timjarsky:
@t-b If the config JSON doesn't specify a `Stim set file name`, I don't think MIES needs to ask the user to specify one before proceeding with the config.

timjarsky:
@t-b super impressed that it can load two da_ephys configs. I'm ready to approve as is.

PR #338, opened 2019-10-28
Hardware Tests: Check rate and unit in NWB files as well

### LINKED ISSUE #333: NWBv1 Export: Check starting_time's rate and unit attributes
We have a report that the exported sampling interval for a TimeSeries is not correct.

And looking at the tests https://github.com/AllenInstitute/MIES/blob/8fc9add71e723e991ee7e42d0ae5c073ab2f336c/Packages/Testing-MIES/UTF_TestNWBExportV1.ipf#L310-L315 revealed that we are not checking it.

Check these as well.

t-b:
Due to a recent bug report we found out that we don't check the stored
rate and unit for the exported NWB files.

While nothing is currently broken let's just add a quick test.

Close #333.


t-b:
@timjarsky Thanks, I'll have a look.

PR #339, opened 2019-10-28
SweepFormula: Add findlevel operation

### LINKED ISSUE #281: SweepFormula: Add FindLevel operation
Add `findlevel(data, level [, edge])` with the following parameters:
- data: some wave data
- level: numeric level to search a value passing through
- edge: Type of crossings to search: 0 (default, both increasing and decreasing), 1 (increasing), 2 (decreasing)

@timjarsky @ukos-git Does that make sense?

t-b:
Close #281.

- [x] Change layer assertion to `<=`

### DIFF
--- a/Packages/MIES/MIES_Constants.ipf
+++ b/Packages/MIES/MIES_Constants.ipf
@@ -401,10 +401,6 @@ Constant TIME_ALIGNMENT_MAX           = 3
 
 StrConstant WAVE_BACKUP_SUFFIX = "_bak"
 
-/// Define error/success values to be used with the WSE engine
-Constant TI_WRITEACK_SUCCESS = 0
-Constant TI_WRITEACK_FAILURE = -1
-
 /// @name Test pulse modes
 /// @anchor TestPulseRunModes
 /// @{
--- a/Packages/MIES/MIES_Constants.ipf
+++ b/Packages/MIES/MIES_Constants.ipf
@@ -1179,3 +1179,11 @@ Constant GAIN_AFTER_DAQ  = 0x2
 
 /// @brief User data on the stimset controls listing all stimsets in range
 StrConstant USER_DATA_MENU_EXP = "MenuExp"
+
+/// @name Find level edge types
+/// @anchor FindLevelEdgeTypes
+/// @{
+Constant FINDLEVEL_EDGE_INCREASING = 1
+Constant FINDLEVEL_EDGE_DECREASING = 2
+Constant FINDLEVEL_EDGE_BOTH       = 0
+/// @}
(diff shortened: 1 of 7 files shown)


timjarsky:
@t-b findlevel with multiple sweeps is working differently than I expected; the level X from the first sweep is being used in the second sweep.

e.g., 
```
data(
[(findlevel(data(
cursors(A,B),
channels(AD1),
sweeps()
),0))
,
(findlevel(data(
cursors(A,B),
channels(AD1),
sweeps()
),0) + 10)],
channels(AD1),
sweeps()
)
```

gives:

image

from:

image

Is that as intended? I expected the spikes from each sweep to be in the formula plot from the zero-crossing forward in time.

t-b:
@timjarsky At the moment `findlevel` is treating its input as 1D wave. Therefore what you expect is probably not yet implemented.

A MWE is

```
findlevel([[10, 20], [11, 21]], 15)
```

as that only returns 1 entry at the moment.

Should it return two?

timjarsky:
https://github.com/AllenInstitute/MIES/pull/339#issuecomment-548572804

@t-b I would expect the same behavior as min, max, or mean. Does that make sense?

timjarsky:
> [#339 (comment)](https://github.com/AllenInstitute/MIES/pull/339#issuecomment-548572804)
> 
> @t-b I would expect the same behavior as min, max, or mean. Does that make sense?

I guess my comment isn't helpful; I just checked and min, max, and mean don't work with the example array above.

t-b:
@timjarsky Jeep, got it.

t-b:
@timjarsky Should be done. I've also introduced a generic FindLevel wrapper which does not need to copy around data if used on a 2D wave. We can use this wrapper also in other places in the future. I'll make an issue to do that.

t-b:
@ukos-git The example from Tim requires us to have different ranges for each sweep. I've hacked something into it right now, but that does not work yet. We need to discuss how we could handle this.

ukos-git:
A union function would be a better and simpler solution.

timjarsky:
@t-b is the assignment still correct? If yes, I'll go ahead and approve.

t-b:
@timjarsky Thanks for checking. The assignment was wrong.

timjarsky:
> 
> Possible solutions:
> 
> * Let `data` return a wave reference wave with the correct x scaling for each extracted sweep
> * Use the union of all sweep ranges and fit the data into it
> 
@t-b @ukos-git  If we have the resources, I propose we only resample when necessary. For example, averaging two waves together will require resampling, but taking the mean of two waves does not. In this way, we can mostly avoid issues that come with resampling (changing the data, ringing, etc.) but the user experience is seamless.

t-b:
@timjarsky I've fixed the layer issue.

PR #340, opened 2019-10-28
More generic sweep formula example

t-b:


Packages/MIES/MIES_DataBrowser.ipf
```diff
@@ -168,10 +168,10 @@ Function DB_ResetAndStoreCurrentDBPanel()
 	CheckBox check_BrowserSettings_DB_Failed WIN = $bsPanel, value= 0
 
 	sfFormula = BSP_GetSFFormula(panelTitle)
-	ReplaceNotebookText(sfFormula, "data(\rcursors(A,B),\rchannels(AD),\rsweeps(0;1;2;3;4;5)\r)")
+	ReplaceNotebookText(sfFormula, "data(\rcursors(A,B),\rchannels(AD),\rsweeps()\r)")
```
timjarsky: @t-b cursors are not on the graph by default. What about `data([0,10], channels(AD), sweeps())`

t-b: @timjarsky If the cursors are not present we currently take the full range so this does not bug out.

timjarsky: @t-b i did not know that. Very helpful! Thank you.

PR #342, opened 2019-10-29
Cooler restore cursors

### LINKED ISSUE #296: RestoreCursors: Fallback to another trace if the old one could not be found

t-b:
Close #296.

### DIFF
--- a/Packages/MIES/MIES_GuiUtilities.ipf
+++ b/Packages/MIES/MIES_GuiUtilities.ipf
@@ -1083,7 +1083,7 @@ Function RestoreCursors(graph, cursorInfos)
 	string graph
 	WAVE/T cursorInfos
 
-	string traceList, cursorTrace
+	string traceList, cursorTrace, info
 	variable i, numEntries, numTraces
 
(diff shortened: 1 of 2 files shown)


Packages/MIES/MIES_GuiUtilities.ipf
```diff
@@ -1083,21 +1083,33 @@ Function RestoreCursors(graph, cursorInfos)
...
-		cursorTrace = StringByKey("TNAME", cursorInfos[i])
+		if(IsEmpty(info)) // cursor was not active
+			continue
+		endif
+
+		cursorTrace = StringByKey("TNAME", info)
 
 		if(FindListItem(cursorTrace, traceList) == -1)
-			continue
+			// trace is not present anymore, use the first one instead
```
timjarsky: @t-b is there a case(s) where the first trace isn't also the lowest number sweep?

t-b: @timjarsky For the databrowser/sweepbrowser we now probably have the first trace being from the first sweep. But as that is a generic routine it does have no idea what the traces are about.

PR #343, opened 2019-10-29
Add test for correct readout of TP data after DAEphys panel was closed

MichaelHuth:
This test checks if ASYNC_ReadOut treats remaining data properly even if the panel was closed.

Currently the tests fails if there is still data in the readout queue after the DAEphys panel was closed. So the fail is timing dependent and does not happen always.

### DIFF
--- a/Packages/MIES/MIES_TestPulse.ipf
+++ b/Packages/MIES/MIES_TestPulse.ipf
@@ -414,7 +414,6 @@ static Function TP_RecordTP(panelTitle, BaselineSSAvg, InstResistance, SSResista
 	variable delta, i, ret, lastPressureCtrl
 	WAVE TPStorage = GetTPStorage(panelTitle)
 	WAVE hsProp = GetHSProperties(panelTitle)
-	Wave GUIState  = GetDA_EphysGuiStateNum(panelTitle)
 	variable count = GetNumberFromWaveNote(TPStorage, NOTE_INDEX)
 	variable lastRescaling = GetNumberFromWaveNote(TPStorage, DIMENSION_SCALING_LAST_INVOC)
 
--- a/Packages/Testing-MIES/UTF_BasicHardwareTests.ipf
+++ b/Packages/Testing-MIES/UTF_BasicHardwareTests.ipf
@@ -1862,6 +1862,43 @@ Function StartDAQDuringTP_IGNORE(device)
 	wv[%$"Analysis function (generic)"][%Set] = "WriteIntoLBNOnPreDAQ"
 End
 
+// UTF_TD_GENERATOR HardwareMain#DeviceNameGeneratorMD1
+Function AbortTP([str])
+	string str
+
+	STRUCT DAQSettings s
+	InitDAQSettingsFromString(s, "MD1_RA0_I0_L0_BKG_1_RES_0")
+	AcquireData(s, str, startTPInstead=1)
+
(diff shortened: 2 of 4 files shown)


t-b:
@MichaelHuth Thanks. I'll look into that.

Packages/Testing-MIES/UTF_BasicHardwareTests.ipf
```diff
@@ -1862,6 +1862,34 @@ Function StartDAQDuringTP_IGNORE(device)
 	wv[%$"Analysis function (generic)"][%Set] = "WriteIntoLBNOnPreDAQ"
 End
 
+Window AbortTP_dummy() : Panel
+	PauseUpdate; Silent 1		// building window...
+	NewPanel /W=(150,77,450,277) as "AbortTP_dummy"
+	SetWindow kwTopWin,userdata(random_change)=  "438.25"
+EndMacro
```
t-b: Why do we need this macro?

Packages/Testing-MIES/UTF_BasicHardwareTests.ipf
```diff
@@ -1862,6 +1862,34 @@ Function StartDAQDuringTP_IGNORE(device)
...
+End
+
+Function AbortTP_REENTRY([str])
+	string str
+
+	string device
+
+	device = StringFromList(0, str)
+	KillWindow $device
+	ASYNC_Stop(timeout = 5)
```
t-b: How about calling ASYNC_START here as in

```igorpro
try
 ASYNC_STOP(timeout = 5)
 aborted = 0
catch
 err = getRTError(1)
  aborted = 1
endtry

ASYNC_Start(threadprocessorCount, disableTask=1)

if(aborted)
  FAIL()
else
  PASS()
endif
```

as that does not influence the other test cases?

t-b:
@MichaelHuth So turns out that deleting the GUI state wave when unlocking the device is the wrong thing to do. I've reverted the relevant commit. And is fixes here.

What do you think?

Packages/Testing-MIES/UTF_BasicHardwareTests.ipf
MichaelHuth: It was a residue from another test approach I missed to delete. I fixed the commit.

Packages/Testing-MIES/UTF_BasicHardwareTests.ipf
MichaelHuth: ok, I fixed the commit.

MichaelHuth:
Your change fixes the crash, but not the problem that TP_RecordTP might get the reference to a wrong GUIState wave. I am also not sure if it is the actual right solution for the problem.

Also: TP_RecordTP does not use the GUIState wave from line 416.

Packages/MIES/MIES_TestPulse.ipf
```diff
@@ -414,7 +414,6 @@ static Function TP_RecordTP(panelTitle, BaselineSSAvg, InstResistance, SSResista
 	variable delta, i, ret, lastPressureCtrl
 	WAVE TPStorage = GetTPStorage(panelTitle)
 	WAVE hsProp = GetHSProperties(panelTitle)
-	Wave GUIState  = GetDA_EphysGuiStateNum(panelTitle)
```
timjarsky: @t-b I guess at one point TP_RecordTP used the GUI state wave?

Issue #302, opened 2019-09-28
plotting vs stimulus set sweep number


timjarsky:
It's common for users to acquire multiple repeats of the same stimulus set. To plot Y data against stimset sweep number the user can currently do something like what is shown below for a 5 sweep stimulus set acquired 3X.

```
max(data([0,1000], channels(AD), sweeps())

vs

[0;1;2;3;4;0;1;2;3;4;0;1;2;3;4;]

```

The above does not work when there are different stimulus sets on different headstages or when user skips sweeps. So, it would be nice to have something like: listFromStimSetSweepNumber() that returns the `set cycle count` for each sweep (on each channel).

t-b:
@ukos-git https://alleninstitute.github.io/MIES/labnotebook-docs.html

t-b:
Closed via #345.

Issue #200, opened 2019-06-12
Make bamboo artifacts publically available


t-b:
One option:
- [ ] Upload artifacts to website using a unique folder
- [ ] Visualize JUNIT files with https://gitlab.com/inorton/junit2html
- [ ] Post link to the folder as "Details"

t-b:
See https://docs.atlassian.com/atlassian-bamboo/REST/6.9.2/#d2e1674 for the bamboo REST API

Recipe:
Fetch JSON from `http://bamboo.corp.alleninstitute.org/rest/api/latest/result/MIES-CG/branch/<XXX>` where XXX is the branch name.

Example output

```xml

<results expand="results">
<results expand="result" start-index="0" max-result="1" size="1">
<result key="MIES-CG0-36" state="Successful" lifeCycleState="Finished" number="36" id="167994449">
<link href="http://bamboo.corp.alleninstitute.org/rest/api/latest/result/MIES-CG0-36" rel="self"/>
<plan key="MIES-CG0" name="MIES Igor - Checks (Github) - master" shortName="master" shortKey="CG0" type="chain_branch" enabled="true">
<master key="MIES-CG" name="MIES Igor - Checks (Github)" shortName="Checks (Github)" shortKey="CG" type="chain" enabled="true">
<planKey>
<key>
MIES-CG</key>
</planKey>
<link href="http://bamboo.corp.alleninstitute.org/rest/api/latest/plan/MIES-CG" rel="self"/>
</master>
<planKey>
<key>
MIES-CG0</key>
</planKey>
<link href="http://bamboo.corp.alleninstitute.org/rest/api/latest/plan/MIES-CG0" rel="self"/>
</plan>
<master key="MIES-CG" name="MIES Igor - Checks (Github)" shortName="Checks (Github)" shortKey="CG" type="chain" enabled="true">
<planKey>
<key>
MIES-CG</key>
</planKey>
<link href="http://bamboo.corp.alleninstitute.org/rest/api/latest/plan/MIES-CG" rel="self"/>
</master>
<buildResultKey>
MIES-CG0-36</buildResultKey>
<planResultKey>
<key>
MIES-CG0-36</key>
<entityKey>
<key>
MIES-CG0</key>
</entityKey>
<resultNumber>
36</resultNumber>
</planResultKey>
<buildState>
Successful</buildState>
<buildNumber>
36</buildNumber>
</result>
</results>
[...]

PR #345, opened 2019-10-30
Feature/sweepformula labnotebook

ukos-git:
#302 

Use `labnotebook("Set Cycle Count", sweeps())` to gather the numerical value for the key `Set Cycle Count`.

Since labnotebook entries do have whitespaces, the formula parser is extended to not strip white spaces in between quotation marks and filters out leading and trailing quotation marks before writing strings.

* [x] documentation
* [x] tests with fake databrowser/sweepbrowser window


ukos-git:
requires documentation

Packages/doc/labnotebook-docs.rst
```diff
@@ -4,18 +4,28 @@ Labnotebook documentation for developers
...
-
-Assuming you read the data from NWB files these arrays are located in
-`/general/labnotebook/$device` and are called `numericalKeys`,
-`numericalValues`, `textualKeys` and `textualValues`. Each pair, the `key` and
-`value` array, is used to describe all metadata entries and also hold the value
-of each entry.
+In MIES, metadata of sweeps and test pulses get stored in a data structure
+called a *labnotebook*. The general structure is like a paper-based lab journal
+with columnar data entry storage. Each measurement has an associated date and
+time of the data acquisition and any further modifications. After data
```
timjarsky: I'm not sure what `modifications` are. Perhaps provide an example?

Packages/Testing-MIES/UTF_SweepFormula.ipf
```diff
@@ -55,6 +55,40 @@ static Function primitiveOperations()
...
+
+	// allow # inside strings
+	jsonID0 = JSON_Parse("\"#\"")
+	jsonID1 = SF_FormulaParser("\"#\"")
+	WARN_EQUAL_JSON(jsonID0, jsonID1)
+
+	// do not evaluate calculations in strings
+	jsonID0 = JSON_Parse("\"1+1\"")
+	jsonID1 = SF_FormulaParser("\"1+1\"")
+	WARN_EQUAL_JSON(jsonID0, jsonID1)
```
t-b: Can we have another one here which inserts just an empty string?

```
	jsonID0 = JSON_Parse("\"\"")
	jsonID1 = SF_FormulaParser("\"\"")
	WARN_EQUAL_JSON(jsonID0, jsonID1)
```

Packages/MIES/MIES_SweepFormula.ipf
```diff
@@ -319,6 +332,8 @@ Function SF_FormulaParser(formula)
 	if(!cmpstr(buffer, formula))
 		if(GrepString(buffer, "^(?i)[0-9]+(?:\.[0-9]+)?(?:[\+-]?E[0-9]+)?$"))
 			JSON_AddVariable(jsonID, jsonPath, str2num(formula))
+		elseif(GrepString(buffer, "^\".*\"$"))
+			JSON_AddString(jsonID, jsonPath, buffer[1, strlen(buffer) - 2])
```
t-b: Is it by design that you exploit the fact that in IP string indexing a la `b[1, 0]` for `\"\"` gives an empty string? I find that a bit difficult to grasp.

Packages/doc/labnotebook-docs.rst
t-b: @ukos-git Nice enhancement. I do second the question regarding `modification`.

Packages/MIES/MIES_SweepFormula.ipf
```diff
@@ -718,6 +734,29 @@ Function/WAVE SF_FormulaExecutor(jsonID, [jsonPath, graph])
...
+			JSONtype = JSON_GetType(jsonID, jsonPath + "/0")
+			ASSERT(JSONtype == JSON_STRING, "first parameter needs to be a string labnotebook key")
+			str = JSON_GetString(jsonID, jsonPath + "/0")
+
+			WAVE sweeps = SF_FormulaExecutor(jsonID, jsonPath = jsonPath + "/1", graph = graph)
+			ASSERT(DimSize(sweeps, COLS) < 2, "Sweeps are one-dimensional.")
+
+			WAVE numericalValues = GetLBNumericalValues(graph)
+			numIndices = DimSize(sweeps, ROWS)
+			Make/FREE/N=(numIndices) out = NaN
```
t-b: Make `out` a double precision wave.

Packages/MIES/MIES_SweepFormula.ipf
```diff
@@ -718,6 +734,29 @@ Function/WAVE SF_FormulaExecutor(jsonID, [jsonPath, graph])
...
 			break
+		case "labnotebook":
+			/// `labnotebook(string key, array sweeps)`
+			///
+			/// return lab notebook @p key for all @p sweeps
+			ASSERT(!ParamIsDefault(graph) && !IsEmpty(graph), "Graph not specified.")
+
+			JSONtype = JSON_GetType(jsonID, jsonPath + "/0")
+			ASSERT(JSONtype == JSON_STRING, "first parameter needs to be a string labnotebook key")
+			str = JSON_GetString(jsonID, jsonPath + "/0")
```
t-b: Unrelated comment: The syntax highlightning seems to choke somewhere.

Packages/MIES/MIES_SweepFormula.ipf
```diff
@@ -718,6 +734,29 @@ Function/WAVE SF_FormulaExecutor(jsonID, [jsonPath, graph])
...
+			WAVE sweeps = SF_FormulaExecutor(jsonID, jsonPath = jsonPath + "/1", graph = graph)
+			ASSERT(DimSize(sweeps, COLS) < 2, "Sweeps are one-dimensional.")
+
+			WAVE numericalValues = GetLBNumericalValues(graph)
+			numIndices = DimSize(sweeps, ROWS)
+			Make/FREE/N=(numIndices) out = NaN
+			for(i = 0; i < numIndices; i += 1)
+				WAVE/Z value = GetLastSetting(numericalValues, sweeps[i], str, UNKNOWN_MODE)
+				if(WaveExists(value))
+					out[i] = value[0]
```
t-b: Nijet. GetLastSetting returns a wave with `LABNOTEBOOK_LAYER_COUNT` rows. The first 8 (aka `NUM_HEADSTAGES`) rows are for the headstage dependent settings, the last one are for the headstage independent settings. So we need to feed the info into the operation which headstages are meant.

I'll discuss with @timjarsky how the interface of the function should look like.

t-b: @ukos-git @timjarsky A quick discussion lead to the obvious `labnotebook("entry", channels(...), sweeps())`.

Packages/doc/labnotebook-docs.rst
ukos-git: I was thinking of the nwb file format, where the modification date is also present. Obviously a mistake. I restored the original formulation.

Packages/MIES/MIES_SweepFormula.ipf
ukos-git: I think, that is a design flaw in IP and strange behavior there.

```igorpro
String test = "012"
print test[0]
  0
print test[0,-1]
  
print test[0,0]
  0
print test[0,1]
  01
print test[1,0]
  
print test[1,1]
  1
print test[1,2]
  12
```

String range evaluation by character is inclusive on the left AND right end. If I had designed it, I would do it non-inclusive on the last element. And of course roll-over for negative values.

See for example python slicing:

```python
▶ python
Python 2.7.16 (default, Apr  6 2019, 01:42:57) 
[GCC 8.3.0] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> test="012"
>>> print test[0]
0
>>> print test[0:1]
0
>>> print test[0:0]

>>> print test[0:1]
0
>>> print test[0:2]
01
>>> print test[0:-1]
01
```

ukos-git: I think it is ok to leave the code like that.

t-b: Nope, sorry that is not really readable. It took me way too long to understand it.

ukos-git: The token needs to include the quotation mark `"string"` at this recursion step. Otherwise, white space are stripped, formulas inside the string will get evaluated etc..

The RegExp is straight forward and cannot get simpler than that. If this should get quicker evaluation, I could check the first and last character with `cmpstr`.

I need to remove Leading and Trailing `"` which we know is there if the `GrepString` is matched, so taking everything from the first character inside the `"string"` to the last character in the string is the easiest solution here.

I am open to suggestions.

t-b: Regarding the difficult to grasp code:

How about instead

```igorpro
JSON_AddString(jsonID, jsonPath, buffer[1, strlen(buffer) - 2]) 
```

something like

```igorpro
length = strlen(buffer)
if(length <= 2)
JSON_AddString(jsonID, jsonPath, ""]) 
else
JSON_AddString(jsonID, jsonPath, buffer[1, length - 2]) 
endif
```

?

t-b: Do we need to support escaped quotation marks in the strings? We don't have any LBN entries with " in them though.

> I am assuming that this only applies inside of a string token or must it also parse 1\+1 to a 
> string?

> I have tried my best but I cannot think of a proper way to implement this. 

I'm okay with leaving it as is if it is not a low hanging fruit.

ukos-git: > something like
> 
> length = strlen(buffer)
> if(length <= 2)
> JSON_AddString(jsonID, jsonPath, ""]) 
> else
> JSON_AddString(jsonID, jsonPath, buffer[1, length - 2]) 
> endif

A separation of zero length string handling would mean the following extension to the already existing `if` statement:

```igor
		elseif(!cmpstr(buffer, "\"\""))
			JSON_AddString(jsonID, jsonPath, "")
		elseif(GrepString(buffer, "^\".\+\"$"))
			JSON_AddString(jsonID, jsonPath, buffer[1, strlen(buffer) - 2])
```

I don't see the advantage of introducing a variable  `length` instead of `strlen()` here, so I skipped that part and the nested `if` statement.

I introduced that in 4517853. The unit test for the zero-length string remains untouched in the original commit c309c84. I have also added a comment to indicate that this is a non-functional change.

I hope it is more readable now.

ukos-git:
we probably need to introduce column labels here.

If sweep1 has 3 associated channels and sweep2 has 2 associated channels, then data is currently mixed. 

```
+--------+----------+----------+----------+
|        | channel1 | channel3 | channel9 |
+--------+----------+----------+----------+
| sweep1 |    data  |    data  |  data    |
+--------+----------+----------+----------+
| sweep2 |    data  |          |  data    |
+--------+----------+----------+----------+
| sweep3 |          |          |  data    |
+--------+----------+----------+----------+
```

It can even more complicated if one inputs `channels(AD1,AD)` 

Another problem is that we don't know the channel numbers for unassociated channels on input `AD`. So we need a function to return all unassociated channel number that are available for a specific channel type.

I find the data layout in MIES rather chaotic so this is as far as I can get here.

t-b:
@ukos-git As mentioned I'm already implementing a generic getter function for querying LBN entries based on the channel type and number.

ukos-git:
I could not see how this helps for this issue

ukos-git:
I am thinking of a combination of

```
FindDimLabel(waveName, dimNumber, labelString )
SetDimLabel dimNumber,dimIndex,label,wavelist
```
Luckily `FindDimLabel` outputs the last matching label and not the first.

ukos-git:
next steps:

Insert into correct column via matching channelnumber label to preserve channel association

* label name: channeltype + channelnumber → AD0, AD3,AD9 ...
* redimension wave to include `LABNOTEBOOK_LAYER_COUNT` additional columns
* backwards search for existing labels
* check if the found label is within the current range (greater than `DimSize(headstage, COLS) - LABNOTEBOOK_LAYER_COUNT`)

  + insert data at label position
  + otherwise: insert a new column and label it

→ remove `j`
→ use global counter for `numIndices`
→ rethink usage of `j` as it is not correct.

t-b:
@ukos-git The new labnotebook querying functions are now in master.

ukos-git:
I have added the new labnotebook functions and am using them to evaluate the `channels(AD)` input ot the actually active channels in the given set of sweeps. 

I generate a *unique* list of input channel types and numbers from the
`channels` and `sweeps` parameters. The output wave contains all
channels as columns and sweep numbers as rows. The column labels specify
the channel type and number.

```
+--------+----------+----------+----------+
|        | channel1 | channel3 | channel9 |
+--------+----------+----------+----------+
| sweep1 |    data  |    data  |  data    |
+--------+----------+----------+----------+
| sweep2 |    data  |          |  data    |
+--------+----------+----------+----------+
| sweep3 |          |          |  data    |
+--------+----------+----------+----------+
```

Missing fields are filled with `NaN`. This means that if i.e. `sweep3`
is only associated with `channel9` but `channel1` and `channel3` are
requested, the missing information is filled with `NaN`. As a usual
labnotebook entry is not allowed to contain `NaN`, such entries always
indicate that information is missing.

If the user requests `channels(AD,AD1)` and `AD1` is already contained
in `AD`, the channelNumber is ignored.

It is possible to mix channelTypes: `channels(DA,AD)`. In this case, the
sort order is AD,DA and follows the constants `ITC_XOP_CHANNEL_TYPE_ADC`
and `ITC_XOP_CHANNEL_TYPE_DAC`.

I need to update also the `data` function to get column labels and unique channels.

ukos-git:
```
max(
	data(
		[0,1e6],
		channels(AD),
		5...10
	)
)

vs

labnotebook(
	"set cycle count",
	channels(AD),
	5...10,
	DATA_ACQUISITION_MODE
)
```

Screenshot from 2019-11-22 15-14-14

ukos-git:
Screenshot from 2019-11-22 15-15-47

timjarsky:
@t-b my default expectation was that extracting `set sweep count` [1] from the lab notebook would overlay duplicate "set sweep count" values, but it appears to be taking into account some additional data to avoid duplicate "set sweep count" values:

image

actual `set sweep count` values for data displayed above:

image

I can see the way it currently plots as being useful, but I'd like to look into changing the default behavior.

[1]
```
labnotebook(
"set sweep count",
channels(AD0),
sweeps(),
DATA_ACQUISITION_MODE
)
```

ukos-git:
There is no "automatic" behavior. I could be mistaken but from my point of view, the two graphs that you provide and that show `set sweep count` look identically except for the marker/line mode.

If you select only one channel (`AD0`), you get a 1-column wave which is a 1-dimensional wave. 1-d waves consist of only one row. Plotting a wave with one row as y-values misses an explicit x-value. By default, the x-value is derived by Igor Pro from the point-scaling of the wave (or the x-scaling which is strict monotonous but not available here). You probably need to plot something against that wave to get the desired result.

For example

```
sweeps()
vs
labnotebook(
"set sweep count",
channels(AD0),
sweeps(),
DATA_ACQUISITION_MODE
)
```

Ideally, you would plot values that are alternating in the dataset against each other. I am unsure what causes the alternating `set sweep count` property but whatever that is should be plotted against `set sweep count` to display that dependency graphically.

t-b:
HS indep data needs fixing:

```
labnotebook(
"Optimized Overlap dDAQ",
channels(),
sweeps(), DATA_ACQUISITION_MODE
)
```

textual values don't work:

```
labnotebook(
"AD unit",
channels(AD0),
sweeps(), DATA_ACQUISITION_MODE
)
```

no cursors present being the same  as full range is not supported anymore:

```
	data(
		cursors(A,B),
		channels(DA0),
		sweeps()
	)
```

t-b:
@ukos-git Looks good. I've pushed to minor fixes (Wave/Z and docu).

The following things needs to be fixed

- [ ] Add documentation for text() and labnotebook() into rst and SweepFormula.ifn

- [ ] 7691822 merge with 6ac2 (which does not compile)

and that should probably also read more like

```
if(DimSize(wvY, ROWS) == 1                                         \
   || (DimSize(wvy, ROWS) < SF_MAX_NUMPOINTS_FOR_MARKERS           \
       && (!WaveExists(wvx)                                        \
           || DimSize(wvx, ROWS) <  SF_MAX_NUMPOINTS_FOR_MARKERS)))
    ModifyGraph/W=$win mode=3,marker=19
endif
```

and then we are ready :)

t-b:
@timjarsky You can test that here while Matthias is writing the documentation.

ukos-git:
> and that should probably also read more like
> if(DimSize(wvY, ROWS) == 1                                         \
   || (DimSize(wvy, ROWS) < SF_MAX_NUMPOINTS_FOR_MARKERS           \
       && (!WaveExists(wvx)                                        \
           || DimSize(wvx, ROWS) <  SF_MAX_NUMPOINTS_FOR_MARKERS)))
    ModifyGraph/W=$win mode=3,marker=19
endif

I think that code with the markers was completely obfuscated by 6b7ad02c0 and hard to read. I simplified it. It would have been good to not do the marker thing in two places at once.

```diff
@@ -1058,16 +1058,10 @@ Function SF_FormulaPlotter(graph, formula, [dfr])
 	// Redimension/N=(-1, dim1Y, dim2Y)/E=1 wvY
 	// Redimension/N=(-1, dim1X, dim2X)/E=1 wvX
 
-	axes = AxisList(win)
-	if(WhichListItem("bottomText", axes) != -1)
-		ModifyGraph/W=$win freePos(bottomText)={0,kwFraction}
-		ModifyGraph/W=$win mode=0
-	else
-		if(DimSize(wvy, ROWS) < SF_MAX_NUMPOINTS_FOR_MARKERS          \
-		   && (!WaveExists(wvx)                                       \
-		       || DimSize(wvx, ROWS) <  SF_MAX_NUMPOINTS_FOR_MARKERS))
-			ModifyGraph/W=$win mode=3,marker=19
-		endif
+	if(DimSize(wvY, ROWS) < SF_MAX_NUMPOINTS_FOR_MARKERS \
+		&& (!WaveExists(wvX) \
+		|| DimSize(wvx, ROWS) <  SF_MAX_NUMPOINTS_FOR_MARKERS))
+		ModifyGraph/W=$win mode=3,marker=19
 	endif
 ```

**Comment by timjarsky** [comment]
@t-b 

SF entry (see image) doesn't call labnotebook, but code does, and debugger opens:

image

pushing through (with `Go` button) produces expected output.


[originally placed this comment in the wrong PR]

**Comment by timjarsky** [comment]
@t-b 

`bracket mistmatch in formula` when using sample code from above:

```
labnotebook(
	"set cycle count",
	channels(AD),
	5...10,
	DATA_ACQUISITION_MODE
)
```

**Comment by timjarsky** [comment]
@t-b failure to parse labnotebook entry?

image

**Comment by t-b** [comment]
@timjarsky Something is buggy indeed. We'll investigate.

**Comment by ukos-git** [comment]
textualvalues is not needed for this loop. It is intentionally a null ref.

**Comment by t-b** [comment]
@ukos-git UnassociatedChannels_REENTRY in UTF_BasicHardwareTests.ipf needs a fix for the changed GetChannelSettings.

**Comment by timjarsky** [comment]
@t-b very nice:

image

**Comment by t-b** [comment]
@ukos-git 

One test fails:

```
ABORT: Encountered "Abort" in test case "UTF_SweepFormula#TestPlotting", procedure file "UTF_SweepFormula.ipf"
```

**Comment by ukos-git** [comment]
> @ukos-git
> 
> One test fails:
> 
> ```
> ABORT: Encountered "Abort" in test case "UTF_SweepFormula#TestPlotting", procedure file "UTF_SweepFormula.ipf"
> ```

```igorpro
Function BSP_HasBoundDevice(win)
	string win

	string device = BSP_GetDevice(win)

	return !(IsEmpty(device) || !cmpstr(device, NONE))
End
```

The `BSP_HasBoundDevice` handles empty device names. I therefore assume that an empty window should return an empty device? - Otherwise the ASSERT is triggered here which failed the tests.

ukos-git:
```
JSON_Version()
  {
  "XOP": {
    "builddate": "Oct 21 2019 13:50:02",
    "compiler": "Visual Studio 1916",
    "name": "JSON XOP",
    "version": "version-520-g357dbb9"
  },
  "igorpro": {
    "build": "34894",
    "freemem": "140731913498624",
    "igorkind": "pro64",
    "igorvers": "8.05",
    "nscreens": "1",
    "physmem": "8587755520",
    "screen1": "DEPTH=32,RECT=0,0,1920,2066",
    "usedphysmem": "539414528"
  },
  "operatingsystem": {
    "igorfileversion": "8.0.5.1",
    "locale": "US",
    "os": "Windows 10 Pro",
    "osversion": "6.3.17763"
  }
}
```

Packages/MIES/MIES_BrowserSettingsPanel.ipf
```diff
@@ -365,6 +365,9 @@ Function/S BSP_GetDevice(win)
 	string mainPanel
 
 	mainPanel = GetMainWindow(win)
+	if(IsEmpty(mainPanel))
+		return ""
+	endif
```
t-b: +1

t-b:
@ukos-git 

> I also assume that in our code an empty string never defaults tokwTopWin.

Yes!

ukos-git:
@timjarsky Could you post the message of the failing hardware tests? I cannot see them here.

t-b:
@ukos-git @timjarsky Let'me chim in before I'm gone for the holidays.

The error is

```
 Start of test "MIES with Hardware"
  Entering test suite "UTF_BasicHardwareTests.ipf"
  Entering test case "BasicHardwareTests#UnassociatedChannels:ITC"
  ITC Device opened, returned deviceID is 0.
  Entering reentry "BasicHardwareTests#UnassociatedChannels_REENTRY"
  "<<missing>>" == "pA" not respecting case: is false  Assertion "CHECK_EQUAL_STR(str, expectedStr)" failed in line 2373, procedure "UTF_BasicHardwareTests.ipf"
  "<<missing>>" == "pA" not respecting case: is false  Assertion "CHECK_EQUAL_STR(str, expectedStr)" failed in line 2373, procedure "UTF_BasicHardwareTests.ipf"
  "<<missing>>" == "pA" not respecting case: is false  Assertion "CHECK_EQUAL_STR(str, expectedStr)" failed in line 2373, procedure "UTF_BasicHardwareTests.ipf"
  "<<missing>>" == "mV" not respecting case: is false  Assertion "CHECK_EQUAL_STR(str, expectedStr)" failed in line 2383, procedure "UTF_BasicHardwareTests.ipf"
  "<<missing>>" == "mV" not respecting case: is false  Assertion "CHECK_EQUAL_STR(str, expectedStr)" failed in line 2383, procedure "UTF_BasicHardwareTests.ipf"
  "<<missing>>" == "mV" not respecting case: is false  Assertion "CHECK_EQUAL_STR(str, expectedStr)" failed in line 2383, procedure "UTF_BasicHardwareTests.ipf"
  Please be patient while we export all existing acquired content of all devices to NWB
```

and I fixed it locally at least in 46b6ab69.

@timjarsky We should have a chat after the break about getting #200 implemented.

t-b:
@ukos-git @timjarsky Rebased. I think this is ready to be merged.

timjarsky:
@t-b @ukos-git  looks great!

PR #346, opened 2019-10-30
Fix ni tp md caching

t-b:
My TP caching did not work with NI hardware.
I've fixed that now and also added the missing test which should have been part of the original PR.

### DIFF
--- a/Packages/MIES/MIES_DAC-Hardware.ipf
+++ b/Packages/MIES/MIES_DAC-Hardware.ipf
@@ -2036,6 +2036,7 @@ Function HW_NI_PrepareAcq(deviceID, [data, dataFunc, config, configFunc, flags, 
 		aiCnt = 0
 		ttlCnt = 0
 		for(i = 0;i < channels; i += 1)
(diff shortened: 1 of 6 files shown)


MichaelHuth:
I think that beside the test that the wave arrives in the cache, there should be a second test that starts the TP twice. This would test if the TP works when retrieved from the cache.

t-b:
@MichaelHuth 

> I think that beside the test that the wave arrives in the cache, there should be a second test that starts the TP a second time. This would test if the TP works when retrieved from the cache.

The test runs the same sweep three times so with the given large inter trial interval, we run the testpulse multiple times.
And that is checked in the REENTRY function as the HITS column of the cache stats wave has an entry larger or equal 1. And that means that we already fetched the wave once from the cache.

t-b:
@timjarsky I've restarted the tests. The reason for the failure was that the discs were running full on the windows 7 box. I'm now making space again.

t-b:
Next try without typo in GetNIDAQChannelWave().

t-b:
@timjarsky I've rebooted the windows 10 CI box. Now everything passes :)

PR #354, opened 2019-11-05
Remove some ip8 todos

t-b:


timjarsky:
@t-b does test coverage include moving items to trash? i.e. are we testing the TS trash? or do I should I manually test?

t-b:
@timjarsky Yes we do use moving waves to trash all the time in the tests.

PR #357, opened 2019-11-07
New utility function num2strEffective

MichaelHuth:
It is generally better than num2str

- Only returns effective number of digits in the string.
- Not limited to 5 digits.


t-b:
@MichaelHuth Assignment correct?

MichaelHuth:
A few more intense tests show that this approach does not work as well as thought due to representation of numbers when calculating base 10 on a base 2 represented number.

A working approach would be to just so a sprintf s, "%320.320f", num then cut after the 18th digit and remove trailing zeros.

PR #360, opened 2019-11-07
Document specialities about unassociated entries from more

t-b:
Unassociated DA/AD labnotebook entries are suffixed with
"UNASSOC_$channelNumber". At first sight this might indicate that the
entries are not distinct if you have multiple unassociated DA/AD
channels with the same channel number.

But it turns out that we don't have labnotebook entries which have the
same name for the DA and AD channels.

In order that it also stays like that we add a small test to check that.

- [x] 'UNASSOC_$channelNumber' -> 'u_$channelType_$channelNumber'
- [x] Add generic wrapper `GetLastSettingChannel` which returns the settings for given channel. Add support for both new and old style entries
- [x] Add documentation, mention in labnotebook.rst

### DIFF
--- a/Packages/doc/labnotebook-docs.rst
+++ b/Packages/doc/labnotebook-docs.rst
@@ -101,10 +101,9 @@ Caveats
   from the headstage independent layer only.
 * In case unassociated DA/AD channels were used (these are channels which are
   not connected to a headstage) during acquisition there are additional entries
-  present which are formatted like `$entry UNASSOC_$channelNumber` and only
-  have entries in the ninth layer (as it is by definition headstage
-  independent data). The entry names also don't overlap for AD/DA so having
-  only the channel number in the name is enough.
+  present which are formatted like `$entry UNASSOC_$channelNumber` (old style)
+  or `$entry u_(AD|DA)$channelNumber` (new style) and only have entries in the
+  ninth layer (as it is by definition headstage independent data).
 * One important concept is valid entries vs placeholder entries. All non-`NaN`
   or non-empty string entries are valid entries. Therefore only valid entries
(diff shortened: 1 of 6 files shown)


Packages/doc/labnotebook-docs.rst
```diff
@@ -99,10 +99,12 @@ Caveats
...
   the headstage independent layer (nineth layer). New code should query the data
   from the headstage independent layer only.
-* In case unassociated DA/AD channels were used during acquisition there are
-  additional entries present which are formatted like `$entry UNASSOC_$channelNumber`
-  and only have entries in the nineth layer (as it is by definition headstage
-  independent data).
+* In case unassociated DA/AD channels were used (these are channels which are
+  not connected to a headstage) during acquisition there are additional entries
+  present which are formatted like `$entry UNASSOC_$channelNumber` and only
+  have entries in the nineth layer (as it is by definition headstage
```
timjarsky: @t-b 
typo: ninth

I'm confused because the unassociated channel number in the experiment we looked at was a single digit. Would you help me out and show what unassocated DA0 and AD0 look like?

t-b: We can discuss that today.

t-b:
@timjarsky Ready.

timjarsky:
@t-b i tested with unassociated AD and DA channels. LBN entries look good. Anything else I should be looking for?

t-b:
@timjarsky Should be good from your side. But the test failures are real, I'll investigate.

t-b:
@timjarsky Should be good now. Some tests were still expecting the old naming scheme.

timjarsky:
@t-b tested again; looks good!

Issue #361, opened 2019-11-08
Fix flaky AbortTP test case


t-b:
Also with NI hardware the tests sometimes passes and sometimes fails.
@MichaelHuth Could you have a look?

PR #369, opened 2019-11-08
fix for AbortTP test failing

MichaelHuth:
DQ_ApplyAutoBias referred to non existing panel

When the DA_Ephys panel is killed while TP is running or TP is stopped then
on next TP start the remaining data is read out from Async.
This means the panel can already have been closed or reopened.

The fix skips the Auto Bias apply if the window is not present any more.

However, if in regular operation a user opens a new DA_Ephys panel and locks
it to the same device, DQ_ApplyAutoBias will work with the current GUI state
when referring to the old TP data.

closes https://github.com/AllenInstitute/MIES/issues/361

### DIFF
--- a/Packages/MIES/MIES_DataAcquisition.ipf
+++ b/Packages/MIES/MIES_DataAcquisition.ipf
@@ -242,10 +242,15 @@ Function DQ_ApplyAutoBias(panelTitle, BaselineSSAvg, SSResistance)
 	Wave BaselineSSAvg, SSResistance
 
 	variable headStage, actualcurrent, current, targetVoltage, targetVoltageTol, setVoltage
-	variable resistance, maximumAutoBiasCurrent
+	variable resistance, maximumAutoBiasCurrent, lastInvocation, curTime
+
+	if(DAP_DeviceIsUnlocked(panelTitle))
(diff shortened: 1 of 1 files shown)


MichaelHuth:
This fixes the failing test, but not the underlying problem that with regular usage on stopping TP there is a (high?) chance that the recent TP data remains in Async limbo.

Packages/MIES/MIES_DataAcquisition.ipf
```diff
@@ -242,10 +242,16 @@ Function DQ_ApplyAutoBias(panelTitle, BaselineSSAvg, SSResistance)
 	Wave BaselineSSAvg, SSResistance
 
 	variable headStage, actualcurrent, current, targetVoltage, targetVoltageTol, setVoltage
-	variable resistance, maximumAutoBiasCurrent
+	variable resistance, maximumAutoBiasCurrent, lastInvocation, curTime
+	string winExists = WinList(panelTitle, "", "")
+
+	if(IsEmpty(winExists))
```
t-b: You probably want `DAP_DeviceIsUnlocked(panelTitle)` here.

PR #370, opened 2019-11-08
Add enable/disable button to SF

### LINKED ISSUE #347: Add "enable" checkbox to SF tab
> can the SF tab have an "enable" like PA does so it doesn't continue to generate plots if you don't want it to?

ukos-git:
The SweepFormula gets an enable button and is disabled b default to
allow omitting calculations on DB update.

Closes #347

### DIFF
--- a/Packages/MIES/MIES_BrowserSettingsPanel.ipf
+++ b/Packages/MIES/MIES_BrowserSettingsPanel.ipf
@@ -588,6 +588,19 @@ Function BSP_SetPAControlStatus(win)
 	BSP_SetDeconvControlStatus(win)
 End
 
+/// @brief enable/disable the SF buttons
+///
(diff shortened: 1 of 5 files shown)


timjarsky:
@t-b I like that it is possible to edit the SF when disabled, but it would also be helpful if the SF tabs were enabled- is this possible?

t-b:
@timjarsky Yes this is certainly possible although it deviates from what we do for the other controls. We can chat about that tomorrow.

timjarsky:
>  it deviates from what we do for the other controls. 

@t-b In this case at least, I think it makes sense to mimic the `prevent update` checkbox in the waveBuilder.

t-b:
@ukos-git @timjarsky I've rebased  and fixed the conflicts. We handle the proposed changes wrt to the enable buttons in general and in a separate PR.

PR #374, opened 2019-11-15
P_PressureControl: Do nothing if the panel is not locked

t-b:
Followup fix from b6b031ab (fix for AbortTP test failing,
DQ_ApplyAutoBias referred to non existing panel, 2019-11-08).

This situation can happen if we called from TP_ROAnalysis which is the
readout function from the async framework.


t-b:
@MichaelHuth FYI a follow up fix for `AbortTP`.

PR #377, opened 2019-11-15
Bugfix/dont allow changed clamp mode during sweep

### LINKED ISSUE #376: Sweeps which were acquired in V-Clamp have some amplifier entries missing
Bug report:

> there's a new NWB file on your ftp server at NWB test data for bugfix/2019_10_02/2019_10_02_163453-compressed.nwb. If you look at sweep 71/AD0 (and a few others) you should find that the recording is in VC but has no VC holding potential in the lab notebook. I think we had encountered this issue long ago and you made a modification to increase logging in the hopes of tracking down the cause?

Looking at the relevant Labnotebook entries reveals the following (sweep 71, HS1):

grafik

"Operating Mode" is the clamp mode as read out by `AxonTelegraphGetDataStruct`, "Clamp Mode" is what we think in MIES the clamp mode is. So these two agree and say VC. But we don't have a "V-Clamp Holding Enable" entry. And that means that we took the elseif branch in 

https://github.com/AllenInstitute/MIES/blob/403c8ec284fa87095c27ea212cf91afc33d68e0f/Packages/MIES/MIES_AmplifierInteraction.ipf#L1458-L1490

And that means that the variable `clampMode` was not `V_CLAMP_MODE` (aka zero) but either I_CLAMP_MODE` or `I_EQUAL_ZERO_MODE`.

The labnotebook entry "Clamp Mode" is added to the sweep settings wave in 

https://github.com/AllenInstitute/MIES/blob/403c8ec284fa87095c27ea212cf91afc33d68e0f/Packages/MIES/MIES_DataConfiguratorITC.ipf#L1338

which is *before* the sweep.

TL;DR

The clamp mode changed between the begin and end of the sweep. This could be the case if the user changed the clamp mode in the MCC app instead of in MIES. I've just tried that in the referenced PR and could reproduce the issue in that way.

t-b:
Close #376.

### DIFF
--- a/Packages/MIES/MIES_AmplifierInteraction.ipf
+++ b/Packages/MIES/MIES_AmplifierInteraction.ipf
@@ -1455,6 +1455,10 @@ Function AI_FillAndSendAmpliferSettings(panelTitle, sweepNo)
 		clampMode = DAG_GetHeadstageMode(panelTitle, i)
 		AI_AssertOnInvalidClampMode(clampMode)
 
+		STRUCT AxonTelegraph_DataStruct tds
+		AI_InitAxonTelegraphStruct(tds)
+		AxonTelegraphGetDataStruct(axonSerial, channel, 1, tds)
+
 		if(clampMode == V_CLAMP_MODE)
 			ampSettingsWave[0][0][i]  = MCC_GetHoldingEnable()
 			ampSettingsWave[0][1][i]  = MCC_GetHolding() * AI_GetMCCScale(V_CLAMP_MODE, MCC_GETHOLDING_FUNC)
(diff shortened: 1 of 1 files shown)


Packages/MIES/MIES_AmplifierInteraction.ipf
```diff
@@ -1455,6 +1455,11 @@ Function AI_FillAndSendAmpliferSettings(panelTitle, sweepNo)
 		clampMode = DAG_GetHeadstageMode(panelTitle, i)
 		AI_AssertOnInvalidClampMode(clampMode)
 
+		STRUCT AxonTelegraph_DataStruct tds
+		AI_InitAxonTelegraphStruct(tds)
+		AxonTelegraphGetDataStruct(axonSerial, channel, 1, tds)
+		ASSERT(clampMode == tds.OperatingMode, "A clamp mode mismatch was detected. Please report that as a bug with a description what you did. Thanks!")
```
timjarsky: @t-b does this get run if Syncronize MIES to MCC is unchecked?

t-b: Yes this is always checked, "Syncronize MIES to MCC" does not play a role in that check here.

t-b: I'm not saying this is the perfect fix for the issue. I've asked luke already for the pxp so that I can have a look at the history. History saving did not work somehow, I've opened an issue for that.

timjarsky:
@t-b this works as expected. We still won't catch clamp mode changes like VC to IC and back to VC all within the same sweep, but I'm pretty sure I don't want to be waiting for MCC communication during a sweep. So, I think we have live with it.

PR #380, opened 2019-11-19
Enhance history storing in nwb

### LINKED ISSUE #375: Fix and check that history is saved into NWB
The file "NWB test data for bugfix/2019_10_02/2019_10_02_163453-compressed.nwb" on the FTP has an empty string for `/general/history`.

- [x] Investigate why this is the case
- [x] Add test to TestNWBExport() to ensure that it stays fixed

t-b:
Close #375.

I've made the possibility of exporting a NWB file without history much smaller.

We could also remove the `writeIgorHistory` parameter from `NWB_ExportAllData` alltogether and always store the experiment history. I would be in favour of that.


t-b:
@timjarsky Thanks, I'll wait for some replies to my slack post before merging. Maybe something related comes up.

PR #382, opened 2019-11-20
Fix ddaq display issues

t-b:
- [x] Add regression test with more stimsets to catch that as well
- [x] Devise a method to fixup buggy region data

@MichaelHuth Any ideas for the fixing up buggy region data? This seems to be quite complex in the general case.

### DIFF
--- a/Packages/MIES/MIES_Cache.ipf
+++ b/Packages/MIES/MIES_Cache.ipf
@@ -89,7 +89,7 @@ Function/S CA_DistDAQCreateCacheKey(params)
 		crc = WaveCRC(crc, params.preload)
 	endif
 
-	return num2istr(crc) + "Version 3"
+	return num2istr(crc) + "Version 4"
 End
 
 /// @brief Cache key generator for @c FindLevel in PA_CalculatePulseStartTimes()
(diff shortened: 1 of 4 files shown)


t-b:
@MichaelHuth I think I have something for fixing up buggy region data:
- Only the very first region is affected as we reset the wave reference `baseRegions` in `if(yoked...` on the first iteration
- buggy data is sorted but not compressed
- all the regions from the first region are duplicated in a buggy entry but offsetted by the offset of the second region

timjarsky:
@t-b works nicely

MichaelHuth:
@t-b I think I got a rough idea what the problem was, but we should probably phone about that briefly.

t-b:
@MichaelHuth Implemented fixup method looks good. Please have a look at it as well.

t-b:
@timjarsky We've added a fixup for buggy entries.

timjarsky:
@t-b tested on a file with corrupted LN data and it worked. Thank you for the fixup.

MichaelHuth:
> @MichaelHuth Implemented fixup method looks good. Please have a look at it as well.

Looks good. Just a minor comment, I think the comment description with "more than three" should be "more than two".

t-b:
@MichaelHuth Thanks for careful reading fixed.

PR #391, opened 2019-12-11
Feature/add support for notebook in set control event

t-b:
@timjarsky So adding support for setting notebook contents from SetControlEvent was the easy part.
Now I originally wanted to brag about how quickly this could be implemented.

But it turns out that this now fails at a more fundamental place.

If you add the following text as analysis parameter content

```
data(
cursors(A,B),
channels(AD),
sweeps()
)
```

we bug out with 

```
  Message: "Written entry contains invalid characters (one of `=:,;`)"
```

And that means that the formula can not hold these mentioned characters. Which makes it pretty useless I guess. The reason for that restriction is that we need to serialize the analysis parameters to the wave note, and not allowing these characters ensures that we can read it back properly.

A not-so-hacky solution to get that working soonish could be to allow the user to set a function returning a string with the formula instead of the formula? Other solutions could be to serialize the textwave differently to the wave note or to change the wave note format (JSON?!). Or we could define unicode replacement characters for `=:,;`.


timjarsky:
@t-b potentially silly question: can we have the parser ignore content in quotation marks?

t-b:
@timjarsky If we would have a real parser, yes we could do it. But we are just reading strings with various serialization schemes. I've chatted with Michael about that as well and one feasible option is to url encode the analysis params. This would only involve changing the wavebuilder text wave and a couple of parameter getter functions.

So our standard formula would then be encoded as

```
•print Urlencode("data(cursors(A,B),channels(AD),sweeps())")
  data%28cursors%28A%2CB%29%2Cchannels%28AD%29%2Csweeps%28%29%29
```

in the stimset wave note.

timjarsky:
> @timjarsky If we would have a real parser, yes we could do it. But we are just reading strings with various serialization schemes. I've chatted with Michael about that as well and one feasible option is to url encode the analysis params. This would only involve changing the wavebuilder text wave and a couple of parameter getter functions.
> 
> So our standard formula would then be encoded as
> 
> ```
> •print Urlencode("data(cursors(A,B),channels(AD),sweeps())")
>   data%28cursors%28A%2CB%29%2Cchannels%28AD%29%2Csweeps%28%29%29
> ```
> 
> in the stimset wave note.

@t-b so URL encoding handles replacement characters for us? I'm okay with this. The only downside is that the wavenote isn't human-readable anymore. I don't think humans try to read it often/at-all anyway.

t-b:
@timjarsky Yes exactly, URL encoding does what we would need to do, as is has a superset of the characters we don't allow. Yes the humand readability suffers, but it is difficult already I think.

timjarsky:
> @timjarsky Yes exactly, URL encoding does what we would need to do, as is has a superset of the characters we don't allow. Yes the humand readability suffers, but it is difficult already I think.

@t-b I think this is such a valuable feature and the tradeoff is insignificant. Let's go for it.

t-b:
@timjarsky A first version is pushed. I need to go over the code and tests tomorrow though.

t-b:
@timjarsky Ready for testing!

I also thought about making the analysis function parameter panel resizable and maybe use a notebook for the contents instead of a single line setvariable.

t-b:
@timjarsky I've done all of the todo except not showing the values with percent encoding in the listbox. This is not so easily possible at the moment as we currently read the list box values back, and therefore I need a faithful transformation.

But it is certainly usable for a first test.

timjarsky:
> @timjarsky I've done all of the todo except not showing the values with percent encoding in the listbox. This is not so easily possible at the moment as we currently read the list box values back, and therefore I need a faithful transformation.
> 
> But it is certainly usable for a first test.

@t-b it looks great; I love the new notebook in the param panel. In regards to the values, why not add a new human readable value column and stick the URL encoded off to the right?

Would you confirm that there isn't an easy way to resize the windows inside the param panel? Ideally, it would be helpful to be able to simultaneously change the horizontal size of the fields on the left side of the param panel while increasing the horizontal size of the table on the right side.

t-b:
> it looks great; I love the new notebook in the param panel. In regards to the values, why not add a new human readable value column and stick the URL encoded off to the right?

Yes I can do something like that. Thanks for pointing that out ;)

> Would you confirm that there isn't an easy way to resize the windows inside the param panel? Ideally, it would be helpful to be able to simultaneously change the horizontal size of the fields on the left side of the param panel while increasing the horizontal size of the table on the right side.

There are some options which I haven't yet ventured in. I'll give it a shot.

t-b:
> it looks great; I love the new notebook in the param panel. In regards to the values, why not add a new human readable value column and stick the URL encoded off to the right?

Yes I can do something like that. Thanks for pointing that out ;)

> Would you confirm that there isn't an easy way to resize the windows inside the param panel? Ideally, it would be helpful to be able to simultaneously change the horizontal size of the fields on the left side of the param panel while increasing the horizontal size of the table on the right side.

There are some options which I haven't yet ventured in. I'll give it a shot.

t-b:
@timjarsky First thing is fixed and for the second I have something which should do the job (and it was quick). What do you think?

timjarsky:
@t-b initial SF test formula isn't calling lab notebook but code is, and debugger is opening:

image

pushing through (Go button) does produce the expected output.

timjarsky:
@t-b this is working beautifully. Thank you! 

> @timjarsky First thing is fixed and for the second I have something which should do the job (and it was quick). What do you think?

Is the change here the addition of text wrapping? If yes, can it be upper left justified? Sorry if I've missed something obvious.

t-b:
@timjarsky The change was that the notebook for the analysis parameter contents and the listbox now have the same horizontal size and that is also kept while resizing.

t-b:
Changing the text wrapping in the listbox is not possible.

t-b:
@timjarsky IIRC I can merge if the tests pass?

PR #392, opened 2019-12-11
Require latest IP8 nightly

### LINKED ISSUE #390: Upgrade to newer IP8 nightly
The code

```
•make wv = {1}
•MatrixOP/P=1 out = sumCols((-1) * wv)^t
```

which is used in the sweep formula code returns `0` on 32bit IP.

Once we have a fix available let's require that IP8 version.

Edit:
The correct result being `-1`.

t-b:
This fixes a bug for IP 32bit where

make wv = {1}
MatrixOP/P=1 out = sumCols((-1) * wv)^t

gave 0 instead of -1.

And that is used in the sweep formula code itself and was uncovered by
one of the tests.

Close #390.

### DIFF
--- a/Packages/MIES_Include.ipf
+++ b/Packages/MIES_Include.ipf
@@ -3,13 +3,13 @@
 #pragma rtFunctionErrors=1
 
 // Igor Pro nightly installation:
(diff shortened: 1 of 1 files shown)


t-b:
@timjarsky Now I could just merge it, but that might make you miss the fact that we have *all* checks passing.

Issue #396, opened 2019-12-17
Configuration: Support multiple JSON config files which are applied to the same panel


t-b:
`XXX.json`  is the master file and `XXX_rig.json` is the per rig (aka user) file. Abort on duplicated entries.

Sample rig file:

```
{
    "Common configuration data": {
        "Headstage Association": {
            "0": {
                "Amplifier Serial": 0,
                "Amplifier Title": "0,1"
            },
            "1": {
                "Amplifier Serial": 0,
                "Amplifier Title": "0,1"
            }
        }
    }
}
```

Unfortunately the following does not work

```
Function DoStuff()
	variable jsonID1 = JSON_Parse("{ \"abcd\" : { \"a\" : 1 }}")
	variable jsonID2 = JSON_Parse("{ \"abcdefgh\" : { \"b\" : 1 }}")

	JSONXOP_AddValue/O/JOIN=(jsonID2) jsonID1, ""
	print/LEN=1000 JSON_Dump(jsonID1, indent=2)
End
```

as it gives

```
•dostuff()
  {
  "abcdefgh": {
    "b": 1
  }
}
```

We need a minimal solution.

MichaelHuth:
@t-b We miss a feature in our JSON_XOP that would be very useful for syncing two jsons.

lets have this source:
```
{
"array" : [1 , 2, [11, 12], "3"]
}
```
Let's assume array does not exist in the target and I am at "/array/2". I want to add an empty array at "/array/2" in my target. We don't have a function for that.

We need the JSON_AddValue/OBJ also for arrays.

MichaelHuth:
The only workaround would be to use JSON_Parse to parse some array template that contains in every cell empty arrays that are later replaced (except where actualy a sub array exists). The string for parsing need to be build recursively for each array root. The parsed json could then be joined in.

Considering this, it is much easier to patch in an /ARR parameter for JSON_AddValue that works the same as /OBJ just adding arrays.

t-b:
We found a work around for the missing JSON XOP feature.

MichaelHuth:
@t-b should the saved file name in the panels userdata the same as the main json file or main and rig file separated by pipe? (like "main.json|main_rig.json)

PR #397, opened 2019-12-18
Loading split configuration files

MichaelHuth:
The AutoLoader (Load Standard Configuration) is now aware of [Name]_rig.json files that complement [Name].json files with a panel configuration. 
The rig file support is reduced to work for for DA_Ephys panels only.

Changed StimSet path behavior for DA_Ephys configuration laoding:
- when path is "" then StimSet loading is skipped
- when path is set and file does not exist a message is shown
- when path is set and file exists it is loaded without dialog.

I added an adapted configuration using a rig file.

Due to the StimsetPath change it is required to either preLoad the default stimsets or adapt the stimset path to the local location of the default nwb file in 1_DA_Ephys.json. If not adapted the panel restoration will fail when it tries to set a Stimset that was not previously loaded.

closes https://github.com/AllenInstitute/MIES/issues/396

### DIFF
--- a/Packages/MIES/MIES_BackgroundWatchdog.ipf
+++ b/Packages/MIES/MIES_BackgroundWatchdog.ipf
@@ -14,6 +14,8 @@
 #include ":MIES_Structures"
 #include ":MIES_Async"
 #include ":MIES_Utilities"
+// JSON XOP
+#include "json_functions"
 
 /// @endcond
 
--- a/Packages/MIES/MIES_Configuration.ipf
+++ b/Packages/MIES/MIES_Configuration.ipf
@@ -168,6 +168,8 @@ static StrConstant EXPCONFIG_JSON_USERPRESSBLOCK = "User Pressure Devices"
 static StrConstant EXPCONFIG_JSON_USERPRESSDEV = "DAC Device"
 static StrConstant EXPCONFIG_JSON_USERPRESSDA = "Pressure DA"
 
+static StrConstant EXPCONFIG_RIGFILESUFFIX = "_rig.json"
+
 static Constant EXPCONFIG_MIDDLEEXP_OFF = 0
 static Constant EXPCONFIG_MIDDLEEXP_ON = 1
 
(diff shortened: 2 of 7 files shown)


MichaelHuth:
Ready for first tests.

t-b:
- [ ] Note to self: Check which file paths are written to the LBN.

t-b:
@timjarsky I've added the saving of the path and hash of the rig file.

timjarsky:
@t-b @MichaelHuth would it be possible to provide nicer messaging when the main config and rig config files have duplicate info?

Right now I get:
```
  !!! Assertion FAILED !!!
  Message: "Overwrite to non object attempted."
  Stacktrace:
CONF_AutoLoader(...)#L218 [MIES_Configuration.ipf]
CONF_RestoreWindow(...)#L314 [MIES_Configuration.ipf]
CONF_JoinRigFile(...)#L2122 [MIES_Configuration.ipf]
SyncJSON(...)#L4884 [MIES_Utilities.ipf]
SyncJSON(...)#L4884 [MIES_Utilities.ipf]
SyncJSON(...)#L4884 [MIES_Utilities.ipf]
SyncJSON(...)#L4884 [MIES_Utilities.ipf]
SyncJSON(...)#L4871 [MIES_Utilities.ipf]
  Time: 2020-01-07T16:23:27-08:00
  Locked device: [- none -]
  Current sweep: [- none -]
  DAQ: [- none -]
  Testpulse: [- none -]
  Experiment: Untitled.pxp
  Igor Pro version: 8.0.5.1 (34845)
  MIES version:
'''

Something like:
"Config files for the same GUI have conflicting specification data. Please remove the specification for 'Control name' from one of the config files."

MichaelHuth:
@timjarsky The recursive `SyncJSON` function is a generic function. I could however add a parameter that includes a description that is output in case a collision happens. In `CONF_JoinRigFile` I have the file name of the rig file. In `SyncJSON` I have the paths available where an object in the rig file collides with the main config file. Would a error messgae containing the rig file name and the colliding paths in both files sufficiently descriptive?

timjarsky:
@MichaelHuth 
Yes, a message with the file names and the colliding paths would be great. Can we also drop the assertion, or at least drop the assertion message? I ask because my impression is that we typically use assertion messaging as something that MIES developers need to attend to, and in this case, this is a user issue.

t-b:
@timjarsky Rebased to fix conflict. I've created https://github.com/AllenInstitute/MIES/issues/414 to track the requested changes.

t-b:
@timjarsky Approve and merge?

PR #409, opened 2020-01-06
Allow installing for all users with the silent installer

t-b:
- [x] Need to test silent installation
- [x] Need to be able to pass `/ALLUSER` to the installer


MichaelHuth:
I have not found any problems so far.

I have added an additional check in case the installer was built for user mode installation that prevents using /ALLUSER then.

t-b:
@MichaelHuth So this works on your end with and withour `/ALLUSER` in silent mode?

PR #412, opened 2020-01-07
ASYNC_Stop: Avoid debugger popping up on ThreadGroupWait errors

t-b:
Some users have reported that the debugger pops up when ThreadGroupWait
encounters a problem.

As this is not reproducible on this developers machine, let's just use
the standard solution and disable/reenable debug on error.


timjarsky:
>  I also could not reproduce a RTE there quickly.

@MichaelHuth @t-b I believe I most often see this error when I have multiple instances of Igor open for a while.

PR #417, opened 2020-01-13
Bugfix/config amplifier

### LINKED ISSUE #415: Include amp settings found in GetAmplifierParamStorageWave(panelTitle) in Common configuration data
At present, the amplifier settings are not being set on a per headstage basis (it looks like they are only set for headstage 0). Let's make it so the user can configure unique amp settings from config by including all the settings found in GetAmplifierParamStorageWave(panelTitle) in each headstage of the config file[1]

[1]
```
    "Common configuration data": {
        "Headstage Association": {
               "0": {
                      .....................
```

While doing this, create at least two subgroups in Headstage Association: pressure and amplifier

t-b:
- [x] Add write support
- [x] Add amplifier/pressure subgroups
- [x] Adapt tests
- [x] Cleanup code and commits

Close #415.


t-b:
@timjarsky Go for it!

Issue #403, opened 2019-12-21
Configuration: Rename folder `Settings` to `Settings_template`


t-b:
- [x] Add `.placeholder` in `Settings`
- [x] adapt tests

MichaelHuth:
There are no tests explicitly depending on the settings folder.

PR #422, opened 2020-01-14
Moved Settings folder to Settings_template

MichaelHuth:
- moved also default configuration files
- placed .placeholder in old Settings folder.

There are no tests that have hard coded dependency on the settings folder.

closes https://github.com/AllenInstitute/MIES/issues/403


Packages/MIES/MIES_Configuration.ipf
```diff
@@ -127,7 +127,7 @@ static Constant EXPCONFIG_UDATA_MAXCTRLARRAYINDEX = 100
 static Constant EXPCONFIG_JSON_INDENT = 4
 static StrConstant EXPCONFIG_FILEFILTER = "Configuration Files (*.json):.json;All Files:.*;"
 static StrConstant EXPCONFIG_CTRLGROUP_SUFFIX = " ControlGroup"
-static StrConstant EXPCONFIG_SETTINGS_FOLDER = "Settings"
+static StrConstant EXPCONFIG_SETTINGS_FOLDER = "Settings_template"
```
t-b: This variable should stay the same as we want require users to manually fill `Settings`.

MichaelHuth: Ok, so Settings should still be the actual configuration folder, whereas Settings_template should act as library storage (though inactive). I change that.

PR #427, opened 2020-01-31
Misc fixes

t-b:
@timjarsky Is good if CI passes.


timjarsky:
> @timjarsky Is good if CI passes.

@t-b I don't see this branch in bamboo.

PR #437, opened 2020-02-09
AddTraceToLBGraph: Avoid too long string error

t-b:
sprintf is still limited to some maximum length of the string inserted via
%s.


timjarsky:
@t-b Ummm, ITC1600 passed!?

PR #440, opened 2020-02-10
Documentation: Clarify Multiclamp installaton steps for CI machine

t-b:


Packages/doc/developers.rst
```diff
@@ -323,7 +323,10 @@ Setting up a continous integration server (Windows)
    -  Java 8
    -  Git (choose the installer option which will make the Unix tools
       available in cmd as well)
-   -  Multiclamp Commander (see above for specifics)
+   -  Multiclamp Commander (see above for specifics). After installing the
+      32-bit version, save the folder ``c:\Program Files (x86)\Molecular Devices``
```
timjarsky: @t-b is saving it somewhere sufficient or does the file also need to be removed from Program Files?

t-b: @timjarsky Saving somewhere is sufficient.

PR #462, opened 2020-02-18
Feature/switch to docker for documentation

t-b:


t-b:
@timjarsky http://bamboo.corp.alleninstitute.org/browse/MIES-PLAYG0-DD-2 passes so this is good to go.

PR #467, opened 2020-02-27
Tests: Fix failing json configuration tests

t-b:
In 38b90aba (Configuration: Introduce subgroups and add more amplifier
settings, 2020-01-30) some changes to the JSON configuration code was
added. This also changed the EOLS in the reference JSON files for testing.

And since then these tests are failing.

By normalizing the JSON documents, via parse and dump, the tests now
pass again.


t-b:
@timjarsky Sorry for cheating ;)

PR #471, opened 2020-02-27
added IVS_LoadStimSet to ivscc ipf as wrapper for NWB_loadallstimsets

aleonlein:


Packages/MIES/MIES_IVSCC.ipf
```diff
@@ -367,6 +367,16 @@ Function IVS_RunGigOhmSealQC()
 	CtrlNamedBackground IVS_finishGigOhmSealQCCheck, start
 End
 
+// @brief Loads a single stimulus for the user when using the ZMQ Proxy
```
timjarsky: @aleonlein missing forward slash

t-b: @timjarsky Right. But unfortunately I was too eager to merge it. I'll fix that.

aleonlein: @timjarsky  good catch - I will fix and resubmit.

t-b: @aleonlein I've already fixed it in 2471.

PR #480, opened 2020-03-11
Bugfix/more robust uncompile hook

t-b:


Packages/MIES/MIES_DataAcquisition.ipf
```diff
@@ -9,6 +9,27 @@
...
+	variable i, numDev
+	string device
+
+	SVAR devices = $GetDevicePanelTitleList()
+	numDev = ItemsInList(devices)
+	for(i = 0; i < numDev; i += 1)
+		device = StringFromList(i, devices)
+
+		try
+			ClearRTError()
```
timjarsky: @t-b what's giving rise the RT error (when a new experiment is being initiated)?

t-b: @timjarsky If you load old MIES experiments the window hook responsible for unlocking and cleaning up does bug out on old panels. This is expected. But pretty annoying with the new uncompile hook.

PR #481, opened 2020-03-11
Bugfix/safe new experiment calls via zeromq

### LINKED ISSUE #453: IVSCC: NewExperiment sometimes crashes/locks up IP when called with TP running over ZeroMQ XOP
At https://github.com/AllenInstitute/ZeroMQ-XOP/blob/6f0b4f41dfa9c1fcf03432c11219b30ec676f6c7/src/ZeroMQ.cpp#L39 we don't do anything suspicous, `CLEANUP` is according to the documentation not called on `NewExperiment`.

The reason is very likely that we stop and restart the zeromq message handler in https://github.com/AllenInstitute/MIES/blob/831ec8e323999de4d280a27ccaaf0ef4c7b06527/Packages/MIES/MIES_IgorHooks.ipf#L161

And another thing to fix is also that we don't stop DAQ/TP as the `NewExperiment` call is usually done with TP running.

t-b:
Needs verification by @aleonlein.

This should fix the hangs/crashes in MIES experienced in the new IVSCC setup.

Close #453.

### DIFF
--- a/Packages/MIES/MIES_MiesUtilities.ipf
+++ b/Packages/MIES/MIES_MiesUtilities.ipf
@@ -5849,3 +5849,14 @@ Function NewExperiment()
 
 	Execute/P/Q "NEWEXPERIMENT "
 End
+
+/// @brief Remove the volatile part of the XOP error code
+///
+/// The result is constant and can therefore be compared with constants.
+///
+///	 From http://www.igorexchange.com/node/7286
+threadsafe Function ConvertXOPErrorCode(xopError)
+	variable xopError
+
+	return xopError == 0 ? 0 : ((xopError & 0xFFFF) + 10000)
+End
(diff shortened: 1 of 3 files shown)


Packages/MIES/MIES_MiesUtilities.ipf
```diff
@@ -5849,3 +5856,14 @@ Function NewExperiment()
 
 	Execute/P/Q "NEWEXPERIMENT "
 End
+
+/// @brief Remove the volatile part of the XOP error code
```
timjarsky: @t-b what does volatile mean in this context?

t-b: @timjarsky It should mean something like "moving"/"flexibel"/"non-constant". Does that make sense?

t-b:
@timjarsky Do we want to wait for testing from IVSCC?

aleonlein:
I will be able to test this next Monday. I understand if you don’t want to wait

t-b:
@aleonlein Next monday sounds good!

aleonlein:
@t-b Apologies for the delay but I will be testing this today

aleonlein:
@t-b I tested this out with my message handler and I am not experiencing any crashes.

t-b:
@aleonlein Thanks! So I'll merge.

Issue #483, opened 2020-03-13
Function TS_ThreadGroupPutDFR might crash Igor Pro.


MichaelHuth:
The function has the problems:
- it does have no protection against dangling references to objects in the given dfr.
- dfr itself is a reference to the datafolder pushed to the thread
- The function exits with the current datafolder set that was just pushed.

We solved that in ASYNC_putDF by duplicating and setting the data folder through $name and setting it back afterwards.

t-b:
@MichaelHuth Can you make a PR? Thanks.

PR #484, opened 2020-03-13
Fix for TS_ThreadGroupPutDFR

MichaelHuth:
fixes https://github.com/AllenInstitute/MIES/issues/483
requires: https://github.com/AllenInstitute/MIES/pull/485

also adapted ASYNC framework to use the threadsafe utility function instead of its own.

### DIFF
--- a/Packages/MIES/MIES_Constants.ipf
+++ b/Packages/MIES/MIES_Constants.ipf
@@ -141,7 +141,7 @@ Constant EVENT_KILL_WINDOW_HOOK = 2
 /// Used by CheckName and UniqueName
 Constant CONTROL_PANEL_TYPE = 9
 
-/// @name CountObjects and CountObjectsDFR constant
+/// @name CountObjects, CountObjectsDFR, GetIndexedObjName, GetIndexedObjNameDFR constants
 /// @anchor TypeFlags
 /// @{
 Constant COUNTOBJECTS_WAVES      = 1
--- a/Packages/MIES/MIES_ThreadsafeUtilities.ipf
+++ b/Packages/MIES/MIES_ThreadsafeUtilities.ipf
@@ -162,17 +162,26 @@ End
 
 /// @brief Push a datafolder to the thread queue
 ///
-/// Accepts a free datafolder for `dfr` unlike `ThreadGroupPutDF`.
+/// Works on a duplicate of the input DF to remove all references.
+/// dfr can be a free DF
 threadsafe Function TS_ThreadGroupPutDFR(tgID, dfr)
 	variable tgID
 	DFREF dfr
 
-	string dataFolder
+	string dfrName
 
 	ASSERT_TS(DataFolderExistsDFR(dfr), "ThreadGroupPutDFR: dfr does not exist")
 
-	SetDataFolder dfr
-	ThreadGroupPutDF tgId, :
+	DFREF dfrSave = GetDataFolderDFR()
+
+	SetDataFolder NewFreeDataFolder()
+	DuplicateDataFolder/Z dfr, :
+	ASSERT_TS(!V_flag, "ThreadGroupPutDFR: Could not duplicate data folder")
+	dfrName = GetIndexedObjName(":", COUNTOBJECTS_DATAFOLDER, 0)
+
+	ThreadGroupPutDF tgID, $dfrName
+
+	SetDatafolder dfrSave
 End
 
 /// @brief Returns 1 if all worker threads have finished
--- a/Packages/MIES/MIES_Async.ipf
+++ b/Packages/MIES/MIES_Async.ipf
@@ -150,7 +150,7 @@ threadsafe static Function ASYNC_Thread()
 
 		endif
 
-		ASYNC_putDF(dfrAsync, 0)
+		TS_ThreadGroupPutDFR(0, dfrAsync)
 		KillDataFolder dfr
 	endfor
 End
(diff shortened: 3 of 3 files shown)


Packages/MIES/MIES_ThreadsafeUtilities.ipf
```diff
@@ -165,11 +165,18 @@ threadsafe Function TS_ThreadGroupPutDFR(tgID, dfr)
...
-	SetDataFolder dfr
-	ThreadGroupPutDF tgId, :
+	dataFolder = UniqueDataFolderName($":", "temp")
+	DuplicateDataFolder dfr, $dataFolder
+	dfrSave = GetDataFolderDFR()
+
+	SetDataFolder $dataFolder
+	ThreadGroupPutDF tgID, :
+
+	SetDatafolder dfrSave
```
timjarsky: @t-b the next time we chat (no rush) may we walk through the rationale some of the see lines of code? - it's for my own edification.

t-b: @timjarsky Sure!

MichaelHuth: @t-b @timjarsky I am sorry, but I had to fix another flaw. So the code has changed slightly.

The main change is that before
`	dataFolder = UniqueDataFolderName($":", "temp")
`
a
`	SetDataFolder root:
`
is added.

If we don't do that and the current datafolder is a free data folder when the function is called, then UniqueDataFolderName returns "<number>:temp", where number is the free data folder id. However such DF name is not usable by `DuplicateDataFolder`. (would need just ":temp")
If DuplicateDataFolder fails here then the function crashes Igor. This is because then the `SetDataFolder `in front of `ThreadGroupPutDF `fails as well. So `ThreadGroupPutDF `operates on the current DF and not on the duplicated. As there are usually some references in use, Igor crashes.

So I added a try catch if DuplicateDataFolder fails, to stop execution in case before Igor crashes.

Secondary consideration:

UniqueDataFolderName should probably ASSERT if dfr is a free data folder as the returned DF path is not usable then.

What do you think?

t-b: > UniqueDataFolderName should probably ASSERT if dfr is a free data folder as the returned 
> DF path is not usable then.

Yes let it assert out. I don't think it makes sense to be used with a free datafolder.

t-b: > SetDataFolder root: 

Can we only do that in the case we need to? Our general approach, which worked really well, is that we only use `SetDataFolder` if there is no other way. IIRC `DataFolderRefStatus` can tell you if the current datafolder is a free data folder.

MichaelHuth: @t-b From you comment in the chat recently I thought about a way to use generally a free datafolder to duplicate into, which would let us get rid of UniqueDataFolderName in this code part.

So I reworked the whole function.

Maybe there is even a faster way to find out if dfr is the root of a free DF?

t-b: Looks good!

t-b: @MichaelHuth I looked into that again with Tim and how about using `GetIndexedObjName` instead of manually deriving dfrName?

MichaelHuth: You mean instead of `dfrName = "freeroot"` ?

Thats given by Igor and not exactly a manual derivation.

t-b: @MichaelHuth yes!

MichaelHuth: Ok, though conceptionally I think that such constants could be derived at MIES startup and reused.

t-b: The whole drfName derivation can be replaced by `GetIndexedObjName` or?

MichaelHuth: Yes, just pushed that. It assumes however that the free to free datafodler duplicate still results in a new datafolder. It also uses the fact that for no DF found an empty string is returned.

Packages/MIES/MIES_Constants.ipf
```diff
@@ -1223,6 +1223,15 @@ Constant WINTYPE_CAMERA = 15
...
 /// @}
 
+/// @name objectType codes of the Igor GetIndexedObjectName and GetIndexedObjectNameDFR function
+/// @anchor GetIndexedObjectName_ObjectTypes
+/// @{
+Constant OBJTYPE_WAVES = 1
+Constant OBJTYPE_NUMVARS = 2
+Constant OBJTYPE_STRVARS = 3
+Constant OBJTYPE_DATAFOLDERS = 4
+/// @}
```
t-b: We have these already as `COUNTOBJECTS_WAVES` and friends.

MichaelHuth: ah ok

Packages/MIES/MIES_ThreadsafeUtilities.ipf
t-b: @MichaelHuth Nice!

PR #485, opened 2020-03-13
Removed non required ASYNC include in Backgroundwatchdog

MichaelHuth:
In independent module Backgroundwatchdog ASYNC was included but not used.
This prevented compilation when ASYNC used function that are not included in
BackgroundWatchdog as well.


t-b:
@timjarsky Good if CI passes.

PR #493, opened 2020-03-19
Make MIES much faster with > 100 sweeps

### LINKED ISSUE #165: Speedup acquistion and make it work with many (> 1000) sweeps

t-b:
Close #165.


t-b:
@timjarsky Ready for testing.

timjarsky:
@t-b in this branch, but not master, the following error occurs when using automated configuration.

image

t-b:
@timjarsky Thanks. The error message is heavily misleading, but the error is fixed. The matches wave must not be empty at this point.

timjarsky:
@t-b What are your thoughts on:

image

timjarsky:
@t-b without DB plotting, this PR speeds up data collection nicely with hundreds to >1000 sweeps. I propose we merge, and tack DB plotting speed separately.

timjarsky:
> @t-b What are your thoughts on:
> 
> image

@t-b this gets called when the DB is plotting the next sweep (at least that is what I was doing at the time).

t-b:
We call this function *once* before DAQ so this is suprising to see it take 10%.

PR #494, opened 2020-03-20
DRAFT: Add Labnotebook entry descriptions

t-b:


timjarsky:
@t-b What if we added another row to the LN key wave; group. We'd then use this group assignment for each entry to categorize the LN menu. Thoughts?

t-b:
@timjarsky Sounds good. But we would merge #507 before we finish this PR or?

timjarsky:
> @timjarsky Sounds good. But we would merge #507 before we finish this PR or?

In a new PR, I could categorize each column of the LN relatively quickly. Options are: merge #507, and then create a new PR that uses the new LN entries OR wait on #507 and adapt it to use entries when they are available. I suspect you (@t-b ) favor the former?

t-b:
@timjarsky Jeep I favour the former ;)

Issue #500, opened 2020-03-20
TS_ThreadGroupPutDFR: Recent changes trigger IP bug


t-b:
From my testing this is due to the changes in 4b4a965b (fix possible crash in TS_ThreadGroupPutDFR, 2020-03-17).

The top root node is missing after running the TP.

![grafik](https://user-images.githubusercontent.com/1185677/77206300-e5e96a00-6af6-11ea-9c39-ac720d9e357d.png)

t-b:
@MichaelHuth Can you have a look?

MichaelHuth:
@t-b 
I tested a bit and using
`print GetDataFolder(1), SetDataFolder and DataFolderDir(-1)`
the state and tree of the folders looks ok.

The problem seems to be with the Data Browser which shows a wrong tree view and after reopening just empty and no other objects.

PR #501, opened 2020-03-23
Fix for data browser showing empty root data folder after ThreadGroupPutDF

MichaelHuth:
this bug was introduced in commit 7b2cd9b9c4ae6d251616dfa13c974a35741f9511

After using any functionality that pushed threaded workloads the data browser
root started to show an empty name. Also the data browser did not update any
more.

closes https://github.com/AllenInstitute/MIES/issues/500

### DIFF
--- a/Packages/MIES_Include.ipf
+++ b/Packages/MIES_Include.ipf
@@ -3,13 +3,13 @@
 #pragma rtFunctionErrors=1
 
 // Igor Pro nightly installation:
-// - Download from https://www.byte-physics.de/Downloads/WinIgor8_24FEB2020.zip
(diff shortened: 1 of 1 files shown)


MichaelHuth:
I have set the latest nightly as the build that still has the bug. In case WM releases more nightlies with the bug unfixed then the build number check must be adjusted.

MichaelHuth:
I have also verified that I don't see the bug any more in build 35432.

t-b:
@MichaelHuth Thanks, I'll take it from here.

t-b:
We are waiting for another bugfix for DuplicateDataFolder.

PR #502, opened 2020-03-26
IVS_GetSetQCForSweep: Add convenience function for querying Set QC passed/failed state

### LINKED ISSUE #410: IVSCC: Add custom routine to query pass/fail state of a RA cycle (Set QC)
The information from the dashboard is basically required.

t-b:
Close #410.

Functionality requested by @aleonlein.

### DIFF
--- a/Packages/MIES/MIES_IVSCC.ipf
+++ b/Packages/MIES/MIES_IVSCC.ipf
@@ -537,3 +537,39 @@ Window IVSCCControlPanel() : Panel
 	Button button_runBaselineQC,pos={48.00,61.00},size={190.00,30.00},proc=IVS_ButtonProc_BaselineQC,title="Run baseline QC"
 	Button button_runAccessResisQC,pos={48.00,145.00},size={190.00,30.00},proc=IVS_ButtonProc_AccessResist,title="Run access resistance QC check"
 EndMacro
+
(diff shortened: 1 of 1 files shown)


Packages/MIES/MIES_IVSCC.ipf
```diff
@@ -537,3 +537,39 @@ Window IVSCCControlPanel() : Panel
 	Button button_runBaselineQC,pos={48.00,61.00},size={190.00,30.00},proc=IVS_ButtonProc_BaselineQC,title="Run baseline QC"
 	Button button_runAccessResisQC,pos={48.00,145.00},size={190.00,30.00},proc=IVS_ButtonProc_AccessResist,title="Run access resistance QC check"
 EndMacro
+
+/// @brief Return the Set QC passed/failed state for the given sweep
+///
+/// @return 1 if passed, 0 if not (or not yet) and
+/// asserts out on all other errors.
```
timjarsky: @t-b Do we want assertions in this function? I'm reluctant to have a call from the work sequence engine turn off whatever the user is doing at that time. What about returning an error message to the caller?

t-b: Calling a function via zeromq XOP which asserts out does not stop anything running in IP. So switchting to just return errors is giving only less clutter in the history.

PR #504, opened 2020-03-26
DAP_CheckSettings: Avoid using async channels for DAQ as well

### LINKED ISSUE #405: DAP_CheckSettings: Check async channels for double usage
Channels used for async acquisition should not be used in the AD tab.

t-b:
Close #405.


timjarsky:
@t-b remember that the ITC1600 has dedicated async channels and therefore they can share AD names.

t-b:
@timjarsky Fixed.

timjarsky:
Tested with ITC1600 and ITC18_USB using AD0 for asyn and AD0 for HS0. Desired behavior was observed with both devices [ITC18USB acq was aborted; 1600 acq proceeded normally].

PR #505, opened 2020-03-30
Show headstage in databrowser/sweepbrowser as well

### LINKED ISSUE #122: DataBrowser/AnalysisBrowser Add option to show headstage on Y-axis labels
The Y-axis label in the MIES browsers presently displays the AD channel. I'd like to give the user the option to display the headstage number instead.

t-b:
Close #122.

How about not adding a setting but showing both? This is what I implemented now.

### DIFF
--- a/Packages/MIES/MIES_AnalysisBrowser_SweepBrowser.ipf
+++ b/Packages/MIES/MIES_AnalysisBrowser_SweepBrowser.ipf
@@ -358,6 +358,7 @@ Function SB_UpdateSweepPlot(win, [newSweep])
 
 	WAVE/T cursorInfos = GetCursorInfos(graph)
 	RemoveTracesFromGraph(graph)
+	RemoveFreeAxisFromGraph(graph)
 
 	WAVE/T sweepMap = SB_GetSweepBrowserMap(sweepBrowserDFR)
 	WAVE channelSel = GetChannelSelectionWave(sweepBrowserDFR)
--- a/Packages/MIES/MIES_DataBrowser.ipf
+++ b/Packages/MIES/MIES_DataBrowser.ipf
@@ -238,6 +238,7 @@ Function/S DB_ClearAllGraphs()
 
 		if(WindowExists(graph))
 			RemoveTracesFromGraph(graph)
+			RemoveFreeAxisFromGraph(graph)
 		endif
 	endfor
 End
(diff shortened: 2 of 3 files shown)


timjarsky:
@t-b is there a way to resize the text automatically so that Y-axis labels don't overlap?

t-b:
@timjarsky Not really. I've played around some more but the only option is either to make the axis labels smaller or show only one of them. I've opted for the latter open.

timjarsky:
@t-b I don't like having the headstage label settings checkbox. Vertical space, but not horizontal space, is limiting. Why not something like:

image
```
HS0
\Z08DA0
(mV)
```

timjarsky:
@t-b one more note: when AD and DA channels are displayed, there are two identical HS labels. In that case, can the DA and AD share a HS axes label?

t-b:
Note to self: Check if it is feasable to have only one HS label when two adjacent AD/DA channels are shown.

t-b:
@timjarsky One can have one axis label for the DA/AD channels here is what needs to be done:
- Figure out in CreateTiledChanneLGraph when two DA/AD channels are adjacent and belong to the same headstage
- Calculate the center of the spanned vertical area length
- Position the axis label there using lblPosMode with margin scaled

Not very straightforward. Now? New Issue? Forget?

timjarsky:
> @timjarsky One can have one axis label for the DA/AD channels here is what needs to be done:
> 
> * Figure out in CreateTiledChanneLGraph when two DA/AD channels are adjacent and belong to the same headstage
> * Calculate the center of the spanned vertical area length
> * Position the axis label there using lblPosMode with margin scaled
> 
> Not very straightforward. Now? New Issue? Forget?

@t-b I think we should take the time to do the above. Will it be possible to restrict this calculation to times where the channel configuration is modified or does it need to be done with each newly plotted sweep?

t-b:
@timjarsky We build the plot from scratch each time.

t-b:
@timjarsky I've added something :)

timjarsky:
@t-b resizing the DB changes the relative position of the label such that it can disappear off the graph. If possible, can the HS label be outside the channel label?

t-b:
@timjarsky You beat me to it ;) Now enhanced with using a free axis.

timjarsky:
> @timjarsky You beat me to it ;) Now enhanced with using a free axis.

@t-b it's looking good. Can the HS label be moved outside of the channel labels and vertically centered between them?

t-b:
Note to self: It should look like 
grafik

t-b:
@timjarsky Go for it!

timjarsky:
@t-b when using >2 HS I get the error captured in the image below.

image

t-b:
@timjarsky Next try

Issue #228, opened 2019-06-26
Avoid very long popup menues


t-b:
Very long popup menues are difficult to navigate. Currently there is no builtin support for auto grouping.

But Adam posted someting at https://www.wavemetrics.com/comment/18551#comment-18551 which works quite well.

![bandicam 2019-06-26 21-12-05-751](https://user-images.githubusercontent.com/1185677/60207819-15fd2880-9857-11e9-97b7-38d4a73a6653.jpg)

The names of the menu entries can be dynamic as well, this was done until 1f5693da (Always include MIES_TangoInteract, 2018-02-20) for the tango interact functionality.

The real problem to solve are the labnotebook entry lists for the databrowser. The two popup menues can also be changed to one.

MichaelHuth:
@t-b So I found a crash in combination with the dynamical generated menus and reported it to WM.
I also found that `PopupContextualMenu/N/ASYN=Callback` does not work. It would be exactly the combination we would need though. The other /ASYN variants are not usable for dynamicly generated menus. I also wrote WM an email regarding this, since `PopupContextualMenu/N/ASYN=Callback` is not really documented and we could use that feat.

t-b:
- [ ] Don't use GetCurrentWindow()
- [ ] Unify both popup menues for numerical and text

Extract 

```
   col = FindDimLabel(labnotebookValues, COLS, setting)

   if(col < 0)
	   return $""
   endif
```
from `GetNonEmptyLBNRows` into its own function and use that in DB_PopMenuProc_LabNotebook.

MichaelHuth:
[ ] PEXT_SplitToSubMenus should take a wave, convert list to wave in button proc for default FuncRef
[ ] GetLabNotebookSortedKeys should return wave, as well as DB_GetLBTextualKeys/DB_GetLBNumericalKeys/DB_GetLBKeys
[ ] Change Analysisbrowser LBN_GetLBTextualKeys/LBN_GetLBNumericalKeys to use function that returns WAVE with keys, use TextWaveToList there

PR #507, opened 2020-03-30
Extended PopupMenus procedures to replace long popup menus

MichaelHuth:
The procures try to keep a compatible upgrade, where long popupmenus
get replaced by a button and previous proc definition move to the
buttons userdata.
After the menu finishes the old proc gets called as if it would be a
popupmenucontrol.

See SetupPopupMenuExt() for demo code.

closes https://github.com/AllenInstitute/MIES/issues/228

### DIFF
--- a/Packages/MIES/MIES_GuiPopupMenuExt.ipf
+++ b/Packages/MIES/MIES_GuiPopupMenuExt.ipf
@@ -1,0 +1,633 @@
+#pragma TextEncoding = "UTF-8"
+#pragma rtGlobals=3 // Use modern global access method and strict wave access.
+#pragma rtFunctionErrors=1
+
+#ifdef AUTOMATED_TESTING
+#pragma ModuleName=MIES_GUIPOPUPEXT
+#endif
+
+/// @file MIES_GuiPopupMenuExt.ipf
+/// @brief Helper functions related to GUI controls
+///
+/// @anchor PopupMenuExtensionShortDescription
+///
+/// These procedures provide a way to replace popup menus with many entries with context menus.
+/// Context menus can provide submenus, that allow better organization of menu entries.
+///
+/// Example:
+///
+/// \rst
+/// .. code-block:: igorpro
+///
+///    Function SetupPopupMenuExt()
+///        KillWindow/Z panel0
+///        NewPanel/N=$"panel0"/K=1
+///        Button popupext_menu1, pos={3.00, 20.00}, size={200,20},proc=PEXT_ButtonProc,title="DropDownMenu1 ▼", userdata=(PEXT_UDATA_ITEMGETTER + ":GetMenuList1;" + PEXT_UDATA_POPUPPROC + ":DemoProc")
+///        Button popupext_menu2, pos={3.00, 55.00}, size={200,20},proc=PEXT_ButtonProc,title="DropDownMenu2 ▼", userdata=(PEXT_UDATA_ITEMGETTER + ":GetMenuList2;" + PEXT_UDATA_POPUPPROC + ":DemoProc")
+///        Button popupext_menu3, pos={3.00, 90.00}, size={200,20},proc=PEXT_ButtonProc,title="DropDownMenu3 ▼", userdata=(PEXT_UDATA_ITEMGETTER + ":GetMenuList3;" + PEXT_UDATA_POPUPPROC + ":DemoProc")
+///    End
+///
+///    Function DemoProc(pa) : PopupMenuControl
+///        STRUCT WMPopupAction &pa
+///
+///        print "EventCode/Win/Selected/ctrlName: ", pa.eventCode, pa.win, pa.popStr, pa.ctrlName
+///    End
+///
+///    Function/S GetMenuList1()
+///
+///        Make/FREE/T/N=300 testItems = num2char(97 + mod(p, 26)) + "_" + num2str(p)
+///        return TextWaveToList(testItems, ";")
+///    End
+///
+///    Function/WAVE GetMenuList2(panelTitle)
+///        string panelTitle
+///
+///        Make/FREE/T/N=(2) menus
+///        menus[0] = "Kriemhild;Brünhild"
+///        menus[1] = "Siegfried;Gunther"
+///        SetDimLabel ROWS, 0, $"Women", menus
+///        SetDimLabel ROWS, 1, $"Men", menus
+///
+///        return menus
+///    End
+///
+///    Function/WAVE GetMenuList3(panelTitle)
+///        string panelTitle
+///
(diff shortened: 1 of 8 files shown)


MichaelHuth:
@t-b See SetupPopupMenuExt() procedure for the button replacement that should replace existing long popupmenus as glue code. The non-blocking Async functionality is not working with dynamic menus. I wrote a bit in the issue regarding that. We can also discuss that on the phone.

t-b:
@MichaelHuth I've chatted with Tim about that and we have one feature request. Can we allow the user to define the categories and which entries go into which categories? E.g. for the databrowser LBN entries it makes more sense to group them by things like "Amplifier"/ "Testpulse" than by first letter.  On way to do that could be to have `GetMenuList1()` return [wave/T elements] where elements is a nx2 wave with the entry in the first row and the category in the second. You could even fall back to the current grouping if the wave has only one column.

MichaelHuth:
@t-b 
Ok, this means I have to reduce the category row to the non-duplicates to determine the number of submenus. Like:
```
wave[0][0] = "tea"
wave[0][1] = "drinks"
wave[1][0] = "coffee"
wave[1][1] = "drinks"
[....]
wave[4]]0] = "bread"
wave[4]]1] = "food"
[....]
```



Another suggestion: How about a 1D wave where each element contains a semicolon separated list for a submenu and the DimLabel is the submenu name.

Like:
```
wave[0] = "tea;coffee;water" // with DimLabel drinks
wave[1] = "bread;rice;potatoe" // with DimLabel food

```

t-b:
> Another suggestion: How about a 1D wave where each element contains a semicolon separated list for a submenu and the DimLabel is the submenu name.

If you think this is better we can have that as well.

MichaelHuth:
Ok, due to the Duplicate resolution being rather inefficient to do in the menu generation functions for each submenu creation, I will move that to the buttonproc. So the itemListWave will have a format that is directly resolvable to menus in the two creation functions.
This is also important since the Callback gets not called if the user aborts a menu selection by clicking somewhere else. For that case the wave note gets not cleared, so that for all further menu clicks our creation functions will NOT do an early abort.

I currently have no idea how to catch that selection abort.

Packages/MIES/MIES_GuiPopupMenuExt.ipf
```diff
@@ -0,0 +1,623 @@
...
+/// When a user selects an entry in the context menu the PopupMenuControl control defined through PEXT_UDATA_POPUPPROC is called.
+/// So the proc of the previous popupmenu can be used without further adaptations.
+///
+/// User defined sub menus:
+///
+/// User defined submenus can be created as well. Therefore an extension was added to the function that returns the menu items.
+/// Instead of a string as described before a function can be defined that returns a 1D text wave, e.g. GetMenuList2() in the example.
+/// Each element contains a semicolon separated list of menu items. Each rows DimLabel is used as submenu name, where the elements
+/// menu items are put. The DimLabels must not be empty. The wave can have up to MAX_SUBMENUS elements.
+/// The button popupext_menu2 shows this extension in the example.
```
timjarsky, line 81: @MichaelHuth @t-b This looks fantastic. Thank you! I would like to implement the user defined lists for the labnotebook but I'm not sure what would happen if GetMenuList2() has list items that don't yet exist in the labnotebook?

Packages/MIES/MIES_GuiPopupMenuExt.ipf
```diff
@@ -0,0 +1,623 @@
...
+	else
+		ASSERT(0, "Unknown method for automatically splitting to submenus")
+	endif
+
+	return splittedMenu
+End
+
+/// @brief Automatically generates submenu names and applies them on the given wave
+///
+/// @param[in] splittedMenu 1d text wave with menu item lists for sub menus as returned by PEXT_SplitToSubMenus()
```
timjarsky, line 564: @MichaelHuth `splittedMenu` -> `splitMenu`

Packages/MIES/MIES_GuiPopupMenuExt.ipf
MichaelHuth: AddTraceToLBGraph() gets called with the menu entry as key. Therein GetKeyWaveParameterAndUnit() would return NaN and does
`		printf "Could not find %s in keyWave\r", key
`
Then AddTraceToLBGraph() returns NaN and this is silently ignored.

So except the print to the history nothing happens.

Packages/MIES/MIES_GuiPopupMenuExt.ipf
MichaelHuth: I renamed the WAVE reference to splitMenu in all related functions.

Packages/MIES/MIES_GuiPopupMenuExt.ipf
timjarsky: @MichaelHuth In cases where we know it will be common to have a missing list item, would it be possible to suppress the error message?

MichaelHuth: @timjarsky Probably yes. But as user I would consider present menu items that jsut do nothing inconvenient. 

How about checking for each menu item from the user defined list against the actual present keys. Then we disable menu items that are not present in the LNB by preceding the menu item with a '('. 
See also `DisplayHelpTopic "Special Characters in Menu Item Strings"`

With such approach the menu items would still be visible, just greyed out and not selectable. No change on the handler would be needed.

timjarsky: @MichaelHuth Checking and disabling sounds good to me but do you think it would add noticeable lag to the menu?

MichaelHuth:
Rebased.

Packages/MIES/MIES_GuiPopupMenuExt.ipf
t-b: @MichaelHuth @timjarsky I think the current design needs some more explanation before we dive into changing it. The groups for the labnotebook keys are calculated when you click the "Lab Notebook Entries" GUI control. So all the present labnotebook keys can be sorted into their respective group at this point. So from my understanding we don't need to support disabled menu items as we build the groups on the fly from the existing labnotebook keys.

timjarsky:
@MichaelHuth After initializing and configuring miss (only TP has run) I get the following bugout when selecting the DB labnotebook button:

image

t-b:
@timjarsky @MichaelHuth I presume the textual labnotebook is empty at this point.

MichaelHuth:
@timjarsky The menu setup functions as well as DB_GetLBKeys are aware now of null wave references and should handle them properly.

timjarsky:
@t-b @MichaelHuth flawless!!

PR #510, opened 2020-04-02
Allow user to adjust the setvariable limits

### LINKED ISSUE #450: Think about ways to include setvariable limits
One option is via configuration:
- [ ] Needs to work on per group basis to not clutter the JSON config
- [ ] Don't include default values

Or add a contextual popup menu where the users can select different increments.

t-b:
Close #450.

I've changed the behaviour back that the user must *not* click at the up/down buttons. As this would have the side effect of changing the value.

I've also added two more gems ;)

- [x] Add to ALL/VC/IC controls
- [x] investigate default not being highlighted on scale controls on first use

### DIFF
--- a/Packages/MIES/MIES_Constants.ipf
+++ b/Packages/MIES/MIES_Constants.ipf
@@ -584,7 +584,7 @@ Constant HARDWARE_DAC_EXTERNAL_TRIGGER = 0x1
 /// @}
 
 /// Used to upgrade the GuiStateWave as well as the DA Ephys panel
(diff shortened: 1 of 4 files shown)


timjarsky:
@t-b it won't past testing because it does not compile; 		JSONXOP_GetKeys/Z=(ignoreErr)/Q=(JSON_QFLAG_DEFAULT)/FREE jsonID, jsonPath, result; AbortOnRTE

t-b:
@timjarsky Done. The reason that the default was not highlighted for some controls was that these controls were missing the limits all together. Fixes now.

PR #520, opened 2020-04-09
Bugfix/too short dimension labels makes entry not queryable from labnotebook

t-b:
That was a fun one.

Our dimension label usage in the labnotebook did not work for one LBN entry as we cut the entry names off after 31 bytes.

I've now fixed that for new labnotebooks and added an upgrade path for old ones. And a test.


t-b:
I found that while testing https://github.com/AllenInstitute/MIES/pull/507.

t-b:
@timjarsky I've fixed the failing tests.

PR #527, opened 2020-04-15
Bugfix/uploading crash dumps

t-b:
Found by Michael Huth.


timjarsky:
@t-b what is the expected behavior on Igor initialization? I ask because I observed changes to the crash dump folder (it was renamed) only after I selected upload crash dumps from the MIES menu.

t-b:
@timjarsky The crash dump uploading is happening on device locking once a day. If we have uploaded crash dumps the `Diagnostics` folder is renamed to `Diagnostics_old$num`.

Issue #428, opened 2020-02-01
Installer: Should apply ASLR patch and make ITC hardware work out of the box


t-b:
- https://github.com/AllenInstitute/ITCXOP2#windows-10
- https://github.com/AllenInstitute/MIES/blob/6a127c5cdc8b23b81769c901dd6e8ca798677f8b/Packages/MIES/MIES_DAC-Hardware.ipf#L722

t-b:
[ITC-Registry-files.zip](https://github.com/AllenInstitute/MIES/files/4183486/ITC-Registry-files.zip)

ms-ai:
The 64-bit version is bad. Error is
Exception calling "AddAccessRule" with "1" argument(s): "This access control list is not in canonical form and therefore cannot be modified."

I let Windows fix the error in the registry, and re-exported the 64-bit file. Attached is the new zip:
[itc_registry_files.zip](https://github.com/AllenInstitute/MIES/files/4215238/itc_registry_files.zip)

ms-ai:
Apparently, the new exported version that I sent also has the wrong canonical form. How can that be? I'll dig more.

ms-ai:
There was nothing wrong with your files - they don't specify the ACL themselves. The problem and solution are described here:
https://answers.microsoft.com/en-us/office/forum/office_2016-office_install/registry-keys-in-hklmsoftwarewow6432node-are/c15da441-cf0c-4bb7-b193-e6cae4adcd50

Cheers!

MichaelHuth:
The ITC XOP documentation states that ASLR needs to be disabled for Igor 32 Bit, but the script disables it for 64 and 32 bit Igor Pro. Is it also required for 64 bit Igor Pro?

t-b:
@MichaelHuth Let's disable it for both just to be sure.

PR #528, opened 2020-04-16
Installer update for ITC XOP compatibility

MichaelHuth:
If the installer is executed as admin and installation with hardware XOPs is
chosen:

- PSScript to disable ASLR for Igor Pro from ITC XOP is called
- PSScript to fix Office365 bug that makes additional Wow3264Node entries
  ineffective is called
- Registry is set for ITC devices for 32 bit and 64 bit if running a 64 bit OS.

- existing ITC registry entries are overwritten in installation
- ITC registry entries are not removed on uninstallation
- ASLR setting for Igor Pro is not removed upon uninstallation

closes https://github.com/AllenInstitute/MIES/issues/428

requires latest ITC XOP version with the scripts in tools folder


MichaelHuth:
I tried several times and the keys are always overwritten (as intended). Do you have some special settings (rights maybe) that prevent the keys from being written?

MichaelHuth:
@t-b Could I have a registry dump of the keys that are relevant for ITC1600 ?

MichaelHuth:
@t-b small detail, but probably not critical: When usually creating a key the default value is a string with 'no value set'. The NSIS default API allows only to write a value, so I write an empty string instead.

t-b:
Note to self: Test on fresh Windows 10 on 32bit/64bit in vmware

t-b:
Rebased to latest master.

t-b:
- [ ] Change ALLUSER check to ISADMIN
- [ ] Test

timjarsky:
@t-b has a `clean` installation been tried with this installer?

t-b:
@timjarsky Yes, that is what I meant with `fresh`.

PR #515, opened 2020-04-07
Cleanup of release package and added .editorconfig

t-b:
Close #320 

Spring cleanup!

PR #514, opened 2020-04-03
Add help and check functions for expected analysis function parameters

t-b:
- [x] Implement check function support
- [x] Execute check function somehow earlier in DAP_CheckSettings compared to the pre DAQ event
- [x] Convert the check and help for the patchseq analysis functions
- [x] Add more tests

Close #124.

PR #529, opened 2020-04-16
Bugfix/analysis parameter checking

t-b:
If you look into the diff on github you might notice the reduced indentation. This is due to editorconfig support from #515. 

I've found these while working on #514.


t-b:
I've fixed the tests as well. They were relying on the bug..

PR #538, opened 2020-04-24
Removed WaveClear in DAG_Update

MichaelHuth:
WaveClear does not seem to be required at this location.

Just some leftover?


t-b:
Nice find.

PR #541, opened 2020-05-05
MIES_Include.ipf: Require the latest nightly build

t-b:
Our new GUIPopupExtension triggered an Igor Pro bug by having long dimension labels in zeroed out dimensions.

The code

```igorpro
Function Dostuff()
     Make/O/N=1 data
     SetDimLabel 0, -1, $PadString("", 32, 10), data
     Redimension/N=0 data
End
```

triggers that crash on save.

Let's require a more recent nightly build to avoid that crash.

### DIFF
--- a/Packages/MIES_Include.ipf
+++ b/Packages/MIES_Include.ipf
@@ -3,13 +3,13 @@
 #pragma rtFunctionErrors=1
 
 // Igor Pro nightly installation:
-// - Download from https://www.byte-physics.de/Downloads/WinIgor8_10APR2020.zip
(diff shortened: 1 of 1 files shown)


MichaelHuth:
Didn't crashed here on save, but got
`BUG: CopyDimLabelToCString`
in history.

t-b:
The failed tests output is real. The following tests fail

```
MWWO_HandlesLockedDest, MWWO_RecursiveWorks, rename_wave_only 
```

This is an upstream issue, already reported to WM.

t-b:
I've upgraded to a new nightly.

PR #550, opened 2020-05-28
Fix duplicated session start times

t-b:


timjarsky:
@t-b here or in a chat, can you describe the bug that led to this PR? Thanks!

t-b:
@timjarsky I've messed up the PR. Now the diff it is much shorter.

The bug was that the session start time was not touched on a save and split experiment invocation. This leads to NWB files with the same session start time from different experiments.

We now reset the session start time and work around too far off session start times on NWB export.

PR #179, opened 2019-05-30
Feature support for NWBv2

t-b:
#### Cross links

https://github.com/AllenInstitute/MIES/issues/22
https://github.com/AllenInstitute/MIES/pull/179 (TODO list)
https://github.com/AllenInstitute/IPNWB/issues/5
https://github.com/AllenInstitute/IPNWB/pull/16
https://github.com/t-b/ndx-MIES/pull/1

Requires:
- https://github.com/NeurodataWithoutBorders/pynwb/pull/1146

Issues:
- https://github.com/NeurodataWithoutBorders/nwb-schema/issues/410

TODOs:
- [x] Ready to merge according to Wayne
- [x] Fix ISO8601 timezone issues
- [x] Add pynwb read tests as CI
- [x] Fix documentation building
- [x]  Test with I=0 data
- [x] Check that unassociated channels are stored correctly
- [x]  Add pynwb code for using validation via the cmdline tool
- [x]  add schema extension for all our additions
    - [x] clarify recursive folder usage with Ryan
    - [x] Add ndx-MIES repo as submodule to IPNWB 
    - [x] convert YAML files to JSON and store them as well
    - [x] Write ndx-MIES specification to NWBv2 files and check that it validates
    - [ ] add schema extension to ndx-cataloge
- [x] remove some fields which should not be there (units and scale)
- [x] add objectIDs, https://pynwb.readthedocs.io/en/latest/tutorials/general/object_id.html and https://tools.ietf.org/html/rfc4122
- [x] Wait for Wayne doing conversion tests again
- [x] Fix wrong object_id
- [x] https://github.com/hdmf-dev/hdmf/issues/354#issuecomment-630948209

Close #22.

PR #552, opened 2020-05-29
DQ_StopOngoingDAQAllLocked: Make it more robust

t-b:
This function is called from BeforeUncompiledHook and thus needs to
always work. Even with really old experiments which don't have the
expected GUI controls and thus assert out.


t-b:
This is also incorporated since #179 was merged.

PR #560, opened 2020-06-11
SweepFormula: Add area operation

### LINKED ISSUE #269: SweepFormula:  area function
The `area` function should get implemented instead of the `Integrate` operation.

From my basic understanding, there is only disadvantages from implementing `area` instead of `integrate`. Both can do trapezoid integration but area is much more primitive. e.g. it can not handle dimension-based operation, does not do end-point filling and thus looses the data alignmen, shifting the x-values by one point and introduces `NaN` at the end.

I am not a fan of end point guessing either but we can also deactivate such feature for differentiate and integrate which then allows to have the above described behavior also for our functions but the user must be aware that multiplying e.g. `integrate(data))*data()` can lead to strange results.

The IP area function:

> area(waveName  [, x1, x2  ] )
> The area function returns the signed area between the named wave and the line y=0 from x=x1  to x=x2  using trapezoidal integration, accounting for the wave's X scaling. If your data are in the form of an XY pair of waves, see areaXY.

The IP Integrate Operation

> Integrate [/DIM = d  /METH=m  /P/T][typeFlags ] yWaveA  [/X = xWaveA ][/D = destWaveA ] [, yWaveB  [/X = xWaveB ][/D = destWaveB ][, ...]]
> The Integrate operation calculates the 1D numeric integral of a wave. The X scaling of each wave is taken into account. Rectangular integration is used by default.
> Integrate is multi-dimension-aware in the sense that it computes a 1D integration along the single dimension specified by the /DIM flag or along the rows dimension if you omit /DIM.

 steps to integrate area

- reduce 2d input waves to 1d waves.
- determine calculation x-ranges for all columns using `ScaleToIndex` and `DimSize` splitting
- use area in a for loop
- assemble the result to a new wave
- transfer wave scaling

#238

t-b:
This just returns the last array entry of the integrate function.

Close #269.


timjarsky:
@t-b I'm not seeing an output from the SF area function:

image

t-b:
@timjarsky Thanks. I've fixed two bugs leading to that result. Please give it a shot again. Breaking it is as usual encouraged ;)

timjarsky:
> @timjarsky Thanks. I've fixed two bugs leading to that result. Please give it a shot again. Breaking it is as usual encouraged ;)
> 
> Edit: I also fixed another bug. I presume the area below a full cycle of a sine should be twice as much as that of half the cycle and not zero. (This means I can not just take the the last value from integrate but have to take the absolute value of the data first).

@t-b I think we may still be failing to account for another edge case; I would expect the area to be the same for both sweeps in the image below.
image

t-b:
@timjarsky Vertically shifting the curve should kep the area constant? In that case I need to zero first or?

timjarsky:
> @timjarsky Vertically shifting the curve should kep the area constant? In that case I need to zero first or?

let's add a default zeroing option if you think that would work?

PR #561, opened 2020-06-11
Packages/MIES/MIES_SweepFormula.ipf: Assert out on invalid sweep range

### LINKED ISSUE #468: SweepFormula: Should error out when the range in data does not include any points
image

>SF does funny business when the epoch in the formula doesn't exist.

Add error message instead of smiley in this case.

t-b:
If the range for the data function is not completely inside the given
sweeps we currently return bogus data.

Let's assert out instead as this is more correct.

This has yet to use proper error handling, but that will be not part
of this commit.

Closes #468.

This is actually stricter to the plain wording of the issue. We now bug out if the full range is not completely inside all sweeps. Is that the right thing to do?


t-b:
@timjarsky Did you see my question?

> This is actually stricter to the plain wording of the issue. We now bug out if the full range is not completely inside all sweeps. Is that the right thing to do?

timjarsky:
- [ ] Fix test

PR #578, opened 2020-07-20
SweepFormula: Add area operation

t-b:
Successor of #560.


t-b:
@timjarsky No more modulus, so negative area is possible again. And optional zeroing which defaults to 1.

PR #593, opened 2020-07-31
Feature/add scalebars for pa

t-b:
- [x] fix wave read error
- [x] allow multiple PA plots, i.e. one PA plot for a browser
- [ ] place the y-scale bar *exactly* over the y axis

### DIFF
--- a/Packages/MIES/MIES_Cache.ipf
+++ b/Packages/MIES/MIES_Cache.ipf
@@ -163,7 +163,7 @@ End
 Function/S CA_AveragingKey(waveRefs)
 	WAVE/WAVE waveRefs
 
-	return CA_WaveCRCs(waveRefs, crcMode=2) + "Version 3"
+	return CA_WaveCRCs(waveRefs, crcMode=2) + "Version 4"
 End
 
 /// @brief Calculate the CRC of all metadata of a dimension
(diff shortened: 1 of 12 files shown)


timjarsky:
@t-b v-clamp scale bars looked good. I-clamp bars aren't optimally positioned:

image

t-b:
@timjarsky Looks like my positioning hack is not as good as I hoped. I'll get that properly positioned without hacks.

t-b:
@timjarsky Updated. Regarding the y scale setvariable: When you change between sweeps with different clamp modes, the user currently needs to adapt the y scale so that it is not totally off scale. That is IMHO not very pleasant.

timjarsky:
> @timjarsky Updated. Regarding the y scale setvariable: When you change between sweeps with different clamp modes, the user currently needs to adapt the y scale so that it is not totally off scale. That is IMHO not very pleasant.

@t-b in the DA tab of the DA_ephys panel, we handle a similar issue by having separate controls for I- and V-clamp. We could also manage this, the same way as the x-axes. The concern with the latter approach is the need to constantly update when with each new sweep - my preference might be to only update once per RA cycle or something like that.

t-b:
Resolved conflicts and rebased. Will merge once CI passes.

PR #606, opened 2020-08-12
GetSettingsJSONid: Handle invalid JSON ID

t-b:
For reasons yet to be understood the settinsJSONid does sometimes refer
to a non-existing JSON document.

We now regenerate the package settings in this case as well.


t-b:
I've created the PR against the `main` branch instead of `release/2.1`. I've fixed that locally and pushed.

PR #607, opened 2020-08-12
OpenAboutDialog: Scale the panel and notebook actions correctly

t-b:
For High-DPI displays, see DisplayHelpTopic "Control Panel Resolution on
Windows", we need to rescale the panel size and the notebook action
scaling.

Verified using 125% scaling, aka ScreenResolution 120, on Windows 10.


t-b:
I've created the PR against the `main` branch instead of `release/2.1`. I've fixed that locally and pushed.

PR #611, opened 2020-08-16
IVS_ExportAllData: Remove existing NWB files first as requested

### LINKED ISSUE #610: nwb2 export all data not overwriting prev data
The "IVS_ExportAllData" command to generate an nwb2 file that has previously been exported does not overwrite previously saved data. It appends the data instead. I need the nwb2 file to be overwritten if exporting on multiple occasions as the code used to do.

t-b:
Close #610.

@aleonlein Here you go. I can't find the place in MIES where we deleted the NWB file in the old code before exporting into it. Maybe that was done on the client side (i.e. outside of MIES)?

The patch is based on the 2.1 release, ping my if you need something built on top of master.

### DIFF
--- a/Packages/MIES/MIES_IVSCC.ipf
+++ b/Packages/MIES/MIES_IVSCC.ipf
@@ -435,6 +435,9 @@ End
 Function IVS_ExportAllData(filePath)
 	string filePath
 
+	CloseNWBFile()
+	DeleteFile/Z filePath
+
 	printf "Saving experiment data in NWB format to %s\r", filePath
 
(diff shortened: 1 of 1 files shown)


aleonlein:
Thank you, Thomas. I didn't check the deprecated, client-side repo to see if the file was removed and then exported. This issue came up after users analyzed the nwb2 files and I assumed it was a new issue during integration testing. It's possible that no one looked at the old nwb files I was saving out in earlier during integration tests and this problem just wasn't found. I appreciate the quick fix.

aleonlein:
This has been tested and looks good

PR #613, opened 2020-08-17
Remember window positions

t-b:
This needs real live user testing to see if it is useful or just annoying,


timjarsky:
@t-b this is fantastic!!!

PR #619, opened 2020-08-30
Bugfix/make restoring sweeps in databrowser work again

### LINKED ISSUE #154: Make restoring the original data in the databrowser/sweep browser work again
Broken since the introduction of splitted 1D waves in 68109cc1 (Add artefact removal, 2016-12-07).

One solution could be to never backup the splitted 1D waves but instead resplit the sweep waves on restore.

t-b:
- [x] Add tests for basic backup functions

Close #154.

### DIFF
--- a/Packages/MIES/MIES_AnalysisBrowser.ipf
+++ b/Packages/MIES/MIES_AnalysisBrowser.ipf
@@ -1713,7 +1713,8 @@ static Function AB_LoadSweepFromNWBgeneric(h5_groupID, nwbVersion, channelList, 
 		WAVE/Z/SDFR=sweepDFR targetName = $channelName
 		// nwb files created prior to 901428b might have duplicated datasets
 		ASSERT(!WaveExists(targetName) || (WaveExists(targetName) && WaveCRC(0, targetName) == WaveCRC(0, loaded)), "wave with same name, but different content, already exists")
-		Duplicate/O loaded, sweepDFR:$channelName
+		Duplicate/O loaded, sweepDFR:$channelName/WAVE=targetRef
+		CreateBackupWave(targetRef)
 		WaveClear loaded
 	endfor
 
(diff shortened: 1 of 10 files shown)


t-b:
@timjarsky I've added the tests.

timjarsky:
@t-b have restore clear `Average Traces` and `Zero Traces` checkbox.

t-b:
@timjarsky I've removed that on purpose.
The old approach in main is confusing IMHO.
If I restore them before updating the plot we will immediately modify the restored waves again. Do you think this is desirable? You can get that now by removing the `SetCheckbox` calls at the end in BSP_ButtonProc_RestoreData.

timjarsky:
@t-b bug recipe: open analysis browser, use headstage removal on a few sweepszero sweeps, kill analysis browser, open new analysis browser

sweeps are still zeroed except those that were removed - Not sure this is a bug. I might expect ignored sweeps to be zeroed but I wouldn't expect ignored sweeps to be included in the average.

sweeps cannot be restored in new analysis browser

potential solution:
automatically restore data when the analysis browser is closed?

timjarsky:
> @t-b bug recipe: open analysis browser, use headstage removal on a few sweeps, zero sweeps, kill analysis browser, open new analysis browser
> 
> sweeps are still zeroed except those that were removed - Not sure this is a bug. I might expect ignored sweeps to be zeroed but I wouldn't expect ignored sweeps to be included in the average.
> 
> sweeps cannot be restored in new analysis browser
> 
> potential solution:
> automatically restore data when the analysis browser is closed?

@t-b I think you can ignore this.  I think the weird behavior was because I didn't have clean start after switching branches. Sorry!

timjarsky:
> @timjarsky I've removed that on purpose.
> The old approach in main is confusing IMHO.
> If I restore them before updating the plot we will immediately modify the restored waves again. Do you think this is desirable? You can get that now by removing the `SetCheckbox` calls at the end in BSP_ButtonProc_RestoreData.

@t-b Do you recall why we have a restore button at all? Why not just restore when zeroing is unchecked?

timjarsky (approved):

PR #623, opened 2020-08-31
Bugfix/more user friendly error messages in sweep formula

### LINKED ISSUE #275: SweepFormula: Error Handling
Currently, Aborts are used to quickly bail out of the parser on error. This is a better approach than asserts for the parser, as it is not technically a program failure at that point if the user e.g. forgot to add a closing bracket.

At the point of the Parser we do not have Variables space to write to but I would really also like to add line number/position of the parsing error and highlight that in the formula window.

Adding `AbortOnValue` here would be great but It requires that the Abort is catched, which it was not in my tests.

First thing to do:
- Don't use `ASSERT`
- Don't use dialogs like in bracket mismatch
- Better error messagrs for bracket mismatch

t-b:
Close #275.

### DIFF
--- a/Packages/MIES/MIES_DataBrowser_Macro.ipf
+++ b/Packages/MIES/MIES_DataBrowser_Macro.ipf
@@ -11,7 +11,7 @@
 
 Window DataBrowser() : Graph
 	PauseUpdate; Silent 1		// building window...
-	Display /W=(404.25,116.75,839.25,386)/K=1  as "DataBrowser"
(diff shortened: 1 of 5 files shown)


t-b:
@timjarsky Both are fixed. You now also don't need to resize the panel to view the error message as it is shown in full length in the tooltip as well.

timjarsky:
@t-b the indicator light isn't moving with the panel when the panel is resized

e.g., 
image

Everything else checks out! Thank you!!

t-b:
@timjarsky Should be fixed.

timjarsky (approved):

PR #629, opened 2020-09-12
Bugfix/tp during daq and repeated acquisition

### LINKED ISSUE #517: TP during DAQ and RA do not work together if only TP during DAQ channels are present
The reason is that https://github.com/AllenInstitute/MIES/blob/2481d1c8a0f33c083228f475635c6f1f5c150094/Packages/MIES/MIES_Indexing.ipf#L221
which is responsible for calculating the number of sweeps for RA, called by DAP_UpdateSweepSetVariables, does ignore TP during DAQ channels. That was added so that Indexing with TP during DAQ channels work.

We can now either forbid it or work around it. I tend to think that we to make it work, at least for #313.

t-b:
Close #517.

### DIFF
--- a/Packages/MIES/MIES_Indexing.ipf
+++ b/Packages/MIES/MIES_Indexing.ipf
@@ -250,7 +250,7 @@ Function IDX_MaxNoOfSweeps(panelTitle, IndexOverRide)
 			continue
 		endif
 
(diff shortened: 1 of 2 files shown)


t-b:
@timjarsky It was quick to do.

timjarsky:
@t-b works perfectly!  Also, 900 ms between recorded TP events when going from TP during ITI to a sweep is pretty good! It's faster still when going from sweep to TP during ITI.

timjarsky (approved):

PR #641, opened 2020-09-21
Open config files in notebook

### LINKED ISSUE #630: Add option to edit package json file in igor
We could just open it in a plaintext notebook.

t-b:
Close #630.

This is the bare-bones solution.

I could also create a panel, have a notebook subwindow in that and add a save button on the panel. The save button would allow to check the JSON file for syntax errors. This would catch the most obvious errors.


timjarsky:
@t-b perhaps for a different PR, but could we add a way to open the window position JSON?

timjarsky (approved):

PR #653, opened 2020-09-27
AB_ScanFolder: Prevent loading the currently written to NWB file

t-b:
This does not work as we get a HDF5 library error when we try to read
with one file descriptor from it and write into it with another one.

Even if we used the same file descriptors for it, we would need updating
support in the analysis browser to make it really worthwhile using it.

So now we ignore the currently exported into file and warn the user.

Another one from the presentation.

### DIFF
--- a/Packages/MIES/MIES_AnalysisBrowser.ipf
+++ b/Packages/MIES/MIES_AnalysisBrowser.ipf
@@ -2155,7 +2155,8 @@ End
 static Function AB_ScanFolder(win)
 	string win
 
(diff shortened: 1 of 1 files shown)


timjarsky:
@t-b what's the definition of currently written? I ask because I can open an NWB that I am acquiring into in one instance of Igor, and load in another instance of igor.

t-b:
I meant in the same instance of IP.

timjarsky:
@t-b okay, I confirmed it can't be opened in the same instance.

timjarsky (approved):

PR #662, opened 2020-10-02
Various minor indexing fixes

t-b:


timjarsky:
@t-b weird bug: the labnotebook and the da tab indicate recovery stimset was being acquired but data shows a different stimset!

timjarsky:
@t-b i think i figured out the above issue - it's a problem with the stimset. Sorry!

timjarsky (approved):

PR #663, opened 2020-10-02
RA_SkipSweeps: Fix various corner cases

t-b:
- [x] Check tests. The new limiting probably broke the test. We can also wait to the second IT for skipping back.

The current code did not expect that you call first without
limitToSetBorder=0 and then with limitToSetBorder=1.

It was also not working for -inf and did not adjust the active set count
correctly when limitToSetBorder=0.

These bugs are now fixed. The solution is to first calculate a skipCount
which is limited to the set borders if requested. And when a new count
is calculated using RA_SkipSweepCalc and the so calculated count offset
is used for adjusting count and the activeSetCount.

### DIFF
--- a/Packages/MIES/MIES_Async.ipf
+++ b/Packages/MIES/MIES_Async.ipf
@@ -6,7 +6,7 @@
 #pragma ModuleName=MIES_ASYNC
 #endif
 
-/// @file MIES_ASYNC.ipf
+/// @file MIES_Async.ipf
 /// @brief __ASYNC__ This file holds the asynchronous execution framework
 ///
 /// \rst
--- a/Packages/MIES/MIES_DAEphys.ipf
+++ b/Packages/MIES/MIES_DAEphys.ipf
@@ -5082,7 +5082,7 @@ Function DAP_ButtonProc_skipSweep(ba) : ButtonControl
 
 	switch( ba.eventCode )
 		case 2:
-			RA_SkipSweeps(ba.win, 1)
+			RA_SkipSweeps(ba.win, 1, limitToSetBorder = 1)
 			break
 	endswitch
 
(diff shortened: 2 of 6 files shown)


timjarsky:
@t-b with indexing, I'm finding I can skip to the end of a RA cycle but there is no indexing. The behavior I was expecting was that sweep skipping would be limited to the number of sweeps in the active set.

t-b:
> @t-b with indexing, I'm finding I can skip to the end of a RA cycle but there is no indexing. The behavior I was expecting was that sweep skipping would be limited to the number of sweeps in the active set.

We can do that, but that is not what we are currently doing.

See https://github.com/AllenInstitute/MIES/blob/6177df3e29dc3238c254ab8067d1adff0d9307e9/Packages/MIES/MIES_DAEphys.ipf#L5080-L5090 and https://github.com/AllenInstitute/MIES/blob/6177df3e29dc3238c254ab8067d1adff0d9307e9/Packages/MIES/MIES_DAEphys.ipf#L5092-L5102

We would just need to pass `limitToSetBorder=1`.

timjarsky:
> We would just need to pass `limitToSetBorder=1`.

@t-b let's try it!

timjarsky:
@t-b I'm finding that indexing works correctly with skipping back and not skipping forward. 

option A, use limitsweepborder=1 with skip forward (and not back).
option B, make indexing work with skip forward.

t-b:
@t-b Implement option A and add sweep skipping from ITI.

t-b:
Done, I've also added one more test.

timjarsky:
@t-b skipping backward is creating unexpected indexing steps (it indexes forward!). Do you want to limit skipping to set borders for forward and back?

t-b:
> @t-b skipping backward is creating unexpected indexing steps (it indexes forward!). Do you want to limit skipping to set borders for forward and back?

Yes, I'll do that as well.

t-b:
@timjarsky Done. Funny storry, I actually wondered why we would only use limitToSetBorder for one direction and not both.

t-b:
Fixed the new test.

timjarsky:
@t-b with limit to set border =1 and indexing, it is still possible to index backward so we aren't avoiding the indexing issues.

timjarsky (approved):
Merged with the knowledge that `limit to set border` only works with a single headstage or when the sets have the same length across multiple headstages.

PR #667, opened 2020-10-06
Bugfix/kill too long running jobs

### LINKED ISSUE #533: Enhance hung build killer on bamboo CI boxes
Currently when the "Hung build killer" for bamboo kicks in it leaves the Igor instances open. This is due to https://confluence.atlassian.com/bamkb/script-builder-doesn-t-kill-child-processes-189891893.html.

The obvious solution, trapping inside the bash script, does not work due to https://stackoverflow.com/a/49378862.

So is looks like we need to resurrect the solution using taskkill from 4332a705 (Bugfix: Kill correct igor instance in auto run script for unit testing, 2017-08-28).

t-b:
Close #533.

### DIFF
--- a/tools/unit-testing/autorun-test.sh
+++ b/tools/unit-testing/autorun-test.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-# Usage: $0 [-p <name of pxp to run against>] [-v <igor version string>]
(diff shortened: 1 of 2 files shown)


t-b:
I'm merging now as I also had to tweak bamboo for this.

@MichaelHuth All new/chagned PRs need to be rebased to latest main before the tests pass again.

PR #686, opened 2020-10-19
Fix bug where only chaning the failed pulses level did not result in a replot

t-b:


timjarsky:
@t-b does this take into account that the pulses may already be zeroed (testing for failed pulses would normally happen before zeroing)?

t-b:
> @t-b does this take into account that the pulses may already be zeroed (testing for failed pulses would normally happen before zeroing)?

Yes, nothing changed in that area.

timjarsky (approved):

PR #687, opened 2020-10-19
PA plot: Speedups for zero wave

t-b:


Packages/MIES/MIES_PulseAveraging.ipf
```diff
@@ -1628,7 +1628,7 @@ static Function PA_ZeroPulses(WAVE/WAVE set, STRUCT PulseAverageSettings &pa)
 	endif
 
 	Make/FREE/N=(DimSize(set, ROWS)) junkWave
-	MultiThread junkWave = ZeroWave(set[p]) + PA_UpdateMinAndMax(set[p])
+	MultiThread junkWave = ZeroWave(set[p]) && PA_UpdateMinAndMax(set[p])
```
timjarsky, line 1631: @t-b can we talk about this line the next time we chat?

timjarsky: @timjarsky If zeroWave(set[p]) returns 1, then PA_UpdateMinAndMax(set[p]) is run

timjarsky (approved):

MichaelHuth:
@t-b Please don't merge, this is not threadsafe and causes the Igor crashes I see. This is only safe without Multithread.

timjarsky:
> @t-b Please don't merge, this is not threadsafe and causes the Igor crashes I see. This is only safe without Multithread.

@t-b uh oh, someone at Wavemetrics isn't going to be happy about this discovery ;)

t-b:
I'm merging this now because:
- I does not provoke a crash as long as `set` does not have duplicates waves
- This code will be rewritten anyway in one of the new speedup PRs

PR #688, opened 2020-10-19
Bugfix/pa plot image various

t-b:
This fixes the "additional data" assertion and makes the x zero line available to both plots.


timjarsky:
@t-b line on trace plot looks great!

timjarsky (approved):

Issue #691, opened 2020-10-23
pulse start times in PA plot outputs warning in CI


t-b:
I'm getting

```
  Warn: Pulse start time from epochs:
  '_free_'[0]= {520,620,720,820,920,1020,1120,1220,1320,1420}
  
 from Calculation:
  '_free_'[0]= {619.99,719.99,819.99,919.99,1019.99,1119.99,1219.99,1319.99,1419.99}
```

when running the new spike control analysis function. Experiment [HardwareTests.zip](https://github.com/AllenInstitute/MIES/files/5430153/HardwareTests.zip) should be enough for testing. 9726da07 (changes, 2020-10-23) has that code, or grab feature/add-analysis-function-failed-pulses from #650.

PR #697, opened 2020-10-26
Fix for PA_CalculatePulseStartTimes sometimes ignoring the first pulse

MichaelHuth:
The issue was that for the case that totalOnSetDelay equals the start time of
the first pulse then FindLevels could not detect the first edge any more.

The fix is that FindLevels starts now from totalOnsetDelay - delta.
Start position is capped to be >= 0.

close https://github.com/AllenInstitute/MIES/issues/691

### DIFF
--- a/Packages/MIES/MIES_PulseAveraging.ipf
+++ b/Packages/MIES/MIES_PulseAveraging.ipf
@@ -201,7 +201,7 @@ static Function/WAVE PA_CalculatePulseStartTimes(DA, fullPath, channelNumber, to
 	string fullPath
 	variable totalOnsetDelay
 
-	variable level, delta
+	variable level, delta, searchStart
 	string key
 	ASSERT(totalOnsetDelay >= 0, "Invalid onsetDelay")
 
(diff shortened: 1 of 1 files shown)


Packages/MIES/MIES_PulseAveraging.ipf
```diff
@@ -1692,7 +1695,7 @@ static Function PA_ZeroPulses(WAVE/WAVE set, STRUCT PulseAverageSettings &pa)
 	endif
 
 	Make/FREE/N=(DimSize(set, ROWS)) junkWave
-	MultiThread junkWave = ZeroWave(set[p]) && PA_UpdateMinAndMax(set[p])
+	MultiThread junkWave = ZeroWave(set[p]) + PA_UpdateMinAndMax(set[p])
```
t-b, line 1698: @MichaelHuth unrelated change?

t-b (changes requested):

Packages/MIES/MIES_PulseAveraging.ipf
MichaelHuth: Yes, thats weird. I kick that out.

t-b (approved):
LGTM

timjarsky:
@MichaelHuth If we ever need to look back at this PR it may be helpful to define delta in the description.

timjarsky (approved):

t-b:
@timjarsky The code has `delta = DimDelta(DA, ROWS)` so I think we should be able to get an understanding of the logic.

PR #665, opened 2020-10-06
Pulse Averaging: Pulse starting times are now read from the epoch information from lab notebook

MichaelHuth:
- Pulse starting times are in PA_GetPulseStartTimes extracted from the epoch
  information stored in the lab notebook
- as fallback the pulse starting times are determined the old direct way
  through findlevels.

close https://github.com/AllenInstitute/MIES/issues/207

PR #700, opened 2020-10-30
PA_DrawScaleBarsHelper: Draw x scale bar length if different

### LINKED ISSUE #696: PA plot: Rework scale bar drawing
Add label to every x scale bar which has a differnt physical unit length compared to the original value. Store the original value as graph user data.

### LINKED ISSUE #698: PA_DrawScaleBarsHelper assert in PA_plot
Reproduce with:
Enable PA plot, select some sweep, then 
- enable average plot, deconvolution plot
- disable OVS and show individual pulses, so that just average and deconvolution is shown
- now disable average plots, so that only deconvolution is shown 

image

image

t-b:
We have one x axis  and scale bar for each region (column). In order to
not clutter the image we only add one x scale bar the bottom right set.

We allow users zoomming in and out into each x axis, and then
dynamically resize the x scale bar to a fitting length.

This makes it really hard to know the exact physical length when
zooming.

We solve that by remembering the initial x bar length and always draw
the x bar length if it is different from the inital value.

Close #696.
Close #698.

### DIFF
--- a/Packages/MIES/MIES_Configuration.ipf
+++ b/Packages/MIES/MIES_Configuration.ipf
@@ -1586,7 +1586,7 @@ static Function CONF_ControlToJSON(wName, ctrlName, saveMask, jsonID, excCtrlTyp
 	string excCtrlTypes, excUserKeys
 
 	variable ctrlType, pos, i, numUdataKeys, setVarType, arrayIndex, oldSize, preferCode, arrayElemType
-	string wList, ctrlPath, controlPath, niceName, jsonPath, udataPath, udataKeys, uDataKey, uData, s, arrayName, arrayElemPath
+	string wList, ctrlPath, controlPath, niceName, jsonPath, udataPath, uDataKey, uData, s, arrayName, arrayElemPath
 
 
 	ASSERT((saveMask & (EXPCONFIG_SAVE_POPUPMENU_AS_STRING_ONLY | EXPCONFIG_SAVE_POPUPMENU_AS_INDEX_ONLY)) != (EXPCONFIG_SAVE_POPUPMENU_AS_STRING_ONLY | EXPCONFIG_SAVE_POPUPMENU_AS_INDEX_ONLY), "Invalid popup menu save selection. String only and Index only can not be set at the same time.")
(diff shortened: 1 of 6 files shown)


t-b:
@timjarsky Done.

timjarsky:
@t-b the exact value of the user input for the y-scale should be used by default. Only after zooming, does the user input get used as a scalar. Sorry I didn't make this distinction initially.

t-b:
> @t-b the exact value of the user input for the y-scale should be used by default. Only after zooming, does the user input get used as a scalar. Sorry I didn't make this distinction initially.

Will do, no problem.

t-b:
@timjarsky Done.

timjarsky:
@t-b y-scale bar doesn't rescale quickly enough; it invades neighboring plot when zooming in (zoom out seems good).

t-b:
@timjarsky I've fixed the issue of the y scale bar position. I could not find a way to avoid having recalculated y scale bar on deconvolution data addition. I've asked a question on igor exhange, https://www.wavemetrics.com/forum/general/adapt-user-zooming-axis, to clarify that.

t-b:
@timjarsky So the bug from https://github.com/AllenInstitute/MIES/pull/665#issuecomment-723393282 is fixed here. I've now also found a way to get the y scale bar behaviour you want.

timjarsky:
@t-b If I select the next sweep using `Popup_SweepControl_Selector` the scale bars are not reset to the value specified in `setvar_pulseAver_vert_scale_bar`

image


The scaling on zooming with the y-scale bars works pretty well but it could reset more quickly when zooming in. I the example below, the y-scale bars from the two plots are overlapping.

image

t-b:
@timjarsky Go for it!

timjarsky:
@t-b scale bars are nicely following the rules we came up with. Occasionally, but not always, the y-scale bar waits too long to rescale the original scale bar, and invades the plot above when zooming in on data. Once it has rescaled the first time, it scales on-time going forward.

timjarsky:
@t-b when testing this PR I noticed that if I have a sweep displayed in the databrowser (e.g., sweep 31) and OS in enabled, sweep 0 is overlaid. This feels like a regression bug to me, but I am not certain. It is also present in main.

t-b:
@timjarsky Yes that was not fixed good enough. I've fixed it again.

t-b:
Note to self:
- y scale bar extends outside the trace set area and changes it's size too slowly

t-b:
@timjarsky Go for it.

timjarsky:
@t-b I'm still seeing: https://user-images.githubusercontent.com/5693697/98583658-93251400-2279-11eb-8310-fbb475c13c3b.png

It most commonly occurs in the bottom left; the scale bar with the label

t-b:
I need a demo.

t-b:
Well not really a demo but we need to decide what should happen.

t-b:
@timjarsky Rebased

timjarsky (approved):

PR #676, opened 2020-10-14
Adapted PA_Average to support incremental averaging

MichaelHuth:
Generally for incremental averaging the function has to return
the sum per row as well as the counts per row additionally.
For this the new optional parameter getComponents was introduced.

To keep auto indexing calls to MIES_fWaveAverage possible, instead of
using MRS the function now returns a wave reference wave with 3 rows.

The first row is the average wave as before.
The second row are the sums and the thrid the counts.
The second and third row are only setup with valid return data when
getComponents = 1 was used.

Having the sums per row and counts per row, it is also possible to call the
averaging incremental. Previous averaging data can be given by the optional
prevAvgData wave reference parameter.
The first entry must contain the sums and the second the counts per row.

This data is then included in the averaging of the yWaves.

close https://github.com/AllenInstitute/MIES/issues/733
close https://github.com/AllenInstitute/MIES/issues/673
close https://github.com/AllenInstitute/MIES/issues/207
close https://github.com/AllenInstitute/MIES/issues/716
close https://github.com/AllenInstitute/MIES/issues/713
close https://github.com/AllenInstitute/MIES/issues/698
close https://github.com/AllenInstitute/MIES/issues/695
close https://github.com/AllenInstitute/MIES/issues/705

PR #715, opened 2020-11-06
Databrowser: Fix overlaying sweeps doing an incremental update for ex…

### LINKED ISSUE #709: PulseAveraging: When enabling OVS, first sweep is shown twice
When PA is enabled and then OVS is enabled the first sweep is shown twice in the PA plot.

t-b:
…isting sweeps

Recipe for reproduction:
- Open Databrowser
- Open PA plot
- Enable OVS

You know have duplicated traces in the PA plot. The reason is the
incremental OVS update introduced in ffca2076 (Add incremental update
support for overlay sweeps, 2020-07-15). The call to OVS_UpdatePanel in
BSP_CheckProc_OverlaySweeps does not recognize any new sweeps, the
selected sweeps list will be empty, but we do a full update
including calling UpdateSweepPlot. But as we have no sweeps selected we
take the early return in DB_UpdateSweepPlot, and thus don't update the
PA plot as well.

Then we select a sweep with OVS_ChangeSweepSelectionState, this new
sweep is found with the incremental update and we plot the new data from
that sweep in the PA plot.

But as that data is already shown in the PA plot we now have
duplicated traces.

The fix is to make the sweep selection when nothing is selected already
in OVS_UpdatePanel, thus replotting everything correctly, and skipping
the incorrect update later on.

Close #709.

### DIFF
--- a/Packages/MIES/MIES_BrowserSettingsPanel.ipf
+++ b/Packages/MIES/MIES_BrowserSettingsPanel.ipf
@@ -1136,29 +1136,15 @@ End
 Function BSP_CheckProc_OverlaySweeps(cba) : CheckBoxControl
 	STRUCT WMCheckBoxAction &cba
 
-	string graph, bsPanel, scPanel
-	variable index, sweepNo
+	string graph, bsPanel
 
 	switch(cba.eventCode)
 		case 2: // mouse up
 			graph   = GetMainWindow(cba.win)
 			bsPanel = BSP_GetPanel(graph)
-			scPanel = BSP_GetSweepControlsPanel(graph)
 
 			BSP_SetOVSControlStatus(bsPanel)
 			OVS_UpdatePanel(graph, fullUpdate = 1)
-
-			WAVE/Z sweeps = GetPlainSweepList(graph)
-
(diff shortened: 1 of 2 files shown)


MichaelHuth:
This fixes the double appearance of traces, but there is one thing problematic:
For a 2x2 sweep:
If PA is switched on then the traces are numberes 0 - 3.
If I switch on OVS then there are now 4 traces (which is correct) but they are numbered 4 - 7.
image

If I add another sweep now incremental the new sweep traces start counting also from 4:
image

If the second sweep would have the same channel/region this collides.

I think the error is that the numbering after switching on OVS changes to 4 - 7, should be 0 - 3.

t-b:
@MichaelHuth Yes this is a bug. But I don't think this is worth fixing as you change how that works in #676. Or what do you think?

MichaelHuth:
image

@t-b Hmm ok,I think we have nothing in the code that relies directly on the trace names.

MichaelHuth (approved):
Fixes the number of traces

timjarsky (approved):

PR #724, opened 2020-11-11
NWB_ExportAllData: Handle missing config waves

t-b:
We have a report of a PXP file which has a sweep wave but no config
wave. At this point it is not clear how that can happen, but we need to
allow exporting that data as well.

Reported by Rusty Mann (AI).

This can happen when sweep rollback is used as that was buggy prior to merging #731.


timjarsky (approved):

t-b:
Tests are not running because we changed the CI config in main. Needs https://github.com/AllenInstitute/MIES/pull/739 to be merged before we can proceed here.

t-b:
@timjarsky Will merge then CI passes.

PR #753, opened 2020-11-26
PA: Speed optimization in PA_LayoutGraphs

MichaelHuth:
In PA_LayoutGraphs the trace plot is set to draw lines between points.
Profiling shows that this uses up a lot of time.
On the other hand drawing lines between points is the Igor default.
So unless a user changes the drawing mode, this change has no effect
but a notable speedup.

Found with a better adapted profiling routine.


t-b (approved):

timjarsky:
@MichaelHuth So, explicitly setting the default graph mode is slower than relying on the default mode. Do I have this right?

How much faster is it?

[Nice find BTW]

timjarsky (approved):

PR #760, opened 2020-12-02
PA_Tests.pxp: Make it usable for CI

t-b:
- Removed non-standard igor symbolic paths
- Removed superfluous folders
- Cleaned up internal procedure window

Can you check that the tests pass locally on your side as well?


MichaelHuth:
Tests run through here.

Issue #752, opened 2020-11-24
clip average wave when used with time alignment


timjarsky:
The average trace may have end effects [1] when it is computed using the non-overlapping regions of time aligned traces.

Let's restrict the average trace to points where all the individual traces overlap (when using the time alignment feature).

[1]
![image](https://user-images.githubusercontent.com/5693697/100164457-c0261900-2e6c-11eb-9062-e3e6a1cb0191.png)

PR #763, opened 2020-12-02
PA: With TA the average wave is restricted to common range of all traces

MichaelHuth:
With time alignment enabled the pulses get shifted against each other,
such that their x position and the maximum position matches.
When average was applied to these sets the different count of pulses
at the left and right boundaries resulted in artefacts due to the
different statistics.

Now the average wave is restricted to the range where all pulses have data.
This is automatically carried over to the deconvolution data.

close https://github.com/AllenInstitute/MIES/issues/752

### DIFF
--- a/Packages/MIES/MIES_PulseAveraging.ipf
+++ b/Packages/MIES/MIES_PulseAveraging.ipf
@@ -1938,7 +1938,7 @@ static Function [STRUCT PulseAverageSetIndices pasi, variable needsPlotting] PA_
 #ifndef PA_HIDE_EXECUTION_TIME
 	execTime_StartLocal = stopmstimer(-2)
 #endif
-		PA_CalculateAllAverages(pasi, mode)
+		PA_CalculateAllAverages(pa, pasi, mode)
 #ifndef PA_HIDE_EXECUTION_TIME
 	execTime_CalculateAllAverages = stopmstimer(-2) - execTime_StartLocal
 #endif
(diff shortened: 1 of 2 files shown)


t-b:
@MichaelHuth please rebase against latest main.

t-b (changes requested):
@MichaelHuth Review:
- Code to restrict the range in PA_TAAdaptAverageWave looks good
- Test looks good
- But I think CA_OPTS_NO_DUPLICATE does not do the right thing here as you now modify the average wave or? And you should not modify a wave from the wave cache if you grabbed it via CA_OPTS_NO_DUPLICATE. Dropping `CA_OPTS_NO_DUPLICATE` for fetching would be easiest.

MichaelHuth:
The cache issue is a valid point. Though I don't think it makes sense to do the duplicate there, as I do a duplicate directly after anyway to generate permanent waves.

Beside that, Setting a new scale on the pulse waves changes the wave modification and thus invalidates the cache. Thus, I need to do the restriction only before I put the average wave into the cache, not as currently done as after touch.

MichaelHuth:
The other thing I talked about, where I thought was wrong was actually right. It was just some tricky code to allow a WAVE/Z in the function head of PA_StoreMaxAndUnitsInWaveNote (Because WAVE/WAVE/Z does not work).

t-b (approved):

timjarsky (approved):

PR #765, opened 2020-12-02
Bugfix/ni acquisition pa tests

### LINKED ISSUE #712: DA_Ephys Panel: On opening a new panel with already acquired data on the same device, Next Sweep starts at 0
If the next sweep is not set manually a runtime error is generated after the next data acquisition was run as it tries to overwrite the already present sweep 0.

DAEphys should start with the sweep AFH_GetLastSweepACquired, if finite, instead of 0.

### LINKED ISSUE #740: DA_Ephys Panel: Previous HS association prevents new data acquisition, when old HS stays unused.
When re-opening a DA_Ephys panel it sets the previous HS association in the hardware tab.

Set the case I had HS 0 associated to DA 0, AD 0 and other HS associated to their respective channels as well (e.g. HS 1 and 3) in a previous acquisition. Then I close the DA_Ephys panel and re-open it. The previous HS associations are restored.

After device lock I clear the association of HS 0. In the Data Acquisition tab I enable HS 1 and 3 only, setup remaining stuff and try to start an acquisition. It will assert with "Invalid column".

The problem is: The channelClampMode wave has HS 0, 1, 3 setup from the first time panel. The "Clear Association" button does not update the channelClampMode wave. Only checking/unchecking the HS in the DataAcquisition tab updates the wave.
On panel reopening the all HS are unchecked there, when I check 1 and 3, then 1 and 3 is added to the channelClampMode wave, which is OK, but the information for HS 0 is still there and never removed.
It would be removed if I would have unchecked HS 0, but it is already unchecked.

Current workaround:
After reopening the DA_Ephys panel:
For all HS where the association is cleared in the hardware tab check and then uncheck the HS in the Data Acquisition tab. This removes the data from the channelClampMode wave.

t-b:
@MichaelHuth This fixes the issue here with acquiring data again in the PA_Tests.pxp experiment.

@timjarsky This PR fixes some bugs which surface if you acquired data, close the panel, reopen it, lock it again, and acquire data again. This is probably only a developer issue, but still quite annoying.

Close #712.

### DIFF
--- a/Packages/MIES/MIES_DAEphys.ipf
+++ b/Packages/MIES/MIES_DAEphys.ipf
@@ -4736,6 +4736,10 @@ Function DAP_LockDevice(panelTitle)
 		SetupBackgroundTasks()
 		CtrlNamedBackground _all_, noevents=1
 		UploadCrashDumpsDaily()
+		// avoid problems with IP not keeping the dimension labels
+		// of columns when we have no rows
+		// we kill the wave here so that it is recreated properly
+		KillOrMoveToTrash(wv = GetDQMActiveDeviceList())
 	endif
 
 	WAVE deviceInfo = GetDeviceInfoWave(panelTitle)
--- a/Packages/MIES/MIES_DAEphys.ipf
+++ b/Packages/MIES/MIES_DAEphys.ipf
@@ -4642,8 +4642,7 @@ static Function DAP_UpdateDataFolderDisplay(panelTitle, locked)
 	GroupBox group_Hardware_FolderPath win = $panelTitle, title = title
 End
 
-Function DAP_LockDevice(panelTitle)
-	string panelTitle
+Function DAP_LockDevice(string win)
 
 	variable locked, hardwareType, headstage
 	string panelTitleLocked, msg
@@ -4654,7 +4653,7 @@ Function DAP_LockDevice(panelTitle)
 		DEBUGPRINT_OR_ABORT("The MIES version is unknown, locking devices is therefore only allowed in debug mode.")
 	endif
 
-	panelTitleLocked = GetPopupMenuString(panelTitle, "popup_MoreSettings_Devices")
+	panelTitleLocked = GetPopupMenuString(win, "popup_MoreSettings_Devices")
 	if(windowExists(panelTitleLocked))
 		DoAbortNow("Attempt to duplicate device connection! Please choose another device number as that one is already in use.")
 	endif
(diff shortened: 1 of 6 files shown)


t-b:
@MichaelHuth This will not work. I'll fix it before assigning you again.

t-b:
@MichaelHuth I've removed the fix for #740 as that will not work. A proper fix for that is quite complex.

Packages/MIES/MIES_DAEphys.ipf
```diff
@@ -1683,18 +1683,36 @@ Function DAP_SweepRollback(string paneltitle, variable sweepNo, variable newSwee
...
+static Function DAP_UpdateSweepLimitsAndDisplay(string panelTitle, [variable initial])
 
 	string panelList
 	variable sweep, nextSweep, maxNextSweep, numPanels, i
 
 	panelList = GetListofLeaderAndPossFollower(panelTitle)
 
-	if(DAP_DeviceIsLeader(panelTitle))
-		sweep = DAG_GetNumericalValue(panelTitle, "SetVar_Sweep")
+	if(ParamIsDefault(initial))
```
MichaelHuth, line 1693: initial = ParamIsDefault(initial) ? 0 : !!initial

would be shorter.

Packages/MIES/MIES_DAEphys.ipf
```diff
@@ -1683,18 +1683,36 @@ Function DAP_SweepRollback(string paneltitle, variable sweepNo, variable newSwee
...
+	else
+		initial = !!initial
+	endif
+
+	if(initial)
+		// we are not implementing sweep adjustment for yoked devices
+		if(!DeviceHasFollower(panelTitle) && !DeviceIsFollower(panelTitle))
+			sweep = AFH_GetLastSweepAcquired(panelTitle) + 1
+			if(IsFinite(sweep))
+				SetSetVariable(panelTitle, "SetVar_Sweep", sweep)
```
MichaelHuth, line 1704: Why not use PGC_SetAndActivateControl ?

MichaelHuth (approved):
Just two minor things.

Packages/MIES/MIES_DAEphys.ipf
t-b: Yeep, but I tend to think that if/else is more readable.

Packages/MIES/MIES_DAEphys.ipf
t-b: Because that will call the GUI control procedure which will think that we are rolling back the sweep, see https://github.com/AllenInstitute/MIES/blob/0b0cf1569a1523aebae2509a0005b35623de7cf1/Packages/MIES/MIES_DAEphys.ipf#L1631. And we don't want to do that.

timjarsky (approved):

Issue #749, opened 2020-11-23
Update DataBrowser to use grouped plotting of traces


MichaelHuth:
The grouped plotting of traces increases the drawing speed a lot. This should also be used for the regular DataBrowser trace plot.

PR #767, opened 2020-12-04
DataBrowser: Added grouped drawing when sweeps are added.

MichaelHuth:
This increases the speed when the DataBrowser graph is updated.

The same logic as in the grouped drawing for the PA plot is used,
with the expension that a structure is introduced to carry the draw
information. This allows to draw the graph elements at a later point in
the code.

close https://github.com/AllenInstitute/MIES/issues/749

### DIFF
--- a/Packages/MIES/MIES_Constants.ipf
+++ b/Packages/MIES/MIES_Constants.ipf
@@ -1331,3 +1331,16 @@ Constant PA_PROPERTIESWAVES_INDEX_PULSE = 0
 Constant PA_PROPERTIESWAVES_INDEX_PULSENOTE = 1
 
 /// @}
+
+/// @name Header labels for draw groups within the json of a BufferedDrawInfo structure
+///       Currently this method is only used in @sa CreateTiledChannelGraph
+/// @anchor BDIHeaderLabels
+///
+/// @{
+
+StrConstant BUFFEREDDRAWAPPEND = "AppendToGraph"
+StrConstant BUFFEREDDRAWLABEL = "Label"
+StrConstant BUFFEREDDRAWHIDDENTRACES = "HiddenTraces"
+StrConstant BUFFEREDDRAWDDAQAXES = "dDAQAxes"
+
+/// @}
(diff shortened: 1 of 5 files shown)


t-b:
@MichaelHuth Review:
- missing JSON_Release for BufferedDrawInfo.jsonID
- Why is bdi optional for CreateTiledChannelGraph and friends? Why not always make it faster? I'm okay with not making the oodDAQ case though, as that would involve postponing the axis enabling/scaling as well.
- You also need to handle AR_UpdateTracesIfReq correctly if you are doing
  buffered draw. Does that still work?
  If it is easier we can also not do buffered draw if that is enabled. Maybe
  this needs a member in BufferedDrawInfo which says enabled/disabled?
- Can you explain why `ModifyGraph/W=$graph hideTrace($trace)=1` in CreateTiledChannelGraph does still work?

MichaelHuth:
* bdi.jsonID is released at the end of TiledGraphAccelerateDraw.
* bdi is optional because the acceleration only make sense when sweeps are added, so for AddSweepToGraph and UpdateSweepInGraph. However all draw calls in the loop from OVS_EndIncrementalUpdate have to be buffered. And other calls in between may not have any dependency on the drawn content. So having a fully generalized approach here requires gathering of all single sweep based draw information to be able to execute dependent functions after the buffered draw.
For UpdateSweepInGraph this would add e.g. CursorState and AxisRange backup/restore.
* I don't see where the AR part depends on the drawn traces from CreateTiledGraph.
* I fixed the hideTraces problem.

t-b:
@MichaelHuth Okay, good points. Let's leave it like it is.

t-b:
@MichaelHuth IIRC this code is obsolete once IP9 is mandatory or?

t-b (dismissed):

t-b:
@MichaelHuth It it's done please assign Tim.

t-b (approved):

timjarsky (approved):

Issue #729, opened 2020-11-13
PulseAveraging: Removing a HS in OVS from a sweep lets the plot ignore the associated channel only.


MichaelHuth:
Currently when setting in OVS a HS to removed from a sweep then only the associated channel is removed from the plot (i.e. on the horizontal). However other HS still see the removed HS (on the vertical) and there is no change.

@timjarsky Is it intentional (in respect to the data analysis) that there is in the vertical direction no change? If I would see the whole plot similar to a cross correlation I would expect a change in horizontal as well as vertical?

timjarsky:
@MichaelHuth Are you asking if headstage removal (in OS) is being used, what happens to data on other headstages from the same sweep?

If yes, in the PA plot, the corresponding data from the same sweep on other headstages should also be removed.

In the data browser, I think we want to give the user to remove the corresponding data. For our experimental use case, we want the corresponding data removed because they are related, but I can envision use cases where data on different headstages is only related in time.

Does that make sense?

t-b:
> If yes, in the PA plot, the corresponding data from the same sweep on other headstages should also be removed.
We are currently not removing in main either. So this is something to follow up on after we have merged the current PA plot work.

PR #769, opened 2020-12-04
Fixed: PA plot HS removal now removes also the same regions in the vertical direction

### LINKED ISSUE #737: PulseAveraging: In OVS the HS removal parser does not handle non finite values or random text
Currently an assertion "Unexpected state" is thrown.

Such input as filter should just return no hits, as when entering a non existent HS number.

MichaelHuth:
Previously HS removal only removed the associated channel of the set sweep but left all other regions in the plot.

Removed expected failure from test.

close https://github.com/AllenInstitute/MIES/issues/729
close #737

### DIFF
--- a/Packages/MIES/MIES_PulseAveraging.ipf
+++ b/Packages/MIES/MIES_PulseAveraging.ipf
@@ -628,6 +628,8 @@ static Function [STRUCT PulseAverageSetIndices pasi] PA_GenerateAllPulseWaves(st
 		regionList = AddListItem(num2istr(regions[i]), regionList, PA_PROPERTIES_STRLIST_SEP, inf)
 	endfor
 
+	Make/FREE/N=(MINIMUM_WAVE_SIZE)/WAVE headstageRemovalPerSweep
+
 	// There is one case where we generate errorneous output:
 	// If we have multiple sweeps that are acquired with more than 1 HS and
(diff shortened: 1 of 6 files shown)


Packages/MIES/MIES_PulseAveraging.ipf
```diff
@@ -743,6 +743,11 @@ static Function [STRUCT PulseAverageSetIndices pasi] PA_GenerateAllPulseWaves(st
 				sweepList = AddListItem(sweepNoStr, sweepList, PA_PROPERTIES_STRLIST_SEP, inf)
 			endif
 
+			WAVE/Z activeHS = OVS_ParseIgnoreList(win, sweepNo=sweepNo)
+			if(WaveExists(activeHS) && activeHS[region] == 0)
+				continue
+			endif
+
```
t-b, line 750: I've not played around with it. This looks like to be technically the correct solution. Nice that it is so short. But isn't that dead slow as you call that for every pulse?

t-b: I'm taking over this PR.
- Parse ignore list on input to OVS and store parsed in waves
- Waves need to be cheap to query with sweepNo and experiment from PA_GenerateAllPulseWaves

timjarsky (approved):

t-b (approved):

PR #778, opened 2020-12-17
PGC_SetAndActivateControl: Give a better assertion message

### LINKED ISSUE #774: PGC_SetAndActivateControl: Add more helpful assertion messages on errors
See https://aiephys.slack.com/archives/C01CS4T8MGA/p1608062998004200?thread_ts=1608062962.003900&cid=C01CS4T8MGA.

t-b:
This assertion message is in the top 10% for users. So let's be a little
nicer and tell them which window and control was missing.

Close #774.

### DIFF
--- a/Packages/MIES/MIES_ProgrammaticGUIControl.ipf
+++ b/Packages/MIES/MIES_ProgrammaticGUIControl.ipf
@@ -235,7 +235,10 @@ Function PGC_SetAndActivateControl(win, control, [val, str, switchTab])
 
 	// call only once
 	ControlInfo/W=$win $control
-	ASSERT(V_flag != 0, "Non-existing control or window")
+	if(!V_flag)
+		ASSERT(WindowExists(win), "The panel " + win + " does not exist.")
+		ASSERT(0, "The control " + control + " in the panel " + win + " does not exist.")
+	endif
 	controlType = abs(V_flag)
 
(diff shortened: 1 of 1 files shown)


Packages/MIES/MIES_ProgrammaticGUIControl.ipf
```diff
@@ -235,7 +235,10 @@ Function PGC_SetAndActivateControl(win, control, [val, str, switchTab])
 
 	// call only once
 	ControlInfo/W=$win $control
-	ASSERT(V_flag != 0, "Non-existing control or window")
+	if(!V_flag)
+		ASSERT(WindowExists(win), "The panel " + win + " does not exist.")
```
timjarsky, line 239: @t-b it'll be a bit tedious for the user if the win and control don't exist but I can't think of a better way to do it.

timjarsky (approved):

Packages/MIES/MIES_ProgrammaticGUIControl.ipf
t-b: @timjarsky I can fix that ;)

t-b: @timjarsky Sorry I amend that statement. This is not fixable as if the windows does not exist I can't check for the control.

Issue #67, opened 2018-09-10
Rename functions/constants which are not ITC specific


t-b:
Since the introduction of NI hardware support some functions names, constants, datafolder names still have "ITC" in them, but are now used for NI as well.

Labnotebook entries should not change.

t-b:
And also rename `GetHardwareDataWave` to `GetDAQDataWave` and `GetITCChanConfigWave` to `GetDAQConfigWave`.

PR #780, opened 2020-12-18
Feature/separate tp daq waves

t-b:
- [x] Reuse DAQDataWave_TP directly if possible, this keeps WaveData pointer fixed

This PR introduces separate DAQDataWaves waves for DAQ and TP. And also renames them as mentioned in https://github.com/AllenInstitute/MIES/issues/67.


timjarsky (approved):
This is passing testing on the itc18 and NI devices. It also passes my basic manual tests. I recommend we merge.

PR #786, opened 2020-12-23
SetupBackgroundTasks: Don't allow dialogs during background function …

t-b:
…execution

Since forever we did not set the dialogsOK parameter for the background
tasks. And the default is dialogsOK = 1 which means that dialogs can run
during background execution.

The Igor help is pretty clear on that, that this can cause crashes when
removing traces from a graph etc.

And that can happen as we call SWS_SaveAcquiredData from the DAQ
background functions and user analysis function as well.

So let's disable that.

Found with user uploaded crash dumps and help from Adam Light
(WaveMetrics).

### DIFF
--- a/Packages/MIES/MIES_MiesUtilities.ipf
+++ b/Packages/MIES/MIES_MiesUtilities.ipf
@@ -6169,13 +6169,13 @@ End
 
 // @brief Common setup routine for all MIES background tasks for DAQ, TP and pressure control
 Function SetupBackgroundTasks()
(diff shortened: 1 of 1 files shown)


timjarsky:
@t-b with this PR, opening a dialog will pause repeated acquisition. Would it be possible to be selective about the application 'dialogsOK = 0' such that data acq continues but there aren't any window updates?

t-b:
@timjarsky The only way to do that is to split the background tasks up into ones which use `dialogsOK=0` and ones updating oscilloscope windows with `dialogsOK=1`. I'm not sure this is really worth the additional complexity.

timjarsky (approved):

PR #791, opened 2021-01-15
WBP_ButtonProc_OpenAnaFuncs: Open the function directly if possible

t-b:
If we have selected a analysis function we should just jump to it,
instead of blindly opening UserAnalysisFunctions.ipf.

This is especially convenient for the builtin PSQ/MSQ analysis
functions.


timjarsky (approved):

timjarsky:
@t-b very cool!

PR #804, opened 2021-01-25
Bugfix/databrowser ddaq display mode

### LINKED ISSUE #803: PA plot and dDAQ option in databrowser does not work together
dDAQ turned on:

grafik

dDAQ turned off:

grafik

grafik

t-b:
Close #803

### DIFF
--- a/Packages/MIES/MIES_BrowserSettingsPanel.ipf
+++ b/Packages/MIES/MIES_BrowserSettingsPanel.ipf
@@ -1205,7 +1205,7 @@ Function/WAVE BSP_GetTextualValues(string win, [variable sweepNumber])
 End
 
 /// @brief Return the wave with the selected channels respecting the overlay
-/// sweeps headstage ignore list. The wave has the same layout as B
+/// sweeps headstage ignore list. The wave has the same layout as BSP_GetChannelSelectionWave.
 Function/WAVE BSP_FetchSelectedChannels(string graph, [variable index, variable sweepNo])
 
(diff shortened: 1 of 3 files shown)


timjarsky:
@t-b this works but it uncovered a related issue (dDaq doesn't work with averaging and overlay sweeps).

t-b:
@timjarsky Fixed

timjarsky:
@t-b old average trace isn't getting cleaned up when a new sweep is added with OS.

image

t-b:
@timjarsky Fixed.

timjarsky:
@t-b another related issue: dDAQ doesn't handle headstage removal with OS.

no  dDAQ:
image



dDAQ

image

t-b:
Hmm something is wrong here. I'll look into it.

t-b:
I can't reproduce that. And also looked at the code and I'd say it makes sense. I've pushed some unrelated fixes only. 

Can you retry with the latest version?

timjarsky:
@t-b 
with HS removal:
image
without:
image

same test file I have been using for PA plot updates:
image

t-b:
Next try. I've removed a "fix" fromt his PR which introduced that error.

timjarsky:
@t-b dDAQ is looking good"

image

I think I can still find an edge case where trying to use headstage removal in dDAQ mode doesn't remove the sweep. Let's try it out the next time we chat. I suspect it will be an issue for a different PR but I want to check before merging this one.

timjarsky (approved):

PR #831, opened 2021-02-19
PA_SetVarProc_Common: Allow escaping NaN for number of spikes controls

### LINKED ISSUE #820: PA plot: Allow user to go from `NaN` number of spikes to 1 with arrow buttons/keys

t-b:
We start with the value of NaN to indicate that we don't care about the
number of spikes. But that also makes the up/down buttons useless as NaN
+/- 1 is always NaN again.

So let's use 1 in this case instead.

Close #820.


timjarsky:
@t-b control can't be manually set back to NaN (as in master).

t-b:
@timjarsky This does get a bit hacky..

PR #841, opened 2021-02-23
Feature/enhance reach target voltage

### LINKED ISSUE #816: ReachTargetVoltage: Add DoPrompt for all active headstages
- [x] This queries the new autobias values, should be prefilled with -70mV. The autobias valures are applied in POST_SET_EVENT but queried in PRE_DAQ_EVENT. Pressing cancel in the dialog stops DAQ with returning 1. Only done when "Enable Indexing" analysis parameter is set.

- [x] Verify that POST_SET_EVENT works with locked/unlocked indexing and if yes adapt the comment at https://github.com/AllenInstitute/MIES/blob/adb2eefe318b3ffbd6c506818e341a621ccb8801/Packages/MIES/MIES_AnalysisFunctions.ipf#L38-L40

t-b:
Close #816.

### DIFF
--- a/Packages/MIES/MIES_AnalysisFunctions.ipf
+++ b/Packages/MIES/MIES_AnalysisFunctions.ipf
@@ -1007,21 +1007,28 @@ Function ReachTargetVoltage(string panelTitle, STRUCT AnalysisFunction_V3& s)
 	switch(s.eventType)
 		case PRE_DAQ_EVENT:
 			WAVE targetVoltagesIndex = GetAnalysisFuncIndexingHelper(panelTitle)
-			WAVE ampParam = GetAmplifierParamStorageWave(panelTitle)
-			WAVE statusHS = DAG_GetChannelState(panelTitle, CHANNEL_TYPE_HEADSTAGE)
 
 			targetVoltagesIndex[s.headstage] = -1
 
-			if(DAG_GetHeadstageMode(panelTitle, s.headstage) != I_CLAMP_MODE)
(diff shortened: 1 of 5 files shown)


t-b:
@timjarsky Go for it!

timjarsky:
@t-b panel pops up twice and auto bias target voltages are not being applied (after reach target voltage / target v).

t-b:
@timjarsky How could you test with multipe headstages ;)? I'll look into it.

t-b:
@timjarsky Next try.

timjarsky:
@t-b the auto-bias target V is being set but the TP isn't running between targetV and the next stim set. So, the autobias never has the opportunity to set the resting voltage of the cell before proceeding to the next stimset.

t-b:
@timjarsky I can adjust the ITI. To what value?

t-b:
@timjarsky Found one more todo. I still need to turn off "send to all amps".

t-b:
Done.

timjarsky:
@t-b 
1) send to all amps is being checked at some point at or around the first index step
2) tp during ITI also got unchecked at one point - not sure how that happened
3) TP cache gets corrupted if acq is manually stopped.

timjarsky:
@t-b I'm guessing that the strange symptoms in the previous post may all being explained if the cache were corrupted? I'm going to test again to see if I can reproduce.

t-b:
@t-b I'm investigating the send to amps thing. Nothing should uncheck "TP during ITI".

t-b:
I've fixed send to amps.

timjarsky (approved):

PR #852, opened 2021-03-01
Bugfix/wavebuilder more help

t-b:
I had to shorten the message to make it work on IP8.

- [x] Wait for WM support question reply, add maybe missing `<html>` around the help entry.

For reference here is the original message:
```html
Delta operation mode: Different operation modes are supported. As these are delta modes they are always applied starting
from the second sweep. For all modes except explicit the delta of the previous step is added onto value for the current
step before recalulating the delta. We therefore only document here how we calculate the delta.
<ul>
<li>None: No delta mode</li>
<li>Multiplier: delta *= multiplier</li>
<li>Logarithmic: delta = log(delta)</li>
<li>Squared: delta = delta^2</li>
<li>Power: delta = delta^(multiplier)</li>
<li>Alternate: delta *= -1</li>
<li>Explicit: delta = listitem</li>
</ul>
```

### DIFF
--- a/Packages/MIES/MIES_Constants.ipf
+++ b/Packages/MIES/MIES_Constants.ipf
@@ -592,7 +592,7 @@ Constant HARDWARE_DAC_EXTERNAL_TRIGGER = 0x1
 /// Used to upgrade the GuiStateWave as well as the DA Ephys panel
 Constant DA_EPHYS_PANEL_VERSION           = 51
 Constant DATA_SWEEP_BROWSER_PANEL_VERSION = 26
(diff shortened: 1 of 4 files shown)


timjarsky:
@t-b what do you think about adding help for `setvar_explDeltaValues_T28_ALL`?

Something like:
semi-colon separated list of ITI deltas
ITI is the sum of all previous deltas, plus the initial ITI.
First list item is applied to the second ITI. i.e., second ITI = first list item + initial ITI.

t-b:
@timjarsky Done.

timjarsky (approved):

PR #899, opened 2021-04-01
Fix package folder location on disk

t-b:
Since c87daee1 (MIES_PackageSettings.ipf: Extract settings folder
creation into separate function, 2021-03-09) we created the log file and
the JSON settings file in the folder

C:\Users\$user\AppData\Roaming\WaveMetrics\Igor Pro 8\Packages

instead of

C:\Users\$user\AppData\Roaming\WaveMetrics\Igor Pro 8\Packages\MIES

.

The reason is that the path contained an additional colon (":") which in
IP lingua means "go down one folder". But that was only the case in IP8,
IP9 has a bug where this was is not done.

Remove the additional colon so that it works with all IP versions.

@timjarsky I don't have a automatic fix, which means that we move the files from the wrong location to the right location, included. Should I?


timjarsky:
@t-b auto-fix sounds good to me.

timjarsky (approved):

t-b:
@timjarsky Done.

timjarsky (approved):

PR #900, opened 2021-04-01
fix: conclude OpenAboutDialog function with "End" instead of "EndMacro"

MichaelHuth:
Code coverage tracing stumbled over this.


t-b:
@MichaelHuth Please rebase. It might be that this is already fixed.

MichaelHuth:
Yes it is already fixed in main.

PR #906, opened 2021-04-09
MIES_AnalysisFunctions_PatchSeq.ipf: Make shortcut on passed baseline QC work

t-b:
We are always writing the baselineQC value as headstage dependent data
into the labnotebook. But for PSQ_DAScale/PSQ_Rheobase and PSQ_Ramp we
queried the headstage independent value when shortcutting if we are
done.

This is now fixed.

The bugs were introduced in:
- f3fbfc8f (PSQ_DA_SCALE: Postpone sweep passed decision, 2019-04-13)
- 3bcdaadb (Add Rheobase patch seq analysis function with tests, 2017-12-10)
- 3a7ef027 (PatchSeq Ramp: Add analysis function, 2018-03-20)

The fallout from the bug is slightly worse runtime, PSQ_EvaluateBaselineProperties would properly return early as it checks each chunk.


t-b:
CI passed if you look in bamboo. I've fixed the problem that this is not shown here in github.

timjarsky:
@t-b could the lab notebook wrappers warn/assert when checking for hs dependent info in the independent layer and vice versa?

timjarsky (dismissed):

t-b:
> could the lab notebook wrappers warn/assert when checking for hs dependent info in the independent layer and vice versa?

I'll have a look.

t-b:
@timjarsky I've added something which is currently only done in CI. Let's see what the mighty bamboo gods have to say.

t-b:
@timjarsky Go for it!

timjarsky (approved):

Issue #923, opened 2021-04-20
Sweepbrowser: Add support for dashboard


t-b:
It is okay that we limit ourselves to sweeps from one experiment.

PR #579, opened 2020-07-20
Add chirp analysis function

t-b:
Succesor of #531.

- [x] Update flowchart
- [x] Update dashboard
- [x] resolve todos in code
- [x] Add tests for new spikeCheck code
- [x] Cleanup commits
- [x] Add CheckDashboard calls

Close #496.

PR #931, opened 2021-04-29
Sweepbrowser dashboard

### LINKED ISSUE #553: Dashboard: Add explanation as tooltip using helpWave keyword of listbox

### LINKED ISSUE #602: Dashboard: Add enable button
For users which do not use the dashboard it makes sense to allow disabling it, as it takes quite a bit of time to compute the information.

t-b:
Close https://github.com/AllenInstitute/MIES/issues/923
Close #553
Close #602

Requires #579 to be merged.

### DIFF
--- a/Packages/MIES/MIES_AnalysisBrowser.ipf
+++ b/Packages/MIES/MIES_AnalysisBrowser.ipf
@@ -2352,6 +2352,7 @@ Function AB_ButtonProc_LoadSweeps(ba) : ButtonControl
 			oneValidSweep = AB_LoadFromFile(AB_LOAD_SWEEP, sweepBrowserDFR = dfr)
 			SVAR/SDFR=dfr graph
 			if(oneValidSweep)
+				AD_Update(graph)
 				panel = BSP_GetSweepControlsPanel(graph)
 				PGC_SetAndActivateControl(panel, "button_SweepControl_PrevSweep")
 			else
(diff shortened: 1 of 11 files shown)


timjarsky:
@t-b I really like how the dashboard allows the user to easily display the sweeps from a stimset run. Could there be a default analysis function that could be attached to a stimset (by default?) that just passes the sweeps? This way any stim set could take advantage of the dashboard display features.

t-b:
@timjarsky Sure we can allow using the dashboard for all stimsets. But I would not bother with a default analysis function. I can just add every sweep to both passed and failed if it does not have an analysis function attached.

Let's see if I can make that happen today before the release.

t-b:
@timjarsky Done. I'll get the other two dashboard issues done as well now.

timjarsky:
@t-b failed sweeps are displaying;nice! DB dashboard now shows all acquired stimsets BUT the DB only shows the first sweep in the stimset when selecting the stimset from the dashboard (see image; all sweeps after sweep 61 are from the same stimset). This may be related to the fact that the dashboard is updating after the first sweep is acquired.

image

t-b:
@timjarsky Can you upload the data? Thanks.

timjarsky:
@t-b is it necessary for each headstage to have its own DB dashboard row for the same set of sweeps? I'm 100% okay with merging as is but let's talk about alternatives.

image

t-b:
> @t-b is it necessary for each headstage to have its own DB dashboard row for the same set of sweeps? I'm 100% okay with merging as is but let's talk about alternatives.

The reason is that each headstage has its own stimset and therefore analysis function and therefore I'm going sweep-by-sweep and headstage-by-headstage basis. One could do some kind of compression where we only have one entry per sweep if all headstages have the same stimset/analysis function.

timjarsky (dismissed):

timjarsky (approved):

PR #940, opened 2021-05-05
NWB_CheckForMissingSweeps: Handle no sweeps gracefully

t-b:
Calling NWB_ExportAllData where at least one device has content but no
sweeps currently leads to NWB_CheckForMissingSweeps bugging out as
GetSweepsWithSetting returns a null wave.

Bug introduced in 6a432cb7 (Add code to allow recovering deleted sweeps,
2021-04-19).

- [x] Add test
- [x] Think about fix again

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,15 +1,12 @@
 [submodule "Packages/unit-testing"]
-	fetchRecurseSubmodules = false
 	path = Packages/unit-testing
 	shallow = true
 	url = https://github.com/byte-physics/igor-unit-testing-framework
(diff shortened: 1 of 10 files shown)


timjarsky (dismissed):

t-b:
@timjarsky Ready. I've slightly tweaked the fix to be more robust. And added a test for it.

I also fixed the issue with debugging asking to save and also fixed the nwb export version selector.

timjarsky (approved):

t-b:
Will merge once CI passes.

Issue #911, opened 2021-04-12
AnalysisBrowser: Allow to resave loaded data


t-b:
Original discussion: https://u01-human-brain.slack.com/archives/CAMJ4T072/p1615802292019600

For some measurements the original PXP files were lost, but NWBv1 files are available. It would be nice if we, after loading them into the AnalysisBrowser, are able to save them in NWBv2. This should be possible as we have the raw data itself, the stimsets and all MIES metadata like labnotebook etc.

t-b:
@timjarsky Is it enough if this works interactively via the analysis browser or do we need a fully automatable solution for many many files?

timjarsky:
@t-b interactively is sufficient

t-b:
@timjarsky And I guess it is also okay to not store saved TPs? We don't yet load that in the analysis browser. But of course if now is the time to do it  ;)

timjarsky:
@t-b do we store saved TPs in MIES NWB2 files already?

t-b:
@timjarsky Yes we always store saved TPs in NWB, both v1 and v2.

PR #954, opened 2021-05-12
Store epoch information in NWBv2 files

t-b:
- [x] Add ihf help for XOP
- [x] Add IP code to read and write epochs
- [x] Add tests when exporting into NWBv2 that the epochs are written properly
- [x] Hand out test files to Tim for all PSQ analysis functions 
- [x] Add documentation
- [x] Cleanup commits
- [x] Must be merged after #918 and #940
- [x] add pynwb read tests
- [x] add assertions for offset/sizes in read epochs code
- [x] Merge https://github.com/AllenInstitute/IPNWB/pull/28 and update submodule commit

Close #569.

PR #918, opened 2021-04-15
Async NWB writing

t-b:
- [x] Check if the code works
- [x] Add documentation and tests for new ASYNC functionality
- [x] Code cleanup
- [x] Add documentation how NWB writing works in the code
- [x] find a proper solution for WaveModCountWrapper issue in preemptive threads
- [x] Wait for #942, #940 being merged
- [x] ~~Decide if we should store stimsets/tpstorage as well. Order is Sweeps, LBN/userComment, stimset, tpstorage~~
- [x] Write all stimsets in NWB_PrepareExport
- [x] Check if we need to clear the preemptive threads datafolder hierarchy on each invocation

PR #960, opened 2021-05-21
Allow resaving NWBv1 files into NWBv2 files

t-b:
- [x] Code cleanup
- [x] Proper commits
- [x] Move marked code into IPNWB once #918 and #954 is done
- [x] Collapsed experiment names don't work
- [x] Only load the stored TPs for reexport
- [x] ~~Do we need to load the stored TPs from pxp as well~~ No
- [x] Reuse overwrite checkbox to avoid asking the user
- [x] Add NWB version/PXP to the analyisbrowser list box

Close https://github.com/AllenInstitute/MIES/issues/911

### DIFF
--- a/Packages/Conversion/MIES_MassExperimentProcessing.ipf
+++ b/Packages/Conversion/MIES_MassExperimentProcessing.ipf
@@ -135,11 +135,6 @@ static Function PerformMiesTasks(outputFilePath)
 
 	printf "Free Memory: %g GB\r", GetFreeMemory()
 
-	if(FileExists(outputFilePath))
-		print "Output file already exists, skipping!"
-		return 0
-	endif
-
 	folder = GetFolder(outputFilePath)
 
 	if(!FolderExists(folder))
--- a/Packages/MIES/MIES_AnalysisBrowser.ipf
+++ b/Packages/MIES/MIES_AnalysisBrowser.ipf
@@ -729,6 +729,40 @@ Function AB_LoadTPStorageFromNWB(nwbFilePath, expFolder, device)
 	H5_CloseFile(h5_fileID)
 End
 
+static Function AB_LoadStoredTestpulsesFromNWB(nwbFilePath, expFolder, device)
+	string nwbFilePath, expFolder, device
+
+	variable h5_fileID, testpulseGroup, numEntries, i
+	string dataFolderPath, list, name, groupName
+
+	WAVE/WAVE wv = GetAnalysisStoredTestPulses(expFolder, device)
+
+	if(DimSize(wv, ROWS) > 0)
+		return NaN
+	endif
+
+	h5_fileID = H5_OpenFile(nwbFilePath)
+
+	groupName = "/general/testpulse/" + device
+	testpulseGroup = H5_OpenGroup(h5_fileID, groupName)
+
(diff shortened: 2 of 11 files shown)


timjarsky:
@t-b analysis browser won't load NWB V1 in this PR (loads in main). I uploaded the NWB file (2021_05_25_160129.nwb) to the FTP. The file was generated from this PR/branch and during acquisition.

t-b:
@timjarsky Thanks. Should be fixed.

t-b:
@timjarsky Ready.

timjarsky:
@t-b Is there a trick to selecting a file in the analysis browser for resaving as NWB? I keep getting:

```
  Please select an experiment to use this feature
```

image

t-b:
@timjarsky Thanks, fixed.

timjarsky:
@t-b would it be possible to have a 'resave all' to NWB option?

t-b:
@timjarsky Sure.

t-b:
@timjarsky Done. I'm now always converting all loaded experiments.

t-b:
@timjarsky This is how the output now looks like:

```
  ####
  Folder: E:\tim-data\nwbv1-loading-analysisbrowser\
  NWbv1: Experiment35-compressed.nwb
  NWbv2 (new): Experiment35-compressed-converted.nwb
  ####
```

timjarsky:
@t-b thanks for printing the file path with slashes instead of colons. Everything looks good. Can you remind me of the differences between an NWB2 file generated from a NWB1 file vs a .pxp?

timjarsky (dismissed):

timjarsky:
@t-b how much work would it be to add a column to show nwb file version to the analysis browser?

timjarsky:
@t-b what happens if the converted file already exists?

t-b:
> @t-b thanks for printing the file path with slashes instead of colons. Everything looks good. Can you remind me of the differences between an NWB2 file generated from a NWB1 file vs a .pxp?

I haven't really examined it, but I don't think there is a practical difference. Still converting from NWBv1 should only be done if the pxp is not available anymore.

> @t-b what happens if the converted file already exists?

I think we append into the NWB file. This is probably not what we want, or? Is it okay to just overwrite the old one?

> @t-b how much work would it be to add a column to show nwb file version to the analysis browser?

An hour or two.

timjarsky:
> > @t-b what happens if the converted file already exists?
> 
> I think we append into the NWB file. This is probably not what we want, or? Is it okay to just overwrite the old one?

Can we let windows handle this the way it normally does; ask the user what to do?

> 
> > @t-b how much work would it be to add a column to show nwb file version to the analysis browser?
> 
> An hour or two.

If you are up for it, let's do it.

t-b:
- [ ] Reuse overwrite checkbox to avoid asking the user
- [ ] Add NWB version/PXP to the analyisbrowser list box

t-b:
@timjarsky I've implemented honoring overwrite checkbox and not touching the files if they already exist and overwrite is off. And also display the NWB type in the analysisbrowser.

timjarsky:
@t-b beautiful!

image

timjarsky (approved):

PR #975, opened 2021-06-09
Bugfix/975 various bugfixes

t-b:


Packages/MIES/MIES_MiesUtilities.ipf
```diff
@@ -785,6 +785,12 @@ threadsafe static Function [WAVE/Z wv, variable index] GetLastSettingChannelInte
 			WAVE/Z settings = GetLastSetting(values, sweepNo, setting, entrySourceType)
 
 			if(WaveExists(settings))
+				WAVE/T settingsT = settings
+				if((IsNumericWave(settings) && IsNaN(settings[indizes[0]]))          \
+				   || (IsTextWave(settingsT) && !cmpstr(settingsT[indizes[0]], "")))
+					return [$"", NaN]
```
timjarsky, line 791: @t-b cool! I did know it was possible to pass an empty string as wave `$""`

timjarsky: I also didn't know you could do WAVE`/Z` with the multiple return syntax

timjarsky (approved):

Packages/MIES/MIES_MiesUtilities.ipf
t-b: @timjarsky `$""` stands for either a null wave reference or a for a null dfref.

You can  even return more things with MRS, for example structs, which you can return with normal syntax.

PR #976, opened 2021-06-09
Allow debugging threadsafe functions

### LINKED ISSUE #967: Make debugging with threadsafe functions possible
Right now IP does not support using the debugger in threadsafe functions. This is quite a hassle. For allowing to debug threadsafe functions the command `SetIgorOption DisableThreadSafe = 1` allows to turn off threadsafeness globally.

For this to work we need:
- Use a function which does the SetIgorOption call and also sets a define, like EnableEvilMode()
- Use that define to branch to a different code path when Threadstart is used. The async framework will just execute the function from ASYNC_Execute, and the threaded FIFO handling will be done from the background function.
- Add warning printouts so that users will not accidentally use this mode

t-b:
Close #967

### DIFF
--- a/Packages/MIES/MIES_Async.ipf
+++ b/Packages/MIES/MIES_Async.ipf
@@ -108,9 +108,6 @@ End
 /// stopped.
 threadsafe static Function ASYNC_Thread()
 
-	DFREF dfrOut, dfrTemp
-	string datafolder
-
 	for(;;)
 		DFREF dfr = ThreadGroupGetDFR(0, 10)
 		if(!DataFolderExistsDFR(dfr))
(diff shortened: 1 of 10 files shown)


timjarsky:
@t-b we may also need to test this PR together; I've seen a couple Igor lock ups.

I'm also finding that function profiling requires threadsafe support. Should we check if function profiling is included before disabling threadsafe?

image

t-b:
@timjarsky Do you have an additional copy of "Function Profiling" included? All the ones MIES includes are disabled when threading is disabled.

timjarsky:
@t-b I don't think so...

image

Am I looking in the wrong place?

timjarsky (approved):

PR #986, opened 2021-06-17
Make epoch viewing work in sweep browser

### LINKED ISSUE #984: Add support for Sweepbrowser in DC_AddDebugTracesForEpochs()

t-b:
- [x] Wait for WM reply to bugreport about DoIgorMenuMode
- [x] Make this functionality available for all
- [x] Always show after changing sweep when checked
- [x] Grey out for IP8 with help message
- [x] Reorganize visualize controls and top AD/DA/TTL checkboxes so that the users knows which belong together
- [x] Grey out controls which depend on each other (does that interfere with experiment configuration??)
- [x] Add IP9 case/switch patch

This PR has a copy of #1032 included so it should be merged after it.

Close #984.


t-b:
@timjarsky Done.

I've checked and disabled controls are not restored with the experiment configuration. Therefore I had to set some config restore priorities to allow saving and restoring controls which are initially hidden.

It looks now as in:

image

timjarsky:
@t-b very nice! I especially like the placement of the dDAQ checkbox with the slider

t-b:
> I want to approve but there is a surprising interaction between Visualize Epochs and Axes Sacling:EqualY. What do you think about disabling EqualY when visualizing epochs?

I look into it.

t-b:
@timjarsky Done.

timjarsky:
@t-b sweep overlay doesn't clean up old epoch info. e.g., select two sweeps with different epoch info, then deselect the first sweep selected.

t-b:
> sweep overlay doesn't clean up old epoch info. e.g., select two sweeps with different epoch info, then deselect the first sweep selected.

This is most likely incremental OVS update not updating the epoch traces. I'll have a look.

t-b:
@timjarsky Ready for test.

timjarsky (dismissed):

PR #1003, opened 2021-07-06
Databrowser set sweep count selection

### LINKED ISSUE #157: OVS: Add option to choose all sweeps from a stimset and a given "Set Sweep count"
This makes the selection more reliable than offset/step as it selects the fifth planned sweep from a stimset and not the fifth acquired sweep.

Add an additional popup menu.

t-b:
- [x] Add stimulus set cycles selection as well
- [x] Limit "set sweep count" list items to only have elements with 2 or more counts
- [x] Selection takes way too long
- [x] Only show stimsets which were acquired more than once (aka have a set cycle count of 1 or higher)

Close #157.

### DIFF
--- a/Packages/MIES/MIES_Constants.ipf
+++ b/Packages/MIES/MIES_Constants.ipf
@@ -1588,3 +1588,11 @@ StrConstant EPOCH_SHORTNAME_USER_PREFIX = "U_"
 Constant SF_DM_NORMAL = 1
 Constant SF_DM_SUBWINDOWS = 2
 /// @}
+
+/// @name Parameters for GetTTLLabnotebookEntry()
+/// @anchor LabnotebookTTLNames
+///
+/// @{
+StrConstant LABNOTEBOOK_TTL_STIMSETS       = "stim sets"
+StrConstant LABNOTEBOOK_TTL_SETSWEEPCOUNTS = "set sweep counts"
+/// @}
--- a/Packages/MIES/MIES_Constants.ipf
+++ b/Packages/MIES/MIES_Constants.ipf
@@ -635,7 +635,7 @@ Constant ANALYSISBROWSER_PANEL_VERSION    =  1
 /// - Changed names of entries
 /// - Changed units or meaning of entries
 /// - New/Changed layers of entries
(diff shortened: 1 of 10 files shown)


t-b:
@timjarsky I interpret being assigned that I really should dive in?

timjarsky:
@t-b yes, but the TP work is a higher priority.

timjarsky:
@t-b I propose limiting the list of stimsets in the `DA Stimulus Sets and Set Sweep Count`  category to those with 2 or more sweeps. This should prevent the list from getting large and make enhancements GuiPopupMenuExt unnecessary. Thoughts?

t-b:
@timjarsky Done.

timjarsky:
@t-b Limit "set sweep count" list items to only have elements with 2 or more counts doesn't seem to be working; every set sweep count is present.

The menu can be very slow to respond - locks IP up for several seconds.

I can demo

t-b:
@timjarsky Jeep let's demo and chat about this PR. I seem to have misunderstood how you want it as the sweep count limitation works here as I think it should.

Regarding execution speed, is it always slow or only on the first opening? We postpone updating the menu entries to save some time during DAQ and do it therefore only once on first opening.

t-b:
@timjarsky Ready to test again. I've now removed stimsets from set sweep count/set cycle count which never cycled. I can not reproduce the slowness, but I also changed the code a bit so please retry.

timjarsky:
@t-b we want to find stimsets that have been acquired more than once regardless of the RA cycle. In the current config, if the set was acquired on different RA cycles, they are not listed:
image

t-b:
@timjarsky Next try.

timjarsky:
@t-b I loaded 5 sweeps from an experiment and one run of a stimulus set. In this case, I would expect that nothing would be listed in the `DA stimulus sets and set sweep count` menu because there are no repeats in the loaded data. 
image

t-b:
@timjarsky And next try.

timjarsky:
@t-b i'm still seeing the behavior described in my last post (screen share?)

timjarsky (approved):

PR #987, opened 2021-06-17
Fetch the correct Rheobase sweep for PSQ_DAScale

t-b:

PR #1005, opened 2021-07-07
Make baseline checking more robust

### LINKED ISSUE #997: PSQ_EvaluateBaselineProperties does not continue in current fifo data after failed chunk
PSQ_EvaluateBaselineProperties does return NaN for a failed chunk for PSQ_BL_POST_PULSE.

Callers like PSQ_DAScale interpret this return value as not enough data for check and return, This behavior is wrong, as the loop over the available chunks should continue with the next chunk.

Suggested solution:
- Add a new return constant for the failed chunk + PSQ_BL_POST_PULSE condition.
- For that return value continue the loop over the chunks in the calling function.

This bug appears on slower machines, when multiple chunks are available from the fifo buffer on call.

t-b:
- [x] Wait for #987 
- [x] Raise analysis function version.

Close #997.


t-b:
Rebased and raised analysis function versions.

t-b:
Rebased and raised analysis function versions.

timjarsky (approved):

PR #1006, opened 2021-07-13
Feature/1006 epoch fixes

### LINKED ISSUE #989: Epoch info additions and fixes
- Load
[Sst-IRES-Cre.Ai14-579951.06.03.05-compressed.zip](https://github.com/AllenInstitute/MIES/files/6677723/Sst-IRES-Cre.Ai14-579951.06.03.05-compressed.zip) into the analysis browser
- Open Sweep 38
- Use PR #986 
- Enable debug view of epoch info via DC_AddDebugTracesForEpochs()
- Enable "Show Trace Info Tags"

Now you see something like

image

Things to note:
- The sweep ended early but the epoch info covers the full planned DAC
- The unacquired range does not have an epoch
- The special DA/AD areas which we introduced via PSQ_Ramp for ITC hardware are not covered

t-b:
- [x] Add support for user level epochs
   - [x] Tests
   - [x] Documentation 
- [x] Rename the epochs to have more consistent names, see https://github.com/AllenInstitute/ipfx/issues/518#issuecomment-861863364
- [x] Prefer tags over `Name`/`epName` for epoch column 
- [x] Add documentation for "unacquired" epoch and epoch adjusting
- [x] Add special AD/DA epochs for PSQ_Ramp
- [x] Save the epoch version in the LBN. And obviously raise it
- [x] Test for AD/DA suppression epochs in PSQ_Ramp tests
- [x] Needs to wait for #986, #954, #1045
- [x] Merge https://github.com/AllenInstitute/IPNWB/pull/30 and update submodule

Close #989.


t-b:
I've added a preliminary patch for AD/DA suppression epochs. These use the tree level -1 which I termed the user tree level.

image

The two top most epochs below the pink line are the DA (long), AD (short) epochs. We don't have a concept of adding epoch info for AD data so this is slightly hacked here. But I can live with that.

t-b:
I talked this over with @MichaelHuth and he suggested to have an explicit epoch tag "ChannelType=XX" to distinguish epoch for DA data (what we have until now only) and epochs for AD data ("AD suppression"). This would also allow us to introduce TTL epochs later on more easily. Note to self: epochsWave needs to save it per channel to support unassociated channels.

A consequence of doing it like that would also be that when we write epoch data into NWB, we will have separate epoch entries for DA and AD timeseries.

t-b:
Rename "AD suppression" to "Unacquired DA data" and forget that this has something todo with AD data.

t-b:
@timjarsky Ready for review. This has three big things: Better and more consistent epoch naming, limiting of epochs to acquired data together with a new unacquired epoch and user epochs with PSQ_Ramp aus guinea pig.

timjarsky:
@t-b very nice!!

timjarsky (dismissed):

t-b:
Will merge once CI passes.

t-b:
The ITC test failures were real. I've fixed them.

PR #1016, opened 2021-07-20
Automate tp in ic

### LINKED ISSUE #887: automate TP amplitude in IC mode
When TP is running during ITI, measure peak voltage deflection relative to rest (we already do this).

Let user specify peak deflection target in TP settings.

Have MIES automatically adjust TP amplitude to hit target voltage.

Once we do this, we can add an additional baseline QC check to IVSCC protocols to make sure the TP is within an amplitude range.

### LINKED ISSUE #1015: TP duration is calculated incorrectly for TP during DAQ and TTL channels
See TP_ReadTPSettingsFromGUI which always passes `TEST_PULSE_MODE` to `DAP_GetSampInt`.

The fallout is that the TP property calculation in `TP_TSAnalysis` is (slightly??) wrong.

t-b:
- [x] Add tests for new tp settings in LBN
- [x] Add percentage and interval to tests
- [x] Check calculated TP settings during DAQ via analysis function
- [x] Add tests with different baselines (Solved in https://github.com/AllenInstitute/MIES/pull/1093)
- [x] Finish Auto Amp
   - [x] Make it work with multiple headstages
   - [x] Baseline is also recalculated
   - [x] Turn off checkbox again after 3 successful TPs
- [x] Every HS has its own TP amplitude, but all use the same baseline for all HS
- [x] More tests
- [x] Commit cleanup

Auto Baseline:
- [x] sample file on FTP, 2021_08_27_095939_autoTP_development.pxp
- [x] Auto TP implies "Save every TP" on, turn back off when "Auto TP" finished or is disabled
- [x] Use two adjacent TPs for fitting, the first TP should also have been used for TP amplitude
- [x] Determine tau time constant via exp fit from two TPs
   - [x] `tau = 1/4 * cursor(A, B)` where 4 is constant for now.
   - [x] Three times in the range `1/3.75` to `1/4.25` means we are done. Not in range means baselineFrac needs adapatation.
   - [x] Count number of times 45% baselineFrac (the maximum) is hit. Turn off feature after XXX times and warn user.
   - [x] Auto Amp -> Auto TP
   - [x] Remove 0.3ms from both ends of the fit
- [x] Switch to dblexp_XOffset, see https://github.com/AllenInstitute/MIES/pull/1016#issuecomment-953220114
- [x] Make fitting routine available as function in SweepFormula, https://github.com/AllenInstitute/MIES/issues/1155

Both amplitude and baseline automation are controlled by the same checkbox and work on the same TPs.

Open issues:
- [x] After we got "Auto TP baseline adaptation failed 3 times in a row for headstage 0 to calculate a value in range." It is impossible to reenable it, as we always after one failure say "hey 3 times again" let's turn it off.
- [x] Don't do it in "TP during DAQ"
- [x] Add issue for "TP during ITI", https://github.com/AllenInstitute/MIES/issues/1149
- [x] Some more Auto TP tests
- [x] Fix GUI layout of new hardware controls
- [x] New values are applied to all headstages if "All HS" is checked
- [x] publish results via zeromq

Open questions:
- [x] Auto Amp % only for amplitude?
- [x] Auto Amp interval is used for both?

Close #887.
Close #1015.


t-b:
Ready for a first test. Only 1 headstages works at the moment.

timjarsky:
@t-b I had pushed you to fit the exponential to the baseline epoch of the TP waveform. I think we should try the pulse epoch of the TP first (if it makes the coding easier). I've changed my mind because I don't think the accuracy of the time constant from the fit will suffer very much and we also don't need a very accurate measurement.

t-b:
@timjarsky If you like you can test drive the auto TP amplitude. The automatic turning off is not yet done, although I already record if I need to adapt the TP amplitude or not. And of course baseline is also not yet done.

t-b:
> I had pushed you to fit the exponential to the baseline epoch of the TP waveform. I think we should try the pulse epoch of the TP first (if it makes the coding easier). I've changed my mind because I don't think the accuracy of the time constant from the fit will suffer very much and we also don't need a very accurate measurement.

Well I can change the approach if you like, but I already have a plan for what we talked about on monday using two consecutive TPs.

timjarsky:
@t-b I gave this a try. A few issues that are probably better suited to a screen share but I'll take a shot here.

It looks like the resistance calculation may be off by an order of magnitude (too big).

I think it makes sense to tune the baseline % before the amplitude, otherwise the baseline won't be a reliable reference to calculate the resistance from.

I also had some assertions when using auto config. Here is the first one:

```
  !!! Assertion FAILED !!!
  Message: "RTError 145 in TP_Analysis thread: a wave write;index out of range"
  Stacktrace:
TPM_BkrdTPFuncMD(...)#L307 [MIES_TestPulse_Multi.ipf]
SCOPE_UpdateOscilloscopeData(...)#L620 [MIES_Oscilloscope.ipf]
ASYNC_ThreadReadOut(...)#L256 [MIES_Async.ipf]
TP_ROAnalysis(...)#L105 [MIES_TestPulse.ipf]
  Time: 2021-09-02T21:27:19-07:00
  Locked device: [ITC1600_Dev_1]
  Current sweep: [NaN]
  DAQ: [DAQ_NOT_RUNNING]
  Testpulse: [TEST_PULSE_BG_MULTI_DEVICE]
  Experiment: 2021_09_02_212708 (Packed)
  Igor Pro version: 8.0.5.1 (37456)
  MIES version:
  Release_2.2_20210503-361-g0bf103bd
Date and time of last commit: 2021-09-02T23:21:17+02:00
Submodule status: 
160000 04508cd1b39cbf6da565fc9a04203d797daffcc0 0	Packages/IPNWB
160000 56f1da31d786a593892280c2d2b2b1190def0fb5 0	Packages/doc/doxygen-filter-ipf
160000 42276ed677273c02bc7b457324081dd463bb0d6b 0	Packages/unit-testing
```

t-b:
@timjarsky Ready for a first test. I probably makes sense to do the test together.

timjarsky:
> @timjarsky
> 
> > I'd like to add a feature request to this PR. Instead of fitting with a single exponential, fit with a double exponential. Use the tau from the exponent that accounts for the majority of the fit for baseline tuning. If the exponential that accounts for the minority of the fit exceeds 10%, print a warning to the user. Access resistance warning: multiple exponentials are contaminating the TP response profile.
> 
> I'm already implementing that. When you say "majority of the fit" do you mean larger tau or something like the curves area? Or to rephrase it. Can I just take the maximum of both tau's and complain if the smaller on exceeds 10% of the larger one?

I was referring to the amplitude values (A1 and A2; see image). However, when playing with fitting I realize now that 10% amp isn't going to work. We'll need to use a combination of tau difference and amplitude difference. For now, let's make constants for the tau diff and amp diff that will trigger the warning. I'll work to identify the correct values.

image




> EDIT: Does warning the user imply that the fit and therefore baseline adjustment failed?

No.

t-b:
This is now auto TP but without the double exponential thing. Please test. If something is not as expected please also upload the data so that I can have a look. Turning on debugging for `MIES_Testpulse.ipf` will also help me.

timjarsky:
@t-b I posted a file to your FTP with failed TP tuning and started a slack thread: https://ai-mies.slack.com/archives/D1P2AN29F/p1635828402013800

t-b:
@timjarsky Ready for the next round of testing.

timjarsky:
@t-b I can break TP tuning badly by updating `setvar_DataAcq_targetVoltage` using the arrow keys (multiple times in a row) when the TP is running (Does not require TP tuning).

The control action procedure calls DAP_TPGUISettingToWave(...) My quick guess is that this shouldn't get called while the TP is running.

Here is the assertion:

```  !!! Threadsafe assertion FAILED !!!
  Message: "Encountered pending RTE: 980, user function TS_GetNewestFromThreadQueue;Invalid Thread Group ID or index."
  Stacktrace:
TPM_BkrdTPFuncMD(...)#L283 [MIES_TestPulse_Multi.ipf]
TS_GetNewestFromThreadQueue(...)#L34 [MIES_ThreadsafeUtilities.ipf]
DAP_SetVarProc_TestPulseSett(...)#L4082 [MIES_DAEphys.ipf]
DAP_TPGUISettingToWave(...)#L5852 [MIES_DAEphys.ipf]
TP_RestartTestPulse(...)#L1226 [MIES_TestPulse.ipf]
TPM_StartTestPulseMultiDevice(...)#L111 [MIES_TestPulse_Multi.ipf]
DAP_CheckSettings(...)#L2283 [MIES_DAEphys.ipf]
HasPanelLatestVersion(...)#L5679 [MIES_MiesUtilities.ipf]
GetPanelVersion(...)#L5698 [MIES_MiesUtilities.ipf]
str2numSafe(...)#L3869 [MIES_Utilities.ipf]
AssertOnAndClearRTError(...)#L4379 [MIES_Utilities.ipf]
  Time: 2021-11-11T17:28:12-08:00
  Experiment: 2021_11_11_165055 (Packed)
  Igor Pro version: 9.0.0.10 (37840)
  Assertion FAILED with message Encountered pending RTE: 980, user function TS_GetNewestFromThreadQueue;Invalid Thread Group ID or index.
  Please lock the panel to a DAC in the Hardware tab```

timjarsky:
@t-b it looks like other auto tp controls can do the same thing

timjarsky:
@t-b Can TP tuning handle the following:

Turn on TP tuning on one HSx
(All of the following happens while TP tuning is active on one HSx)
Move slider to another headstage (HSy)
Adjust TP tuning settings
Start TP tuning on HSy

If the assertion a couple of comments up is fixed, my impression is that TP tuning can handle the above scenario. Can you verify?

t-b:
@timjarsky I fixed some tests and also raised the LBN version as that was forgotten.

t-b:
So the fix with lessening the required TP restarts is in.

timjarsky:
@t-b with autoTP selected on all HS and the slider on the second HS, tuning doesn't go to completion and is applied only to the second HS

image

Also, with tuning turned off, user input to the %baseline is overridden (not that auto TP is unchecked)

https://user-images.githubusercontent.com/5693697/142080672-c070625f-8d87-4ac6-bce8-d1095cd22532.mp4

t-b:
@timjarsky Next iteration.

timjarsky (approved):

t-b:
@timjarsky 🥇

PR #1032, opened 2021-08-17
Make dimension labels valid liberal object names

### LINKED ISSUE #1027: IP9: Make it compatible with recent SetDimLabel tightening
The next beta version of IP9 will have a fix for SetDimLabel. This fix is that it does not accept arbitrary strings anymore but "only" liberal names. Liberal names can contain neither control characters nor `:;'"`. Existing waves in pxp files can still be read as reading compatibility is kept.

This requires some adaptations in MIES as we use the labnotebook keys as dimension labels in the labnotebook. And we have dimension labels which don't follow liberal naming rules ("JSON config file: .*", "Async AD XXX: YYY", possibly more).

Unfortunately I was not aware of this (not-enforced) boundary condition of dimension labels.

t-b:
- [x] Search for invalid dim labels in some sample pxp files
- [x] Add `IsValidLiberalObjectName`, modelled after `IsValidObjectName`
- [x] Enforce valid liberal names in `ED_FindIndizesAndRedimension` via IsValidLiberalObjectName assertion
- [x] Change existing LBN keys which are now invalid to be valid again. Maybe replace ":" with "|"?. Raise LBN version.
- [x] Upgrade existing labnotebooks (values and keys) in UpgradeLabNotebook, raise LBN version
- [x] Change wavebuilder parameter names and upgrade WP/WPT waves (`AddDimLabelsToWP`/`AddDimLabelsToWPT`)
- [x] Adapt `GetAsyncSettingsKeyWave`

Close #1027.

### DIFF
--- a/Packages/MIES/MIES_Constants.ipf
+++ b/Packages/MIES/MIES_Constants.ipf
@@ -623,7 +623,7 @@ Constant HARDWARE_DAC_EXTERNAL_TRIGGER = 0x1
 /// Used to upgrade the GuiStateWave as well as the DA Ephys panel
 Constant DA_EPHYS_PANEL_VERSION           = 53
 Constant DATA_SWEEP_BROWSER_PANEL_VERSION = 30
-Constant WAVEBUILDER_PANEL_VERSION        = 10
+Constant WAVEBUILDER_PANEL_VERSION        = 11
 Constant ANALYSISBROWSER_PANEL_VERSION    =  1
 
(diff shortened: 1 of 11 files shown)


t-b:
@timjarsky Ready for review.

t-b:
@timjarsky Thanks for prodding me about tests. I've added some and found a bug as well. Ready for review again.

timjarsky (approved):

Issue #1041, opened 2021-08-26
Sweepformula: Allow to use epoch info for ranges in the sweeps command


t-b:
New function `epoch(DAC, name, [type])`. The optional `type` argument has the following allowed values:
- `range`: same as ommited, 1D array with 2 elements with start/stop positions in DAC x-units
- `name`: returns the name of the epoch
- `treelevel`: tree level (numeric)

Example commit introducing `area`: 8b9d5444 (SweepFormula: Add area operation, 2020-06-11)

t-b:
Missing `/Z`:
```
			WAVE channelNumberIndex = FindIndizes(traces, colLabel = "channelNumber", var = channels[j][%channelNumber])
				WAVE channelTypeIndex = FindIndizes(traces, colLabel = "channelType", str = StringFromList(channelType, XOP_CHANNEL_NAMES))
```

t-b:
I just chatted with @timjarsky and we realized that the name of the epoch is currently a bit ill-defined. First it is not-unique (multiple baselines) and second it can be very long due our additional tags. So we thought about adding a new `ShortName=XXX` entry to the epochs. This would be preferred (if present) over the full long name.

The naming scheme of the short names is as follows (explanation in parantheses):
```
TP_B0 (Testpulse baseline 0)
TP
TP_P (Testpulse pulse)
TP_B1
B0 (baseline 0)
ST (stimset)
E1_PT_P0_P (epoch 1, pulse train, pulse 0, pulse)
E1_PT_P0_B (epoch 1, pulse train, pulse 0, pulse)
...
```
The main goal of the short names are that these are unique and short and (somehow) user friendly.

@MichaelHuth We can chat about this on friday.

MichaelHuth:
After discussion with Thomas:

Change API to:  `epochs(array sweeps, array channels, string shortName, [type])`

MichaelHuth:
Epoch short name convention:

```
Before TP:

B0_OD // onset Delay user
B0_DD // distributed daq
B0_TD // termination delay
B0_DO // distributed daq opt
B0_TR // trailing baseline from distributed daq

TP:

TP 		// full TP
TP_P // Testpulse Pulse
TP_B0 // Testpulse baseline Front
TP_B1 // Testpulse baseline back

Stimset:

ST // Stimset
Ex // Epoch
Ex_PT // PulseTrain (not used)
Ex_PT_Px // PulseTrain Single Pulse
[...]

PR #1045, opened 2021-09-02
Sweepformula - add epochs operation / epochs shortnames

MichaelHuth:
- SweepFormula moved operation strings to constants
  - operations strings were scattered in the code, moved them to static constants
  with SF_OP_ prefix.
- Added epochs operation
- Moved all operations into own function
- Added short names for epochs
- Changed epochs operation to use short names preferentially

https://github.com/AllenInstitute/MIES/issues/1041


t-b:
@MichaelHuth Please rebase against latest main for the UTF fixes.

MichaelHuth:
Epoch short name convention:

```
Before TP:

B0_OD // onset Delay user
B0_DD // distributed daq
B0_TD // termination delay
B0_DO // distributed daq opt
B0_TR // trailing baseline from distributed daq

TP:

TP 		// full TP
TP_P // Testpulse Pulse
TP_B0 // Testpulse baseline Front
TP_B1 // Testpulse baseline back

Stimset:

ST // Stimset
Ex // Epoch
Ex_PT // PulseTrain (not used)
Ex_PT_Px // PulseTrain Single Pulse
Ex_PT_Px_BT // PulseTrain Single Pulse baseline before left or rightmost pulse
Ex_PT_Px_P // PulseTrain Single Pulse Pulse Active
Ex_PT_Px_B // PulseTrain Single Pulse Pulse Inactive
Ex_PT_BT // PulseTrain baseline trail if zero pulses
ST_B // for multiple sweeps where each sweep has a different length in stimset they have the same length
		 // Stimsets that are extended have a baseline added

```

timjarsky:
@MichaelHuth or @t-b could you share an example of the epoch formula syntax. I was playing around with `epoch([0,1],DA0,TP_P)
` but I must not have it right. Thank you!

MichaelHuth:
@timjarsky 
```
data(
epochs(0, channels(AD1), "ST"),
channels(AD1),
0
)
```

Works to get the Stimset as graph. If you use `sweeps()` it might return a 2xN wave that is not supported for display yet.

Without specifying type explicitly you get the range. If you are unsure how many results are returned try treelevel as type, which returns a 1D wave that can be displayed.

Currently if the type is range and there is only a single result then instead of a 2x1 wave a 1D wave with two elements is returned (i.e. a single range).

t-b:
@MichaelHuth I've rebased the PR to fix some conflicts.

t-b:
Review:

631cb79a (SweepFormula moved operation strings to constants, 2021-09-02)

Good.

a27e2690 (Patch for SweepFormula parser to allow "=" character in strings, 2021-09-03)

Jeep.

16519249 (SweepFormula add epochs command, 2021-09-03)

- In this piece of code

```
+                               if(!IsEmpty(str))
+                                       WAVE/T epochInfo = EP_EpochStrToWave(str)
```

- I don't understand why you are checking for empty. Why not just SF_ASSERT out?

```
FindValue ...
if(V_Value >= 0)
code using V_row
```

Why the switching from V_Value to V_row? Why not always use V_row?

- You are asserting out when there are no matches, but in the labnotebook function we do
```
	if(!WaveExists(out))
		DebugPrint("labnotebook entry not found.")
		Make/FREE/N=1 out = {NaN}
		return out
	endif
```

Any reason?

2ecf7cd4 (fix SweepFormula SF_GetSweepForFormula was missing WAVE/Z that caused an RTE, 2021-09-03)

Jeep.

80d6b3d9 (Introduced seperator constants for epoch wave to list for LBN and back, 2021-09-06)
3d697dc9 (SweepFormula: moved operation code in functions for each op, 2021-09-06)
edaf4e02 (Add short names for Epochs stored in epoch name, 2021-09-06)
40220f87 (Sweepformula: Epoch operation uses ShortNames preferentially, 2021-09-06)
9c1767b3 (SweepFormula rewieved num2str usage, 2021-09-07)

All good.

7ba0ca75 (HardwareTests: Added generic epoch consistency check to test end hook, 2021-09-07)

CheckEpochs: 
-You only need to check DACs, there is no epoch data for ADCs.
- All active DACs run in DAQ mode must have epoch info.
- Same question about `if(!IsEmpty(str))` as above.

910b8838 (Add test for epochs operation in sweepformula, 2021-09-08)

- Can you attach the epoch info to the DA channel to make it obvious that AD does not have one?
- The fail tests should exercise a couple of more conditons (invalid sweep, invalid type, invalid channel).
-  
bde95f66 (Add description of epoch operation in SweepFormula help notebook, 2021-09-08)

Good.

d40498ec (Added documentation on epoch short names, 2021-09-08)
371680c8 (Add documentation for SweepFormula function epochs(), 2021-09-08)

Very nice.

t-b (changes requested):
See comment.

timjarsky (approved):

MichaelHuth:
@t-b Regarding the test: 
I think the reason for the diffference is that there are 16 AD channels, but only 8 DA channels.
With the labnotebook setup, there are q + p^r and the connection order 7,5,3,1,0 the following channels setup:
8, 6, -, 4, -, 2, -, 0, -. Now when looking for DA channels, we can not find 8 as only channelnumber from 0 to 7 are found.
Thus, it returns only 0, 2, 4, 6 -> 4 active channels for DA.

t-b:
@MichaelHuth Makes sense. Hanks for digging.

t-b (approved):

PR #1064, opened 2021-09-13
Change isEmpty to not use string parameter by reference

MichaelHuth:
After a discussion with Thomas about the disadvantages of isEmpty when using strings from structures or text wave elements we decided to give this change a try.

Since Igor generates a RTE if a null string is used as parameter as a non-ref parameter isEmpty will now generate an RTE if it is used to check against a null string. (Where IsNull should be used). Despite the RTE it will still work correctly for the case of a null string parameter though.

CI will hopefully show locations in the code where such lingering RTEs are generated.

### DIFF
--- a/Packages/MIES/MIES_AnalysisFunctionHelpers.ipf
+++ b/Packages/MIES/MIES_AnalysisFunctionHelpers.ipf
@@ -574,7 +574,9 @@ Function/S AFH_GetAnalysisParamType(name, params, [typeCheck, expectedType])
 	string type = ""
 	variable pos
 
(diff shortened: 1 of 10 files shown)


t-b:
@MichaelHuth Test output: [JU_2021_09_13_205331_2021-09-13_21-30-19.zip](https://github.com/AllenInstitute/MIES/files/7160003/JU_2021_09_13_205331_2021-09-13_21-30-19.zip)

t-b:
Rebased.

t-b (approved):

Issue #1063, opened 2021-09-13
EP_AddEpochsFromStimSetNote fails on TestPulse stimsets, RTE is silently cleared


MichaelHuth:
With TP during DAQ there exist stimsets containing testpulses.

EP_AddEpochsFromStimSetNote is called for these stimsets as well, but the stimset has no wave note.

This is not checked properly, whereas in MIES_Epochs.ipf a RTE is generated (epochCount = NaN) and the following code adds wrong epoch information to the TestPulse stimset. (Stimset Baseline / ST_B)
https://github.com/AllenInstitute/MIES/blob/c5830c7b870ea6d371da9d7a4e06a1cd892edc81/Packages/MIES/MIES_Epochs.ipf#L182

This RTE is silently cleared later by DC_PrepareLBNEntries where str2numSafe is used.
https://github.com/AllenInstitute/MIES/blob/c5830c7b870ea6d371da9d7a4e06a1cd892edc81/Packages/MIES/MIES_DataConfigurator.ipf#L975

How should we handle this?

- Should EP_AddEpochsFromStimSetNote ignore stimsets without wavenote?
- Should we detect if the stimset is a test pulse and not call EP_AddEpochsFromStimSetNote then / Assert out on empty wave note in the function then?
- Are the wrong epochs a problem with already present data?

- Silently clearing a RTE should not happen. In the routines that clear RTEs on purpose ASSERTs should be added before that catch pending RTEs.

t-b:
RTE clearing happens also in https://github.com/AllenInstitute/MIES/blob/c5830c7b870ea6d371da9d7a4e06a1cd892edc81/Packages/MIES/MIES_DataConfigurator.ipf#L799. I don't remember why we did that,

> Should EP_AddEpochsFromStimSetNote ignore stimsets without wavenote?

Yes it should ignore stimsets without wavenote. "Third party" stimsets don't need have a special wave note as well. A third party stimset is one which does not have all three parameter waves.

So all simsets for which `WB_StimsetIsFromThirdParty(stimset) || !cmpstr(stimset, STIMSET_TP_WHILE_DAQ)` is true will not have epoch info from the stimset note. And we should, for now, also not have epoch info for `TP during DAQ channels at all, as that needs to be figured out properly on its own, see https://github.com/AllenInstitute/MIES/issues/958.
 
> Are the wrong epochs a problem with already present data?

[...]

PR #1065, opened 2021-09-14
fix EP_AddEpochsFromStimSetNote for empty stimset wave notes

MichaelHuth:
EP_AddEpochsFromStimSetNote was also called for stimsets with empty wave notes
such as TP during DAQ stimsets. This resulted in an ST_B (stimset baseline)
epoch added for this stimset, that is wrong.

Fix: - Added a check in callers of EP_AddEpochsFromStimSetNote to call only if
       the stimset is of a type that should contain epoch data.
	 - Added better checks in EP_AddEpochsFromStimSetNote
	 - increased sweep_epoch_version.

close https://github.com/AllenInstitute/MIES/issues/1063

### DIFF
--- a/Packages/MIES/MIES_Constants.ipf
+++ b/Packages/MIES/MIES_Constants.ipf
@@ -639,7 +639,7 @@ Constant LABNOTEBOOK_VERSION = 49
 Constant STIMSET_NOTE_VERSION = 7
 
 /// Version of the epoch information for DA data
-Constant SWEEP_EPOCH_VERSION = 1
+Constant SWEEP_EPOCH_VERSION = 2
 
 /// @name The channel numbers for the different ITC devices used for accesssing
 ///       the TTLs
--- a/Packages/MIES/MIES_Epochs.ipf
+++ b/Packages/MIES/MIES_Epochs.ipf
@@ -53,33 +53,38 @@ Function EP_CollectEpochInfo(string panelTitle, STRUCT DataConfigurationResult &
 
 	stopCollectionPoint = ROVar(GetStopCollectionPoint(panelTitle))
 
-	Duplicate/FREE s.insertStart, epochIndexer
-
-	// epoch for onsetDelayAuto is assumed to be a globalTPInsert which is added as epoch below
-	if(s.onsetDelayUser)
-		epochBegin = s.onsetDelayAuto * s.samplingInterval
-		epochEnd = epochBegin + s.onsetDelayUser * s.samplingInterval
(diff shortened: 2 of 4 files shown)


MichaelHuth:
Pending RTE handling will be in a separate PR.

t-b:
Review:
- Missing tests
- As outlined in the issue I would like to not see epoch info at all for TP during DAQ channels or third party stimsets

t-b (changes requested):
See comment.

t-b:
Review:

d1f024f9 (fix EP_AddEpochsFromStimSetNote for empty stimset wave notes, 2021-09-14)

```
+       if(IsEmpty(stimNote))
+               return NaN
+       endif
+
````

Does that still make sense? From my understanding this can not happen anymore. If this is the case an assertion would be better. 

095ba1b5 (Test: Epochs added check for analysisFunction string in AcquireData, 2021-09-16)

Nice find!

7055dfbd (WaveBuilder: Added utility function WB_MakeStimsetThirdParty, 2021-09-17)

If you use WB_CreateAndGetStimSet you have a permanent stimset wave in the right place already.

```
	AddEntryIntoWaveNoteAsList(stimset, "Version", var=STIMSET_NOTE_VERSION, appendCR = 1)
```

Thirdparty stimsets don't require a wave note so this should not be needed.

Please use the following pattern to derive the channelType.

```
	type = GetStimSetType(setName)

	if(type == CHANNEL_TYPE_UNKNOWN)
		return $""
	endif
```

c20c01a9 (Tests: Added Epoch tests with TP during DAQ / third party stimset, 2021-09-16)

Looks good. Just a minor nit:

Please prefer CHECK_EMPTY_STR, see https://docs.byte-physics.de/igor-unit-testing-framework/assertions.html#_CPPv415CHECK_EMPTY_STRP6string, instead of comparison with an empty string.

**Comment by t-b** [comment]
Review:

Why do you still have 

```
+
+       WAVE/Z/SDFR=dfr oldStimSet = $setName
+       if(WaveExists(oldStimSet))
+               if(WaveRefsEqual(oldStimSet, stimSet))
+                       return NaN
+               endif
+               KillOrMoveToTrash(wv=oldStimSet)
+       endif
+       MoveWave stimSet, dfr:$setName
```

in WB_MakeStimsetThirdParty? I think that WB_CreateAndGetStimSet creates the wave already in the correct place.

t-b (approved):

timjarsky (approved):

timjarsky:
@t-b tested with a custom wave and tp during DAQ

PR #1066, opened 2021-09-14
fix for unintended clearing of pending RTEs

MichaelHuth:
In some cases we run code that may result in an RTE that we want to catch.
Out previous approach was that we enclose this part in a
try / catch / endtry block, where a critical RTE is caught through
an AbortOnRTE.
To prevent catching pending RTEs instead of the specific targeted RTE
we cleared any RTE condition with ClearRTError() directly after the try.

This poses a problem: Any pending RTE is silently cleared by the code which
generally hid all pending RTEs in the code that appeared before one of the
try clauses where executed. While for non-threadsafe code this still might
be found by executing with Debug-On-Error, with threadsafe code such conditions
got invisible.

The fix is to check the RTE code before clearing it as preparation for
critical code. So we can assert outt, if there is a pending RTE.

### DIFF
--- a/Packages/MIES/MIES_AmplifierInteraction.ipf
+++ b/Packages/MIES/MIES_AmplifierInteraction.ipf
@@ -1026,8 +1026,8 @@ Function AI_SelectMultiClamp(panelTitle, headStage)
 		return AMPLIFIER_CONNECTION_INVAL_SER
 	endif
 
-	ClearRTError()
-	MCC_SelectMultiClamp700B(mccSerial, channel); err = GetRTError(1)
+	AssertOnAndClearRTError()
+	MCC_SelectMultiClamp700B(mccSerial, channel); err = GetRTError(1) // see developer docu section Preventing Debugger Popup
 
 	if(err)
 		return AMPLIFIER_CONNECTION_MCC_FAILED
--- a/Packages/MIES/MIES_AmplifierInteraction.ipf
+++ b/Packages/MIES/MIES_AmplifierInteraction.ipf
@@ -375,6 +375,7 @@ Function AI_UpdateAmpModel(panelTitle, ctrl, headStage, [value, sendToAll, check
 				AmpStorageWave[%$rowLabel][0][i] = value
 				AI_UpdateAmpView(panelTitle, i, ctrl=ctrlToCall)
 				// the pipette offset for the opposite mode has also changed, fetch that too
+				AssertOnAndClearRTError()
 				try
 					oldTab = GetTabID(panelTitle, "ADC")
 					if(oldTab != 0)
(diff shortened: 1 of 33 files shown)


t-b:
Review:
- Why do you always repeat yourself with
```
	err = ClearRTError()
	ASSERT(!err, "Encountered pending RTE " + num2istr(err))
```

and not have a function like `AssertOnRTE()`?
- This change for try/catch is reasonable. But the invocations where we have `$SomeCommand; err = GetRTError(1)` are different. I would think that these should be left as is, but can be persuaded otherwise. I therefore think that the try/catch changes should be in its own commit.

t-b (changes requested):
See comment.

MichaelHuth:
I think that with
`$SomeCommand; err = GetRTError(1)`
we have exactly the same problem as with the ClearRTError() before we enter a try construct. At this point a RTE condition is cleared possibly overwriting a pending RTE condition, which is then lost.
I consider locations in the code where we clear pending RTEs silently dangerous as code may create incosistent data due to some error and we do not even notice it.

t-b:
@MichaelHuth Okay, bought it. Still I would like to see that in a separate commit.

MichaelHuth:
I have tested that now locally and the tests run through here.
(btw HardwareTests crash in between if the DataBrowser Window is open in IP8)

The two Epoch commits were picked from PR https://github.com/AllenInstitute/MIES/pull/1065

t-b:
Review:

0a409c8f (Introduction of function AssertOnAndClearRTError(), 2021-09-16)

Nice!

e38699e9 (Revisited all occurences of ClearRTError() after a critical function., 2021-09-16)

This is changing behaviour:

```
-threadsafe Function str2numSafe(str)
-       string str
+threadsafe Function str2numSafe(string str)

-       variable err, var
+       variable var

-       var = str2num(str); err = GetRTError(1)
+       AssertOnAndClearRTError()
+       var = str2num(str); ClearRTError()
```

GetRTError() clears any RTE *and* does prevent the debugger coming up, where as ClearRTError() does only the former. I would think that we need to keep using GetRTError(1),

6bb7950b (Revisited all occurences of ClearRTError prior try constructs, 2021-09-16)

Change looks good, but I think you missed some. I've found a couple of more with `git grep -C 1 -n "\btry$"` where we have an uncovered try. I would still keep the try calls in the ASSERT functions uncovered for performance reasons.

0f531bd9 (Fix missing clear of RTE in catch part of construct in DAP_WindowHook, 2021-09-16)

Jeep.

a9c32467 (Test: Epochs added check for analysisFunction string in AcquireData, 2021-09-16)

Good.

070f79c6 (fix EP_AddEpochsFromStimSetNote for empty stimset wave notes, 2021-09-14)

Good here, but we should merge #1065 first and then remove it on rebase.

5348c544 (fix for PatchSeqChirp tests failing, 2021-09-16)

Jeep, nice find.

PS: I've opened an issue for  the JSON XOP and json_functions.ipf as well.

t-b:
@MichaelHuth 

documentation error:

```
sphinx-build says:
/home/ci/Packages/doc/developers.rst:445: WARNING: Unexpected indentation.
/home/ci/Packages/doc/developers.rst:446: WARNING: Block quote ends without a blank line; unexpected unindent.
/home/ci/Packages/doc/developers.rst:449: WARNING: Definition list ends without a blank line; unexpected unindent.
```

t-b:
Thx.

MichaelHuth:
I added some more docu on using try / catch / endtry in the developers.rst.

t-b:
@MichaelHuth I've rebased against main.

t-b (approved):
LGTM.

Issue #1054, opened 2021-09-09
SweepFormula: Comment sign `#` does not work reliable


t-b:
Failing:
```
# abcd vs 1
# test
abcd vs 2
```

We have specified in the documentation:

> A comment character # causes the skip state which does not write to the buffer until a line ending character is encountered.

MichaelHuth:
This works as documented:

The first " vs " splits the input code into the formula for y and x.
Formula for y: # abcd
Formula for x: 1\r# test\rabcd vs 2

When the rule for the comment in formulas is applied the result is:
null vs "1abcd vs 2"

PR #1067, opened 2021-09-15
Sweepformula did not process comments with # correctly

MichaelHuth:
The code that failed:
```
# abcd vs 1
# test
abcd vs 2

```

The issue was that the "vs" token for plotting was processed before the
comments were processed in the state based parsing. Since the "vs" token is
parsed with a regex the parsing can not be simply swapped.

Fix: Before the formula is parsed it is preprocessed where based on a per line
basis the comments after # are removed.
The same step is used to unify the line endings to CR by replacing LF with CR.
(CR LF -> CR CR adds an empty line, but the state based parser ignores empty
 lines)

The comment handling in the state based parser was removed.

As testing calls the Formula Parser function directly some additional code was
introduced to execute the preprocessing then there.

close https://github.com/AllenInstitute/MIES/issues/1054

### DIFF
--- a/Packages/MIES/MIES_SweepFormula.ipf
+++ b/Packages/MIES/MIES_SweepFormula.ipf
@@ -199,7 +199,7 @@ End
 /// @param formula  string formula
 /// @param indentLevel [internal use only] recursive call level, used for debug output
 /// @returns a JSONid representation
-Function SF_FormulaParser(formula, [indentLevel])
+static Function SF_FormulaParser(formula, [indentLevel])
 	String formula
 	variable indentLevel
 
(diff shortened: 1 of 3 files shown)


timjarsky (approved):

t-b:
Questions:
- Who does

```
data(
#
[0, 500],
channels(),
sweeps()
```

now work without the formula plotter?

- The usage of `#ifdef AUTOMATED_TESTING` smells badly. In that way we have different behaviour in tests and in real usage.
- Missing tests

I would think this needs to be solved in the parser and not in the formula plotter.

MichaelHuth:
This has to be discussed as the original issue is actually invalid in the sense that it is not a bug. It behaves exactly as documented. So this PR is a feature change of SweepFormula.

t-b:
CI failure looks unrelated, I've restarted it.

Review:
 Trés bien!

t-b (approved):

t-b:
@timjarsky Please review and merge if good.

timjarsky (approved):

PR #1078, opened 2021-09-28
GetLastSettingChannel: Make it work with independent headstage data again

t-b:
`labnotebook("TP Amplitude VC" ,channels(ad),sweeps(), TEST_PULSE_MODE)` works now as expected.

### DIFF
--- a/Packages/MIES/MIES_ExperimentDocumentation.ipf
+++ b/Packages/MIES/MIES_ExperimentDocumentation.ipf
@@ -12,7 +12,7 @@
 /// @brief Add numerical/textual entries to the labnotebook
 ///
 /// @see ED_createTextNotes, ED_createWaveNote
-Function ED_AddEntriesToLabnotebook(WAVE vals, WAVE keys, variable sweepNo, string panelTitle, variable entrySourceType)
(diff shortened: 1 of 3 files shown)


timjarsky:
> `labnotebook("TP Amplitude VC" ,channels(ad),sweeps(), TEST_PULSE_MODE)` works now as expected.

@t-b should the above return an array? It looks like a single value is returned?

I was thinking that ```labnotebook("TP Amplitude VC" ,channels(ad),sweeps(), TEST_PULSE_MODE) vs sweeps()``` would generate a lab notebook browser like plot.

t-b:
Yes it returns an array when it found something.

```
labnotebook("TP Amplitude VC" ,channels(ad1), sweeps(), DATA_ACQUISITION_MODE) vs sweeps()
```

works here. My formula requires that TP during ITI was run for each sweep and RA was active.
Your formula with `TEST_PULSE_MODE` requires that for each acquired sweep we ran TP but not as part of DAQ (as those entries have type DATA_ACQUISITION_MODE) but as part of a normal testpulse.

If you don't care about the entry source type you can also pass `UNKNOWN_MODE` which takes whatever it can find. Does that make sense?

timjarsky (approved):

PR #1084, opened 2021-09-29
Fix wrong sampling interval

t-b:


timjarsky:
@t-b I think the only way to check that it was applied is in a mid or post sweep event. As it is right now, I can change the multiplier during acquisition :imp:.

I'm okay with merging this now and doing another PR later.

t-b:
Rebased to fix conflicts.

> @t-b I think the only way to check that it was applied is in a mid or post sweep event. As it is right now, I can change the multiplier during acquisition imp.
> I'm okay with merging this now and doing another PR later.

In #1086 (which will target main) I'm changing PGC_SetAndActivateControl to bail out if it could not set a control. I think this should help catching bugs earlier in the future.

But what you are describing is an accidental/malicous way of interfering with DAQ. If we want to prevent that, we would need a much more concentrated effort I think. I've added it to your chat list.

timjarsky:
> But what you are describing is an accidental/malicous way of interfering with DAQ. If we want to prevent that, we would need a much more concentrated effort I think. I've added it to your chat list.

@t-b I propose we add an additional baseline qc check on top of RMS and targetV checks. This should be a separate PR.

timjarsky (approved):

PR #1086, opened 2021-09-29
PGC_SetAndActivateControl: Make it less error prone to use

### LINKED ISSUE #1083: PGC_SetAndActivateControl: Should enforce that the set string is a valid entry from the available ones

t-b:
Close #1083.


timjarsky:
@t-b 
- [x] pgc warns appropriately when attempting to manipulate a disabled control. 


I do get a strange warning when attempting to manipulate the sweep browser button; it says I'm using an old version of the panel (the button works normally when using it with the mouse)

image

t-b:
> I do get a strange warning when attempting to manipulate the sweep browser button; it says I'm using an old 
> version of the panel (the button works normally when using it with the mouse)

The reason for that strange error message is that you are not using the full window path. The previous button is in a subwindow, and the button procedure expects the full path. It uses the full path, e.g. `DB_ITC18USB_Dev_0#SweepControl`, to get the parent window/graph and check the panel version there. As you are only passing `SweepControl` into it, there is no version info in that panel and therefore the bogus message.

Fixing that is not so difficult, but I think that the performance hit for this is too large to be really practical. I've opened https://github.com/AllenInstitute/MIES/issues/1139 to document my findings.

timjarsky (approved):

PR #1091, opened 2021-10-02
Append zeromq logfile to nwb

### LINKED ISSUE #1080: Append zeromq log to nwb file
NWB_AppendIgorHistoryAndLogFile does already append the MIES log.

t-b:
Close #1080.


timjarsky:
@t-b does this append the entire logfile or the log entries that were made during the experiment? I think we want the latter, yes?

t-b:
@timjarsky This appends always the entire log file. I'll have a look how difficult it is to figure out which entries belong to this experiment. An easy approach would be to just save the entries from the same day.

t-b:
Done.

Issue #1094, opened 2021-10-08
Stimset API: Inform user when API is called during MID SWEEP that the change is applied on next sweep. 


t-b:

t-b:
Document in stimset API, don't print anything during DAQ.

PR #1105, opened 2021-10-15
Tests: Check TP Storage wave

t-b:
These are the tests requested in #1094.

From my understanding the calculated resistance values should be independent of the amplitudes or? I can verify that behaviour with one HS, but if I have one VC and one IC heastage the resistance values change with amplitude. Is that just a test setup artefact?


t-b:
- [ ] Fix hardware setup of CI machines

PR #1130, opened 2021-11-02
IsConstant: Allow to not ignore NaNs and fix edge cases

### LINKED ISSUE #1124: Fix IsConstant for val == NaN
It does not do the right thing,

t-b:
The issue triggering this [1] claimed that IsConstant does not work with
NaN in the input.

But actually the current behaviour was just to ignore NaNs in the input,
which also all callers relied on.

We now have the option to not ignore NaNs in the input.

We also fixes edge cases for empty input waves (return NaN) and the case
were all input is NaN and we ignore NaN. The latter now gives NaN as this is
equivalent to an empty input wave.

[1]: https://github.com/AllenInstitute/MIES/issues/1124

Close #1124.


t-b:
I'd say this is good to merge once CI passes.

Issue #352, opened 2019-11-04
SweepFormula: Preserve channel information


t-b:
Can we start color-coding the SF plot traces by headstage (in this PR)?

_Originally posted by @timjarsky in https://github.com/AllenInstitute/MIES/pull/341#issuecomment-549095073_

ukos-git:
### current behavior

Currently, channel information is only available for inputs to the `data()` function like  `channels(AD0,AD9)`.

This currently creates a data wave of the following layout:

```
ROWS: data
COLS: sweep number grouping
LAYERS: channels
```

an input of `channels(AD0,AD9)` creates 2 items in `LAYERS`. an input of `channels(AD)` creates 1 item in LAYERS. On the plotting level we therefore do not have the channel information available for inputs like `channels(AD)`.

### preserve channel information

A clean solution to this is to create channel information for an input like `channels(AD)` to evaluate to the available channels. This can only be done in the  `SF_GetSweepForFormula(graph, range[0], range[1], channels[i][0], channels[i][1], sweeps)` which is the wave getter for the data() function. This function needs to handle the complete channels wave and build up a 3d-wave.

If AD consists of `AD0,AD9`, inputs like `channels(AD, AD0)` will then evaluate to a non-unique channel array `AD0,AD9,AD0` alignment in the LAYERS dimension.

### adopt statistical functions

Currently inputs that have more than one item in the LAYERS dimension (3d-waves) do not work with `min,max,mean` etc. as these functions all remove one dimension from the data structure. They act on the ROWS dimension and the dimension is removed afterwards.

I have added asserts for acting on LAYERED waves as it was not clear what to calculate from these waves. I now get what the minimum of a 3dimensional wave should produce and we can implement it. This needs to be secured through unit tests with a proper MWE for a 3-dimensional wave.

### color

Finally, if the last dimension is preserved by all functions and contains the channel information, color information can be added upon plotting.

### trace naming

I would also recommend adding dimension labels for the columns and layers so that traces can get named  with sweep number and channel type/number.

## TLDR;

Coloring the traces requires to preserve channel information in the last dimension.

timjarsky:
@t-b @ukos-git will the SF wave have a fixed number of layers with specific layer/channel associations?

t-b:
No only layers with data fill be added.

ukos-git:
[...]

PR #1131, opened 2021-11-02
1131/feature sweep formula preserve channel information

MichaelHuth:
close https://github.com/AllenInstitute/MIES/issues/352

Note that the PR only covers the coloring of the traces mentioned initially in the issue.

### DIFF
--- a/Packages/MIES/MIES_SweepFormula.ipf
+++ b/Packages/MIES/MIES_SweepFormula.ipf
@@ -858,8 +858,8 @@ static Function SF_FormulaPlotter(string graph, string formula, [DFREF dfr, vari
 		RemoveTracesFromGraph(win)
 		ModifyGraph/W=$win swapXY = 0
 
-		if(IsTextWave(wvY) && WaveExists(wvX))
-			SF_Assert(WaveExists(wvX), "Cannot plot a single text wave")
+		if(IsTextWave(wvY))
+			SF_Assert(WaveExists(wvX), "A single text wave requires X-values to be plotted.")
 			ModifyGraph/W=$win swapXY = 1
 			WAVE dummy = wvY
 			WAVE wvY = wvX
--- a/Packages/MIES/MIES_SweepFormula.ipf
+++ b/Packages/MIES/MIES_SweepFormula.ipf
@@ -83,6 +83,7 @@ static StrConstant SF_OP_LABNOTEBOOK = "labnotebook"
 static StrConstant SF_OP_WAVE = "wave"
 static StrConstant SF_OP_FINDLEVEL = "findlevel"
 static StrConstant SF_OP_EPOCHS = "epochs"
+static StrConstant SF_OP_TRANSPOSE = "transpose"
 
 static StrConstant SF_OP_EPOCHS_TYPE_RANGE = "range"
 static StrConstant SF_OP_EPOCHS_TYPE_NAME = "name"
@@ -749,6 +750,9 @@ Function/WAVE SF_FormulaExecutor(jsonID, [jsonPath, graph])
 			break
 		case SF_OP_APFREQUENCY:
 			WAVE out = SF_OperationApFrequency(jsonId, jsonPath, graph)
+			break
+		case SF_OP_TRANSPOSE:
+			WAVE out = SF_OperationTranspose(jsonId, jsonPath, graph)
 			break
 		default:
 			SF_ASSERT(0, "Undefined Operation", jsonId=jsonId)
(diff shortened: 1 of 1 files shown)


t-b:
@MichaelHuth Test failure: 

```
  | UTF_SweepFormula#MIES_channel                                 UTF_SweepFormula#MIES_channel in UTF_SweepFormula.ipf (9)                                 History
-- | --
```

with

```
FAIL: Assuming equality using mode WAVE_DATA for waves _free_ and _free_; detailed: Waves difference: Wave1: 0x0819eda910 Wave2: 0x08240bdf70 Dimensions  |Labels         |Value   | ------------|---------------|--------| [0][0][0][0]|               |2.000000|             |;channelType;;;|nan     | : is false. Assertion "REQUIRE_EQUAL_WAVES(input, output, mode = WAVE_DATA)" failed in line 627, procedure "UTF_SweepFormula.ipf"
```

t-b:
Please also rebase.

MichaelHuth:
I removed the change in the channels operation as the underlying issue is bigger.

t-b:
@timjarsky I did not yet try it out, but if you like to peek, feel free.

timjarsky:
@t-b well, this is working really nicely!

will it be possible to carry the channel information up another level? e.g., have the arrays that come from analysis of a timeseries share the color code? In the example below, the lower two plots would have their dots colored according to the headstage instead of red
image

t-b:
SF COde

```
data(
cursors(A,B),
channels(AD1),
sweeps()
)

and 

apfrequency(data(
cursors(A,B),
channels(AD1),
sweeps())
, 1, 15)
```

@MichaelHuth Is looking into keeping the colors for apfrequency as well. If that works without mind-bending code changes we can think about using that in more SF functions.

MichaelHuth:
@timjarsky I had to change the output of apFrequency slightly. If the change is ok for you, I add docu and finalize it.
To get apFrequency to transfer the sweep/channel information it has to return the results in the same form as the data operation.
Before it returned a 1D wave that resulted in a single trace, which allowed only a single color.

Now it returns a wave in the same format as data, that is (1, sweeps, channels). Effectively this results in one trace per (sweep * channel) and allows to color each result independently. As each result is now an own trace without X-points it plots them at X = 0. I thought this might be a bit inconvenient and it would be better to have something like e.g.
apfrequency(...)
vs
sweeps()

The issue with this is that sweeps() returns a 1D wave, where the proper combination with apfrequency would be a (1, sweepCnt) wave. To solve that I added a transpose operation.

To relate to the upper example the new way would be:
```
data(
cursors(A,B),
channels(AD1),
sweeps()
)

and 

apfrequency(data(
cursors(A,B),
channels(AD1),
sweeps())
, 1, 15)
vs
transpose(sweeps())
```

if you just want to have the classical behavior with counting from zero on the x-axis then write (e.g. 4 sweeps):
`transpose([[0, 1, 2, 3]])`

Note the double array brackets, that are required as transpose([0, 1, 2, 3]) would be the same as transpose(0, 1, 2, 3) due to the implicit array extension of the latter. transpose takes only a single argument.

I am not sure if some operation that returns the size of a 1D array would be convenient to be able to write something like
`transpose([0...size(sweeps())])` for the X-values?

timjarsky:
> @timjarsky I had to change the output of apFrequency slightly. If the change is ok for you, I add docu and finalize it. To get apFrequency to transfer the sweep/channel information it has to return the results in the same form as the data operation. Before it returned a 1D wave that resulted in a single trace, which allowed only a single color.
> 
> Now it returns a wave in the same format as data, that is (1, sweeps, channels). Effectively this results in one trace per (sweep * channel) and allows to color each result independently. As each result is now an own trace without X-points it plots them at X = 0. I thought this might be a bit inconvenient and it would be better to have something like e.g. apfrequency(...) vs sweeps()
> 
> The issue with this is that sweeps() returns a 1D wave, where the proper combination with apfrequency would be a (1, sweepCnt) wave. To solve that I added a transpose operation.
> 
> To relate to the upper example the new way would be:
> 
> ```
> data(
> cursors(A,B),
> channels(AD1),
> sweeps()
> )
> 
> and 
> 
> apfrequency(data(
> cursors(A,B),
> channels(AD1),
> sweeps())
> , 1, 15)
> vs
> transpose(sweeps())
> ```
> 
> if you just want to have the classical behavior with counting from zero on the x-axis then write (e.g. 4 sweeps): `transpose([[0, 1, 2, 3]])`
> 
> Note the double array brackets, that are required as transpose([0, 1, 2, 3]) would be the same as transpose(0, 1, 2, 3) due to the implicit array extension of the latter. transpose takes only a single argument.
> 
> I am not sure if some operation that returns the size of a 1D array would be convenient to be able to write something like `transpose([0...size(sweeps())])` for the X-values?

@MichaelHuth @t-b apfrequency is used to analyze the data from a single headstage across multiple sweeps (not to analyze multiple headstages for the same sweep).  The user may wish to fit the output with a line. The ability to fit (or do other analysis on related data; data from the same headstage) is lost when each point is its own array. My preference is to revert the output of apfrequency but we can discuss.

t-b:
@timjarsky Let's discuss later on. Here are my notes chatting with @MichaelHuth.

```
# preserving channel info works easily
# and can be extended to other operations
# if and only if we change the returned data format to
# the same as data()
# 
# Drawback is that the implicit x-axis
# does not work anymore as we have a (1, X, Y) wave
# instead of a (X, Y).
# Using something like ... vs transpose(sweeps()) works
# but is a bit tedious to write.
# We propose to add support that every operation can suggest an
# x-axis in case there is none.
# This can be solved via a wave note a la 
# { "X_VALUES" : [111, 114, 116]}
# which would be used if the row dimension only has one point and
# no explicit x-axis is given.
# 
# For operations like max/min/avg/... we keep the input columns/layers for the output format.
# This would allows us to preserve channel info also for those.
#
# This also helps to streamline the internal processing as more operations
# use the same data format as data().
```

t-b:
@timjarsky This is our proposal:

```
Proposal:
# - apfrequency: return standard data format as data(...) -> (data, sweeps, channels) -> (count, sweeps, 1) for method 2
# - add feature to specify y data dimension for plotter as json in wave note and use it in apfrequency for COLS
# - plotter: evaluate wave note and preprocess to move specified y data dimension to rows
#   this allows to keep the current plotting algorithm (otherwise algorithm for COLS, LAYERS would be required)
#
# - try to be clever for vs x plotting, and treat (N, 1) and (1, N) the same for x data
#   (N, 1, 1)
# 
#   vs (N, 1) oder (1, N, 1)
#
#   this avoids users having to use transpose()
# - preserve channel information via colors, take color from first headstage with multiple sweeps if no sweep information in COLS/LAYERS
# - add some more tests for plotting
#
# Later (new PR):
# 
# Move min/max/mean/etc. to standard data format
```

MichaelHuth:
Proposal

Problem:
We want to display sweep data with specific ranges and different x scales.
Currently the SweepFormula approach is to carry the data through the json objects
to each operation in a single multi dimensional array. For data all sweeps are saved in the same 3d array.
This introduces a series of limitations:
- if we have data with a different number of points we have to introduce placeholder values as well as information how the actual length is
- if we have data with different scaling per row we need to save that information somewhere

Solution concept:
Instead of carrying the data through the json objects we store wave paths.
The data that a operation generates is put as a wave with a unique name to a working data folder. As the wave path is a string we can extend that and save
in the strings a dumped json instead.
(I am not sure if json objects would break the executor)
Since the string represents a json we can save additional data to the wave path, like
channel number, sweep number.
Alternative: Save it in wave note.

We can add a string prefix or first element in the array as marker for wave paths with meta data.

Advantage:
Since each part of the data is in a separate wave we have intrinsically all the
X-Scaling, Wave Unit, Wave Note for the part. A part of the data is e.g. a range from a sweep/channel combination.
We can in principle introduce types.

We can carry meta information through every operation, e.g. max(data(...))
Max would create then waves with a single element for each wave from data.

Disadvantage:
Current operations must be changed to support the additional variant with data provided through wave paths. The plotter needs to be adapted.

t-b:
Will be reopened when appropriate.

PR #1137, opened 2021-11-05
Enhance psq analysis functions

### LINKED ISSUE #1072: Make MSQ_RMS_LONG_THRESHOLD  and MSQ_RMS_SHORT_THRESHOLD  configurable for patch seq functions
- [x] Add new optional analysis parameter `SamplingFrequency` defaulting to 50kHz. We then check in `POST_SWEEP_EVENT` that we acquired data with this frequency. Failing to do so is a QC failure.
- [x] Change the noise constants into analysis param; const double MSQ_RMS_SHORT_THRESHOLD = 0.07¶
const double MSQ_RMS_LONG_THRESHOLD = 0.5, optional with current value as default. Store used value in LBN.

t-b:
- [x] Update baseline tests and read out new LBN values
- [x] Add one baseline test with changed thresholds
- [x] Update flowcharts
- [x] Add tests for sampling frequency
- [x] Raise analysis function versions
- [x] Commit cleanup
- [x] Mention default rms values in help message of analysis parameters

Close #1072.


t-b:
@timjarsky Go for it!

timjarsky:
@t-b When ```The used sampling frequency does not match the "sampling frequency" analysis parameter``` have set fail immediately

t-b:
> When The used sampling frequency does not match the "sampling frequency" analysis parameter have set fail immediately

Okay, will do.

t-b:
@timjarsky Catch it while it's hot.

timjarsky (approved):

PR #1142, opened 2021-11-10
Bugfix/1142 find level wrapper preserving input data fail

t-b:


timjarsky (commented):
@t-b should the wrapper also handle chunks?

timjarsky (approved):

t-b:
@timjarsky It could in principle, but I did not see any need.

PR #1161, opened 2021-11-15
Fix DataBrowser SF tab ToFront button not working

MichaelHuth:
Close https://github.com/AllenInstitute/MIES/issues/1148


MichaelHuth:
I would like to note a general drawback of the current approach of window name handling:
Operating based on the name template is ok for the window creation but generally it would be better to save a list of the actual window names from creation returned by Display in S_Name, e.g. in the user data of the data browser.
Such that functions refering to the windows can rely on that list instead of applying a filter based on the name template/ window types again.

t-b (approved):

PR #1165, opened 2021-11-16
tools/sign-installer.sh: Return error code from sign_tool

t-b:
This makes CI fail when the signing fails.

Broken since 7399680d (tools/sign-installer.sh: Add script to sign the
installer, 2021-09-24).


t-b:
Merge once CI passes.

timjarsky (approved):

PR #1166, opened 2021-11-16
LBV_CheckIfXAxisIsTime: Tweak default value depending on existing sweeps

t-b:
The default x axis of the logbook graph was always "sweeps". This is
good for acquired sweep data.

But if you play around with TP only, you currently don't see anything
when adding an entry as there are no sweeps, and plotting something
against NaN does not create anything in the graph.

So let's use a time axis if we don't have sweeps.


timjarsky (approved):

timjarsky:
@t-b funny coincidence: I had this issue during my MIES demo last Friday. Thanks for reading my mind.

PR #1179, opened 2021-11-19
Feature/1179 suppress spurious errors from thread queue getters

t-b:
Assigning to the chief of correct error handling ;)

### DIFF
--- a/Packages/MIES/MIES_ThreadsafeUtilities.ipf
+++ b/Packages/MIES/MIES_ThreadsafeUtilities.ipf
@@ -6,6 +6,7 @@
 #endif
 
 static Constant TS_GET_REPEAT_TIMEOUT_IN_MS = 1
+static Constant TS_ERROR_INVALID_TGID       = 980 // Invalid Thread Group ID or index.
 
 /// @file MIES_ThreadsafeUtilities.ipf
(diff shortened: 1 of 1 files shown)


timjarsky:
> Assigning to the chief of correct error handling ;)

I'm relieved to see that I'm not that person.

MichaelHuth (approved):
The changes are overall reasonable.

I don't overview the actual reason for the race condition to appear, but I suppose that the threa din questions quits itself. This is generally in IP problematic as the main thread notices this only through the invalid tgID. I tend to prefer using some quit request communication to the thread and use threadGroupWait in the main thread  that can then update some threadRunning status flag. Such way the thread ending is somewhat snychronized.

Two minor things:
- For err == TS_ERROR_INVALID_TGID the error is silently ignored. Shouldn't there be at least some DEBUGPRINT output if that happens?
- It's probably not written anywhere but generally Igor error codes are integers, so I would use num2istr consequently.

t-b:
> For err == TS_ERROR_INVALID_TGID the error is silently ignored. Shouldn't there be at least some DEBUGPRINT output if that happens?

I don't think we need to keep track of that at the moment.

> t's probably not written anywhere but generally Igor error codes are integers, so I would use num2istr consequently.

I tend to always use num2str for converting to string. But I guess num2istr would be 110% correct here.

PR #1186, opened 2021-11-20
Feature/1186 use proper variable name instead paneltitle

### LINKED ISSUE #1035: Avoid using `panelTitle` anymore
When I started working on MIES, I immediately had the urge to rename the `panelTitle` parameter, which is actually a device name. But I did not want to push things too far in the beginning.

Now we have ~5500 hits for a name which is really confusing and a misnomer.

- [ ] Change it
- [ ] Add commit to .git-blame-ignore-revs

t-b:
We can of course also choose something else than device. I just think that device is short and not too ambigous.

Close #1035.

### DIFF
--- a/Packages/MIES/MIES_AcquisitionStateHandling.ipf
+++ b/Packages/MIES/MIES_AcquisitionStateHandling.ipf
@@ -48,16 +48,16 @@ End
 ///
 /// We track the acquisition state so that we can perform actions for state transitions when required.
 ///
-/// @param panelTitle  device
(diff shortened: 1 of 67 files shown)


t-b:
Rebased.

t-b:
@timjarsky If we want to have this PR in the upcoming release I prefer to merge it rather sooner than later.

timjarsky (dismissed):

t-b:
Will merge once CI passes.

PR #1197, opened 2021-11-24
Better error detection on device unlocking

t-b:
@timjarsky Can you try again with that branch? And if you can reproduce the error please upload the logfiles and post the ticket ID here.


timjarsky:
@t-b I was able to trigger the issue but it was not detected when testing this PR

t-b:
> Is there anything different that happens when closing Igor without saving vs with saving (other than saving)?

Not for the ITCXOP2. In both cases I get a CLEANUP message which closes all devices.

timjarsky (approved):

PR #1200, opened 2021-11-25
Extend developers documentation for headstage vs channel from LBN functions

MichaelHuth:
Added more developers docu on how to retrieve associated headstages/channels from the lab notebook and how to write them to the LBN for tests.

### DIFF
--- a/Packages/doc/developers.rst
+++ b/Packages/doc/developers.rst
@@ -400,6 +400,92 @@ The programmer might consider evaluating ``V_AbortCode`` after catch.
 It is recommended to comment in the code before the try what the construct is
 intended to handle (RTE, Abort or both).
 
+Retrieving Headstage / Channel Information from the LBN
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+If you would like to retrieve the settings from the last acquisition then look up function like ``AFH_GetHeadstageFromDAC``.
+It retrieves the correct information under the following conditions:
+
+- Data Acquisition is ongoing or
+- Data Acquisition has finished and DAEphys panel was not changed.
+
+This function returns NaN if the active DAC had no associated headstage.
+The same applies for ``AFH_GetHeadstageFromADC``.
+
+In contrast the functions AFH_GetDACFromHeadstage and AFH_GetADCFromHeadstage return DAC/ADC numbers only for active headstages.
+
+One of the most used functions to retrieve specific information from the LBN is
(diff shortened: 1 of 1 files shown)


t-b:
Thanks for working on that!
In case you haven't already read it, please have a look at https://alleninstitute.github.io/MIES/labnotebook-docs.html.
I also think that this a better place for this new content here.

Review:

> If you need the current settings then look up function like ``AFH_GetHeadstageFromDAC``.

This is only useful during data acquisition or immediatly after, in all other cases it will probably give the wrong answer.

>In the returned wave the row refers to the headstage.

Really? The first NUM_HEADSTAGS entries refer to headstage dependent data, the INDEP_HEADSTAGE entry to the headstage independent entry.

> Index refers here to a row, column or layer, depending what function is used.

This is confusing. This might be due to the fact that our API is confusing but still...

> The channel information is saved in the DAC/ADC columns.
I don't understand the concept of DAC/ADC columns.

If you speak about how the data is organized  in the LBN, you need to say that headstage 0 has its entry in layer 0, etc.

> This call specifies a sweep number, a channel type and a channel number and asks for information from the DAC field.
> It returns a 1D wave settings and an index.

The example should not query the "DAC" field as this is a bit too artificial for my case.

> If channel type refers to a ADC channel (XOP_CHANNEL_TYPE_ADC) the returned settings and index contain information about the
> DAC channel and associated headstage. Index equals the headstage and ``settings[index]`` is the DAC channel.

"associated headstage": What do you mean here?

GetLastSetttingChannel also support unassociated channels.

> For the rare case where only the headstage should be retrieved, independently if its activated or not
> ``GetHeadstageForChannel`` can be called with the same channel type as the entry name, e.g.
> XOP_CHANNEL_TYPE_DAC and "DAC". The returned index is the headstage.

Is that still true? I thought we forbid GetHeadstageForChannel to work like that.

> binary flag

boolean entry maybe?

> valid LBN

consistent LBN is better IMHO.

> would violate LBN format rules.

"format rules" sounds odd.

> The corresponding function for single entries is ``ED_AddEntryToLabnotebook``.

Nope. ED_AddEntryToLabnotebook is a function to add user labnotebook entries as the source type will be forced to UNKNOWN_MODE.
ED_AddEntriesToLabnotebook is the generic MIES function to add labnotebook entries. The confusion comes probably from the shi**ty naming.

t-b (changes requested):
See comment.

t-b:
@MichaelHuth Thanks!

t-b (approved):

Issue #28, opened 2018-07-27
SweepFormula: Add tpdelta(sweeps, channels, type) function


t-b:
Where type is steady state resistance or instantaneous resistance. Later versions should also support: baseline average, capacitance and baseline noise quantification.

The function will evaluate all epochs (stock and user) with a shortname matching `^(U_)?TP[[:digit:]]*$`, the baselines and pulse is expected to match `^(U_)?TP[[:digit:]]*_(P|B0|B1)$`. If multiple testpulses are present the raw data is averaged before continuing.

The code in CalculateTPLikePropsFromSweep already does part of the evaluation.

Output: one value per headstage (a sweep can have multiple headstages active) -> 3D wave like data with dimensions
ROW: value
COL: sweeps
LAYER: channels

t-b:
@timjarsky Updated.

PR #1214, opened 2021-12-06
Feature/1214 analysis function pipette in bath

### LINKED ISSUE #1212: Analysis function: Pipette in bath
FTP: `2021-12-06 Pipette in Bath`

And here is the relevant content:
```
Baseline QC analysis function / pipette-in-bath

have max sweep failures analysis param
only 1 sweep is required to pass, if it passes skip remaining sweeps


Pre DAQ
check that indexing is disabled
Verify clamp mode = v-clamp (no check for autobias)
Set Vcom to disabled/unchecked if not already unchecked
Include sampling frequency checks - default to 200 khz and sampling int multiplier = 1


Run Auto pipette offset (should switch clamp modes) - I'm assuming this can only be run in preDAQ

pre sweep

create user epochs for each TP constisting of three epochs and use that in the sweep formula later on.

Run the sweep (should contain 3 test pulses)

Mid sweep

run RMS noise threshold checks (short and long). 
Fail=stop sweep, rerun
	
baseline QC check Average of baseline chunks should be less than MaxLeakCurrent. Fail= stop sweep, run Auto pipette offset, Pass = continue

targetV baselineQC is not done here.

no post pulse baseline QC checks.

Post sweep


Check the pipette resistance against required analysis function param; min/max pipette resistance.
Sweep QC = pipette resistance QC + baseine QC

Calculate pipette resistance by averaging raw data from 3 test pulses that are in the sweep. Steady state resistance = pipette resistance

pulse duration is dynamic

investigate if generating sweepformula code for this works

Consider automate entry into JEM form. We could publish it and WSE can do data entry.

If fails, do usual can-still-pass-checks, if these fail jump to end of set.

Post set

Publish:
1. Set QC
2. leak current QC state
3. leak current value
4. pipette resistance QC state
5. pipette resistance value
6. standard message content


if both pass, in the....

Post DAQ

pick next stimset


List of analysis param:
	SamplingFrequency (optional, default: 200kHz)
	SamplingMultiplier (optional, default: 1)
	BaselineRMSLongThreshold (required)
	BaselinRMSShortThreshold (required)
	MaxLeakCurrent (required)
	MinPipetteResistance (required)
	MaxPiptetteResistance (required)
	NumberOfFailedSweeps (required)
	NextStimSetName (required, changed in POST_DAQ_EVENT)
```

### LINKED ISSUE #1262: SweepFormula: Add ignoreTP entry to notebook docu
Should be done after #1256 and #1214 are merged.

t-b:
- [x] Tests
- [x] AD_GetBaselineFailMsg: Add results for each test type
- [x] Commit cleanup
- [x] SF testpulse from #28
- [x] Flowchart
- [x] Readd "Insert TP"
- [x] Add optional argument to `tp` for ignoring TPs. Ignore the first one, sort them by starting time point of the pre pulse baseline of the TP.
- [x] Fix tests
- [x] #1238
- [x] Turn assertion on invalid stimset into error message and enhance it
- [x] Add checks in each PSQ_PipetteInBath test that a message was published
- [x] Adapt flowchart for message publishing
- [x] Add documentation for published message with example
- [x] Fix issue in LBN_GetEntryProperties with name being NULL
- [x] IVS_runBaselineCheckQC:
    - [x] Select stimset
    - [x] Acquire data
    - [x] remove dead code

Close #1212
Close #1262


t-b:
Ready for a first test.

timjarsky:
@t-b this was installed on an IVSCC rig today. I'll let you know how the testing goes.

timjarsky:
@t-b when attempting to run the in bath sweep I get:

``` The analysis function PSQ_PipetteInBath aborted with error "", this is dangerous and must *not* happen!
  !!! Assertion FAILED !!!
  Message: "Invalid number of test pulses"
``` 

but the number of pulses specified by the analysis param `NumberOfTestpulses` matches the number in the stimset. I've tried changing the number and turning tp insert on and off.

t-b:
Ready for test. Inserted TP is present again.

timjarsky:
@t-b 

control `button_sweepformula_all_code` is not anchored correctly; disappears on window resize.

The implemented SF ```store("Steady state resistance", tp(ss, channels(AD), sweeps()))``` sweep inclusion is too broad. I think it needs to be specific to the passing sweeps of the stimset cycle.

t-b:
> control button_sweepformula_all_code is not anchored correctly; disappears on window resize.|

Thanks, fixed.

> The implemented SF store("Steady state resistance", tp(ss, channels(AD), sweeps())) sweep inclusion is too broad. I think it needs to be specific to the passing sweeps of the stimset cycle.

OVS is turned off, so `sweeps()` only refers to the current sweep.

According to the flowchart 

patch-seq-pipette-bath

a sweep passes if all three properties (baseline QC, sampling frequency QC, pipette resistance QC) pass. And the code does check that the calculated TP resistance is from the expected sweep. So I don't see an issue here.

timjarsky:
@t-b `tp(ss,channels(AD0),sweeps(),[0])` returns ~1500MΩ or nan (depending on the noise in the recording). Expected ~500 M
[pipetteInbathtesting-stimsets.zip](https://github.com/AllenInstitute/MIES/files/7990005/pipetteInbathtesting-stimsets.zip)
Ω

t-b:
Thanks, should be fixed. I forgot to take the onset delay into account when creating the TP-like epochs.

t-b:
Reassigning to me. I forgot the message publishing.

t-b:
Ready for another test. Now also with a published message.

t-b:
Now with published message and tests for it.

timjarsky:
@t-b TP formula isn't accounting for DA scale correctly. See example in image below where DA of TPs is twice the size of inserted TP. The resistance is calculated correctly for the inserted TP but not the stimulus set TPs (both should be about 500 MΩ).

image

t-b:
Found the bug.

t-b:
@timjarsky Go for it.

timjarsky:
@t-b looks like the new result got divided by 10 instead of doubled

image

t-b:
Here is what I did:

- Acquire successfull PSQ_PipetteInBath

Change Formula to
```
# inserted TP
tp(ss, channels(DA), sweeps(), [1, 2, 3])

and

# first user TP
tp(ss, channels(DA), sweeps(), [0, 2, 3])
```

Now I get the same calculated values for the both. The main difference between inserted TP and user TP, is that the amplitude of the former is only the amplitude IC/VC but the latter (user TP) has DAScale times stimset epoch  amplitude. Now I store these amplitudes in the epoch tag info and use these for setting clampAmp used in https://github.com/AllenInstitute/MIES/blob/9781056d92d32bcd0cd18d2a44384ff5225a647c/Packages/MIES/MIES_TestPulse.ipf#L986.

t-b:
Done. Documentation of the JSON message is found at the function comment for PSQ_PB_Publish.

timjarsky (approved):

PR #1217, opened 2021-12-08
Tests: Check TP Storage wave

t-b:
- [x]  Check in tests that gain is 1

Successor of #1105.


timjarsky (approved):

t-b:
@timjarsky This PR requires that the cabling on the ITC18 and NI used on the CI machines is 1:1. IIRC this needs to be still done or?

t-b:
The tests with NI hardware still fail. The second headstage has DAC 0 and ADC 0 connected but these seem to be not connected.

timjarsky:
- [ ] check wiring on NI DAQ

t-b:
Will merge once CI passes.

t-b:
@timjarsky I added support for the gains and filters to `AI_SendToAmp` here. I needed the gains already, so I thought while at it, I do the filters.

PR #1219, opened 2021-12-09
Add tpdelta operation to SweepFormula

MichaelHuth:
Adds tpdelta operation to SweepFormula

close https://github.com/AllenInstitute/MIES/issues/28


t-b:
Review:

885b062b (Docu: Add resistance units to TP algorithm documentation, 2021-12-13)

Good!

92f7ed43 (Introduce CLAMPMODE_ENTRY_KEY constant for "Clamp Mode" key of LBN, 2021-12-13)

- [ ] Some entries are not converted in Packages/Testing-MIES. Found with `git grep -i "\"Clamp Mode\""`.

5b733910 (Introduce TPAMPLITUDEVC_ENTRY_KEY/TPAMPLITUDEIC_ENTRY_KEY constants for LBN, 2021-12-13)

- [ ] Same problem, found with `git grep -iP "\"TP Amplitude (V|I)C\""`. 
The constants should also read more like TP_AMPLITUDE_VC_ENTRY_KEY/TP_AMPLITUDE_IC_ENTRY_KEY.

961c581f (TestPulse: Split function TP_SendToAnalysis into DF preparation and execute, 2021-12-13)

Nice!

23110b87 (SweepFormula add tpdelta operation, 2021-12-09)

- [ ]
```
+    WAVE/Z settings
+    for(i = 0; i < sweepCnt; i += 1)
```
Why the settings wave declaration here?

- [ ] This is not working here.

Test setup: 
- HardwareTests.pxp
- runwithOpts(testcase = "PS_CR1")
-
```
tpdelta(
sweeps(),
channels(AD),
static
)
```

This is giving NaN. I would say  this is because

```
+            channelNr = activeChannels[j][%channelNumber]
+
+            WAVE/Z/T epochMatches = EP_GetEpochs(numericalValues, textualValues, sweep, XOP_CHANNEL_TYPE_DAC, channelNr, epochTPRegExp)
``

as that takes the passed channelnumber which is from ADC here and uses that as DAC. You have to go through  the headstage to get DAC for a ADC.

- [ ] No useless empty checks please. The function EP_GetEpochs states that it either returns a valid non-empty wave or a null wave. I'm fine with converting the check to an ASSERTion here.
```
+            numTPEpochs = DimSize(epochMatches, ROWS)
+            if(numTPEpochs == 0)
+                continue
+            endif

- [ ] The testpulse averaging code should be in its own function. Is it easily doable to avoid the duplicate calls for the averaging?

- [ ] 
```
+            // In theory every sweep could have its own units, though here the unit from the last sweep is used for all sweeps
```

I think this assumption does not hold at all. You can have different clamp modes on different sweeps and also different clamp modes on different channels in one sweep. IIRC this is the reason why sweepData does not have units here.

708528d0 (SweepFormula tpdelta tests, 2021-12-14)

- [ ] Please create a dedicated hardware test for it instead of picky packing on an existing one. This should have two headstages one IC, one VC and ADC!=DAC and acquire three sweeps. You can then check that the data layout is as expected. PatchSeqTestChirp#AcquireData does a similiar setup.

a0ae88a8 (Add In-App help for SweepFormula operation tpdelta, 2021-12-14)
bc62bf21 (Docu: Added docu for SweepFormula operation tpdelta, 2021-12-14)

Documentation looks good!

t-b (changes requested):
See comment.

t-b (dismissed):
LGTM

timjarsky:
@t-b (1) `MΩ` axis label is on x-axis but I think it should be on y-axis

image

(2) The ordering of channels() and sweeps() input should be consistent across all formulas

(3) prefer SS over static, inst over instantaneous, and base over baseline.

t-b:
> (1) MΩ axis label is on x-axis but I think it should be on y-axis

We'll lock into if that is easily fixable.

> (2) The ordering of channels() and sweeps() input should be consistent across all formulas

We don't seem to do a good job in having that consistent. We currently have

```
labnotebook(string key, array channels, array sweeps [, string entrySourceType]) 
epochs(array sweeps, array channels, string name[, string type])
data(array range, array channels, array sweeps)
tpdelta(array sweeps, array channels, string type)
```

> (3) prefer SS over static, inst over instantaneous, and base over baseline.

Will do.

and that looks really awkward when looked at like that.

I suggest to standardize on the argument order

```
..., array channels, arrays sweeps [, optional arguments]
```

like data and labnotebook does.

This would mean we change epochs and tpdelta. @MichaelHuth Any objections?

timjarsky:
@t-b  would it be possible to also accept numbers in place of base, inst, SS? I would suggest 0,1,2 in that order.

t-b:
> @t-b would it be possible to also accept numbers in place of base, inst, SS? I would suggest 0,1,2 in that order.

Yes, I think so. Although it is rare to prefer magic numbers of names ;)

t-b:
> (1) MΩ axis label is on x-axis but I think it should be on y-axis

The axis label is generated by IP from your dimension label. I think that this is currently not correct in IP, I've reported it as a bug.

MichaelHuth:
@timjarsky I put the changes in you requested. Please test again. The epochs operation change to unify the parameter order will be in another PR.

t-b:
And the bug in IP was not a bug. The unit should also be fixed now.

t-b:
@MichaelHuth The SweepFormulaHardware#SF_TPDeltaTest test fails.

MichaelHuth:
@t-b fixed.

timjarsky:
@t-b change the formula string from `TPdelta` to `TP`

MichaelHuth:
Changed formula from "tpdelta" to "tp".

t-b (dismissed):

t-b:
@MichaelHuth Can you fix the merge conflict?

t-b (approved):

PR #1221, opened 2021-12-13
LBN: Add units to TP settings entries

MichaelHuth:
to TP Amplitude VC: pA
   TP Amplitude IC: mV

The units are required for axis labels of graphs that show this data,
e.g. in DataBrowser.

We want to show in graphs to correct units on the axes, so we need the unit names for certain values somewhere in MIES. A good location is the LBN, because it allows to retrieve the units from the time the value was created.

### DIFF
--- a/Packages/MIES/MIES_WaveDataFolderGetters.ipf
+++ b/Packages/MIES/MIES_WaveDataFolderGetters.ipf
@@ -6933,11 +6933,11 @@ Function/WAVE GetTPSettingsLabnotebookKeyWave(string device)
 	wv[%Tolerance][0]  = ""
 
 	wv[%Parameter][1]  = "TP Amplitude VC"
-	wv[%Units][1]      = ""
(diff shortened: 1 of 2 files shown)


Packages/MIES/MIES_WaveDataFolderGetters.ipf
```diff
@@ -6929,15 +6929,15 @@ Function/WAVE GetTPSettingsLabnotebookKeyWave(string device)
 	SetDimLabel 0, 2, Tolerance, wv
 
 	wv[%Parameter][0]  = "TP Baseline Fraction" // fraction of total TP duration
-	wv[%Units][0]      = ""
+	wv[%Units][0]      = "%"
```
t-b, line 6932: We store a number between 0.25 and 0.49 and for me these are not `%`. We don't just store what the DAEphys GUI says.

t-b (approved):

PR #1237, opened 2021-12-22
Bugfix/1237 various

t-b:


t-b:
@timjarsky Fixed on error with powerSpectrum and TP which could explain the crash.

timjarsky:
@t-b I still get the Igor crash with a 1000ms TP, but I think I figured out a way to prevent the computer crash. Let's screen share in the next couple of days when you have some time.

timjarsky:
> @t-b I still get the Igor crash with a 1000ms TP, but I think I figured out a way to prevent the computer crash. Let's screen share in the next couple of days when you have some time.

with TP % duration = 35%, but not 25%

t-b:
@timjarsky I've removed the power spectrum fixes from this PR and extracted them into #1238. Please check if this PR still crashes.

t-b:
@timjarsky Is that good to go?

timjarsky (approved):

PR #1238, opened 2021-12-23
Record sweep formula and add store operation

t-b:
- [x] Document new results wave in labnotebook_docs.rst
- [x] Add all things like sweeps, channels, cursors, device, experiment
- [x] Add proper GUI controls for querying old SF code
- [x] Allow browsing result waves in databrowser/sweepbrowser
- [x] notebook/rst documentation for store
- [x] Tests for store
- [x] Adapt TextWaveToList for 4D with tests
- [x] Write results wave into NWBv2, verify in tests
- [x] Read back in analysis browser, select the correct ones for the sweepbrowser
- [x] Merge https://github.com/AllenInstitute/IPNWB/pull/31 and update IPNWB
- [x] https://github.com/AllenInstitute/MIES/pull/1240
- [x] PopupMenu for old code does not work. 
- [x] Use Extended popup menu for old code

Close #1222
Close #431

PR #1240, opened 2022-01-05
Feature/1240 enhance some misc functions

### LINKED ISSUE #939: NumericWaveToList: Add support for 2D waves
This works with wfprintf as well, see

```igorpro
Function Example1()
	Make/O/N=(10, 2) wave0
	wave0[][0] = sin(p*pi/10)
	wave0[][1] = 1

	wfprintf 1, "", wave0
	
	print "##############"

	wfprintf 1, "%g %g\r", wave0
End
```

t-b:
Preparations for #1238.

@MichaelHuth Please review and can you also provide a fix for `TWTLRoundTrips:4`? IMHO the error is in ListToTextWaveMD.

Close #939.

### DIFF
--- a/Packages/MIES/MIES_Utilities.ipf
+++ b/Packages/MIES/MIES_Utilities.ipf
@@ -3116,6 +3116,26 @@ threadsafe Function/WAVE ListToTextWaveMD(list, dims, [rowSep, colSep, laySep, c
 	return output
 End
 
+#if IgorVersion() < 9.0
+
+threadsafe Function/S ReplicateString(string str, variable numTotalCopies)
+
+	variable i
+	string list = ""
+
+	if(!IsFinite(numTotalCopies) || numTotalCopies <= 0)
+		return str
+	endif
+
+	for(i = 0; i < numTotalCopies; i += 1)
+		list += str
+	endfor
+
+	return list
+End
+
+#endif
+
 /// @brief Convert a numeric wave to string list
 ///
 /// Counterpart @see ListToNumericWave
--- a/Packages/MIES/MIES_Utilities.ipf
+++ b/Packages/MIES/MIES_Utilities.ipf
@@ -4485,19 +4485,27 @@ End
 /// Input numbers are rounded using the "round-half-to-even" rule to the given precision.
 /// The default precision is 5.
 /// If val is complex only the real part is converted to a string.
+///
 /// @param[in] val       number that should be converted to a string
 /// @param[in] precision [optional, default 5] number of precision digits after the decimal dot using "round-half-to-even" rounding rule.
 ///                      Precision must be in the range 0 to #MAX_DOUBLE_PRECISION.
+/// @param[in] shorten   [optional, defaults to false] Remove trailing zeros and optionally the decimal dot to get a minimum length string
+///
 /// @return string with textual number representation
-threadsafe Function/S num2strHighPrec(variable val, [variable precision])
+threadsafe Function/S num2strHighPrec(variable val, [variable precision, variable shorten])
 	string str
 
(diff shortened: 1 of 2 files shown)


Packages/MIES/MIES_Utilities.ipf
```diff
@@ -3144,10 +3144,12 @@ End
 ///
 /// @param wv     numeric wave
 /// @param sep    separator
+/// @param colSep [optional, default = `,`] separator for column entries
 /// @param format [optional, defaults to `%g`] sprintf conversion specifier
-threadsafe Function/S NumericWaveToList(WAVE/Z wv, string sep, [string format])
-
+threadsafe Function/S NumericWaveToList(WAVE/Z wv, string sep, [string format, string colSep])
```
timjarsky, line 3228: @t-b are in-line return parameters (e.g., ```Function/S [string list] NumericWaveToList(WAVE/Z wv, string sep, [string format, string colSep])``` ) only used when there is >1 return param?

timjarsky (dismissed):

Packages/MIES/MIES_Utilities.ipf
t-b: Jeep. This is done as using the single-return signature is nicer at the call site.

In the following example three different use cases are tried with classic and MRS (multiple return value syntax) functions. Only the classic one can be used in all three cases, MRS fails in two.

```igorpro
Function/S FuncClassic()

	return ""
End

Function [string str] FuncMRS()

	return [""]
End

Function Consume(string str)

End

Function Test()
	string str1, str2

	// ignoring return value works
	FuncClassic()

	// using it
	str1 = FuncClassic()

	// pass onto caller
	Consume(FuncClassic())

	// does not compile
	// FuncMRS()
	
	[str2] = FuncMRS()
	
	// does not compile
	// Consume(FuncMRS())
End

```

Packages/MIES/MIES_Utilities.ipf
```diff
@@ -3116,6 +3116,26 @@ threadsafe Function/WAVE ListToTextWaveMD(list, dims, [rowSep, colSep, laySep, c
...
+threadsafe Function/S ReplicateString(string str, variable numTotalCopies)
+
+	variable i
+	string list = ""
+
+	if(!IsFinite(numTotalCopies) || numTotalCopies <= 0)
+		return str
+	endif
+
+	for(i = 0; i < numTotalCopies; i += 1)
```
MichaelHuth, line 3209: Just a minor thought:
I have not benched it, though I would have used a two liner with
PadString + ReplaceString here.

Packages/MIES/MIES_Utilities.ipf
```diff
@@ -3156,14 +3158,28 @@ threadsafe Function/S NumericWaveToList(WAVE/Z wv, string sep, [string format])
 		format = "%g"
 	endif
 
+	if(ParamIsDefault(colSep))
+		colSep = ","
```
MichaelHuth, line 3241: Would it make sense to have some default separator constants defined globally that are used preferrentially?
If yes, lets make an issue for that.

t-b:
Just for the record. The error was in my TextWaveToList changes and not in ListToTextWaveMD.

MichaelHuth (approved):
See comments, the rest looks good to me.

Packages/MIES/MIES_Utilities.ipf
t-b: Might be worth a thought. But IP8 is also on its way out so I'd leave it as it is.

Packages/MIES/MIES_Utilities.ipf
t-b: Good point. An issue would be good.

PR #1248, opened 2022-01-10
Bugfix/1248 power spectrum

### LINKED ISSUE #1232: Powerspectrum aborts with two headstages
Fix it and add test.

t-b:
This now has the power spectrum fixes from #1237.

Close #1232

### DIFF
--- a/Packages/MIES/MIES_DataConfigurator.ipf
+++ b/Packages/MIES/MIES_DataConfigurator.ipf
@@ -2032,12 +2032,9 @@ static Function DC_ReturnTotalLengthIncrease(device, [onsetDelayUser, onsetDelay
 End
 
 /// @brief Calculate the stop collection point, includes all required global adjustments
-static Function DC_GetStopCollectionPoint(device, dataAcqOrTP, setLengths)
-	string device
-	variable dataAcqOrTP
-	WAVE setLengths
-
+static Function DC_GetStopCollectionPoint(string device, variable dataAcqOrTP, WAVE setLengths)
 	variable DAClength, TTLlength, totalIncrease
+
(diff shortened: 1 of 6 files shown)


t-b:
I can reproduce the crash.

The reason for the crash is that the ITC FIFO overflows because with this exact setup (2HS, 1000ms TP, 35% baseline) the execution time of FFT is absurdely slow.

Just try
```
•make/N=(833334,4)/o data = p 
•fft/COLS/DEST=ausg data 
```

and see that it takes ages (~5min). The reason for that slowness is documented

```
The FFT algorithm is based on prime number decomposition, which decomposes the number of points in each dimension of the wave into a product of prime numbers. The FFT is optimized for primes < 5.  In time consuming applications it is frequently worthwhile to pad the data so that the total number of points factors into small prime numbers.
```

and for this example data has 833334 rows and these factor into the primes 2, 3, 138889.

A short example
```
•make/N=(2^FindNextPower(833334, 2),4)/o data = p 
•fft/COLS/DEST=ausg data 
```

verifies that assumption as now data is even larger, but FFT is *much* faster.

@timjarsky Would it be acceptable to zero pad the incoming data before doing FFT?

timjarsky:
@t-b I think padding the data is a good idea. Is padding with zero the best choice?

t-b:
> Is padding with zero the best choice?
From my knowledge yes and also FFT/PAD pads with zeros.

t-b:
@timjarsky I've pushed a WIP fix. I need to look at that again tomorrow with fresh eyes.

t-b:
@timjarsky Go for it!

timjarsky (approved):

PR #1250, opened 2022-01-13
Bugfix/1250 various preparing pipette in bath

t-b:
@MichaelHuth Is 8804a3b81fbe819b4b70e8694dd76d0997958d02 doing the right thing?

### DIFF
--- a/Packages/MIES/MIES_AnalysisFunctions_PatchSeq.ipf
+++ b/Packages/MIES/MIES_AnalysisFunctions_PatchSeq.ipf
@@ -1244,6 +1244,20 @@ static Function PSQ_GetDefaultSamplingFrequency(variable type)
 	endswitch
 End
 
+/// @brief Returns the sampling frequency in kHz of our supported NI and ITC
+/// hardware for one active headstage and no TTL channels.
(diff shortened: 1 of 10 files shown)


timjarsky (approved):

MichaelHuth:
> @MichaelHuth Is [8804a3b](https://github.com/AllenInstitute/MIES/commit/8804a3b81fbe819b4b70e8694dd76d0997958d02) doing the right thing?

Hmm, my original distinction between using ASSERT and SF_ASSERT in the tp operation function was exactly based on the reason given in the commit message that changes all to SF_ASSERT now.  My idea was that SF_ASSERT catches non-critical errors within sweepformula, e.g. if a user specifies function arguments incorrectly. However for the LBN data retrieval for values that should be there by LBN definition I treated that as critical error using ASSERT, because the fault is not within Sweepformula but in the underlying Igor Code from MIES.
Though if also such situations should be treated as non-critical by SF I'am ok with that.

t-b:
@MichaelHuth Thanks for looking into that. The analysis function using `tp` wants to continue even if it aborted  early, and then where will be no TP-like epochs. And with the new SF_Assert this works.

PR #1263, opened 2022-01-25
Feature/1263 turn assert on into bug

t-b:


t-b:
@timjarsky 
Soo the PR would be ready, if I wouldn't haved forgotten one tiny detail. The TUFXOP is only available in IP9. And with my usage now in BUG_TS this makes MIES IP9 only. Which I guess is not feasible at this point?

The reason the TUFXOP is IP9 is that it triggers some bugs in earlier IP versions, which were hard to avoid.

t-b:
@timjarsky Done. I've removed the threadsafe bug counting for IP8 as we run the tests with IP9 anyway.

timjarsky (approved):

PR #1310, opened 2022-03-10
Feature/1310 more psq msq analysis function lbn value checks

t-b:


t-b:
@timjarsky A good thing to check here is proof reading the labnotebook tables. The documented units should be correct now. And yes they are inconsistent in parts, that is the reason why it is now documented so that this is less likely to happen.

timjarsky (approved):

PR #1346, opened 2022-05-02
Feature/1346 baselinechunk adaptations

t-b:
This PR now allows to choose a `BaselineChunkLength` which is smaller as the Wavebuilder epoch.

### DIFF
--- a/Packages/MIES/MIES_AnalysisFunctions_PatchSeq.ipf
+++ b/Packages/MIES/MIES_AnalysisFunctions_PatchSeq.ipf
@@ -1526,7 +1526,7 @@ static Function/S PSQ_GetHelpCommon(variable type, string name)
 
 	strswitch(name)
 		case "BaselineChunkLength":
-			return "Length of a baseline QC chunk to evaluate, defaults to 500 [ms]"
+			return "Length of a baseline QC chunk to evaluate"
 		case "BaselineRMSLongThreshold":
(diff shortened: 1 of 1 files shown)


t-b:
Assigning myself, as something is fishy.

t-b:
Fixed.

timjarsky:
@t-b I'm attempting to test this PR with PSQ_SealEvaluation and I get the error: `The number of present (21) and expected (22) stimulus set epochs differs.` I think 21 epochs is expected and the documentation agrees;

image

t-b:
@timjarsky The documentation starts counting at zero, so the last is 21 and there are 22 in total. Does that make sense?

t-b:
The expected total number of epochs is 22 (this is also what you enter in the wavebuilder. The "epoch to edit" is an index limited to total-1.

timjarsky:
[SealStimset.zip](https://github.com/AllenInstitute/MIES/files/8626144/SealStimset.zip)
@t-b I've attached the seal stimset, can you look to see if I missed an epoch?

t-b:
You have one epoch for 10 and 11. And 11 also needs to be >= 500ms.

timjarsky:
@t-b two baseline epochs for the TP pulses bracket the epoch that splits the groups (epoch 10; see image). Is that not correct? Yes, that's right, I set it to 500 to see how it would fail but I'm getting stuck on the number of epochs.
image

timjarsky:
@t-b sorry, i get it now.

timjarsky (approved):

PR #1380, opened 2022-05-18
Add support for async channel QC

### LINKED ISSUE #1327: Make Analysis functions configurable to respect Bath Temperature min/max range
We are using the Asynchronous channels to record our set and bath temperatures and to trigger the alarm when the bath temperature is out of range (see image below), but we would like to add additional parameters to the analysis functions so that they can be configured to respect the min and max bath temperature and to fail sweeps where the bath temperature is outside of the specified min and max. 
MIES_Async_Temps

There should be a parameter to turn this feature on or off as well as a parameter that specifies which Async channel(s) are being used for temperature monitoring. The following image is an example that I mocked up to illustrate what that might look like in the parameters panel for the PSQ_DAScale analysis function.
MIES_parameters_panel

Please let me know if you have any questions about this request, thanks!

t-b:
Todos:
- [x] Needs #1382 merged
- [x] Needs https://github.com/AllenInstitute/MIES/pull/1394 merged
- [x] Needs #1396
- [x] Store alarm state in ED_createAsyncWaveNoteTags
- [x] Drop tab->spaces cleanup for dot files

- Required analysis parameter: `AsyncChannelsWithQC` (semicolon separated list)
- PRE_SWEEP_CONFIG_EVENT:
   - [x] Check that the alarm is enabled
- POST_SWEEP_EVENT:
   - [x] Is part of sweep QC

Analysis functions:
- [x] PSQ_DAScale
- [x] PSQ_SquarePulse
- [x] PSQ_Rheobase
- [x] PSQ_Ramp
- [x] PSQ_Chirp
- [x] PSQ_PipetteInBath
- [x] PSQ_SealEvaluation
- [x] PSQ_TrueRestingMembranePotential
- [x] PSQ_AccessResistanceSmoke

Flow charts:
- [x] PSQ_DAScale
- [x] PSQ_SquarePulse
- [x] PSQ_Rheobase
- [x] PSQ_Ramp
- [x] PSQ_Chirp
- [x] PSQ_PipetteInBath
- [x] PSQ_SealEvaluation
- [x] PSQ_TrueRestingMembranePotential
- [x] PSQ_AccessResistanceSmoke

Close #1327


timjarsky:
@t-b i get
 ```
  (ITC1600_Dev_1) The required analysis parameters requested by PSQ_AccessResistanceSmoke for stim set smoke_DA_0 were not all supplied (missing are: AsyncQCChannels;)
```
but
image

t-b:
- [ ] labnotebook bug message
- [ ] wrong dashboard reporting

t-b:
Should all be fixed. But please test again.

timjarsky:
@t-b I'm still seeing failing because access resistance is out of range when it appears to be in range.

t-b:
@timjarsky This should be a problem in main as well or?

timjarsky:
@t-b I'm not sure. Is there a hardcoded limit on access that's being hit?

t-b:
@timjarsky No the thing is that we only evaluate the resistances if baselineQC passes. Otherwise the resistances are NaN and thus the QC always fails. I did that to save cycles as we don't pass anyway if baseline QC fails. SHould I change that?

timjarsky:
> @timjarsky No the thing is that we only evaluate the resistances if baselineQC passes. Otherwise the resistances are NaN and thus the QC always fails. I did that to save cycles as we don't pass anyway if baseline QC fails. SHould I change that?

@t-b It makes sense not to measure the access resistance if the baseline fails but the message in the dashboard should say that baseline QC failed rather than access resistance is out of range.

t-b:
> @t-b It makes sense not to measure the access resistance if the baseline fails but the message in the dashboard should say that baseline QC failed rather than access resistance is out of range.

It does now according to my testing.

timjarsky:
@t-b Resistance checking with a clean/new test pxp works!

timjarsky (approved):

t-b:
Fixed tests.

PR #1386, opened 2022-05-21
Feature/1386 add bessel filter chirp

### LINKED ISSUE #1377: PSQ_Chirp: Allow to set MCC filter from analysis parameter and fix repurposed time to 6s
- [x] AmpBesselFilter:
- Optional
- Default to 10e3Hz
- [Hz]
- PRE_SET_EVENT:
- Store original value

- [x] AmpBesselFilterRestore:
- Optional (default to ON)
- On/Off
- On: set it back in POST_SET_EVENT to the previously stored value

- [x] Check what happens if a filter value which is only close a known one is set. Nearest Neighbour?

- [x] Add filter setting to amplifier storage wave and store in LBN. Look at labnotebook as it comes already out of the struct.

- [x] PSQ_EvaluateBaselineProperties needs to set repurposedTime to something so that the ITI is always around 6s. In this way we don't waste too much time.

t-b:
Merged before:

- [x] #1380
- [x] #1382
- [x] #1396
- [x] Flowchart
- [x] Have a look about the naming of "Inter trial interval (effective)". Is that end to start or start to start?

Close #1377

### DIFF
--- a/Packages/MIES/MIES_AmplifierInteraction.ipf
+++ b/Packages/MIES/MIES_AmplifierInteraction.ipf
@@ -1498,8 +1498,6 @@ Function AI_FillAndSendAmpliferSettings(device, sweepNo)
 	WAVE/T ampSettingsTextKey  = GetAmplifierSettingsTextKeyWave()
 	WAVE ampParamStorage       = GetAmplifierParamStorageWave(device)
 
-	ampSettingsWave = NaN
-
 	for(i = 0; i < NUM_HEADSTAGES; i += 1)
 
 		if(!statusHS[i])
--- a/Packages/MIES/MIES_AnalysisFunctions_PatchSeq.ipf
+++ b/Packages/MIES/MIES_AnalysisFunctions_PatchSeq.ipf
@@ -68,6 +68,7 @@
 ///  PSQ_FMT_LBN_CR_BOUNDS_STATE              Upper and Lower bounds state according to min/max pos.      (none)   Textual      CR                                   No           No
 ///  PSQ_FMT_LBN_CR_SPIKE_CHECK               Spike check was enabled/disabled                            (none)   Numerical    CR                                   No           No
 ///  PSQ_FMT_LBN_CR_INIT_UOD                  Initial user onset delay                                    ms       Numerical    CR                                   No           No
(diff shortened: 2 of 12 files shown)


timjarsky:
@t-b 

- ampBesselFilterRestore has the wrong help message. Include the unit in the updated message. It also doesn't seem to restore the correct filter value;
image
[should be restored to 10khz but is setting to 2.2khz]

t-b:
@timjarsky Please upload the file. Thanks.

t-b:
As specified in #1377 `ampBesselFilterRestore` is a boolean which restores the previously set value in the PRE_SET_EVENT.

timjarsky (approved):

t-b:
Please retest. I'm now using the correct 10e3 as default LPF.

t-b:
@timjarsky I've checked the new "Inter-trial interval (effective)" labnotebook entry. And it is start to start. It is also 100% honest and can therefore also be negative.

t-b:
I need to investigate the failing test.

t-b:
So the test should be fixed. Another "bummer"...
See 14d1fd97 (LeftOverSweepTime: Fix it for NI hardware, 2022-06-03).

timjarsky:
@t-b the checks prevent a user from setting an optional analysis param to NaN or blank. 

e.g.,   
for blank:
The analysis parameters are not valid and the stimset can therefore not be saved.

for Nan
  The error message are:
	- AmpBesselFilter: Must be a positive value.

timjarsky:
> @t-b the checks prevent a user from setting an optional analysis param to NaN or blank.
> 
> e.g., for blank: The analysis parameters are not valid and the stimset can therefore not be saved.
> 
> for Nan The error message are: - AmpBesselFilter: Must be a positive value.

@t-b 

the workaround is to delete the param. I'm find as is if you are.

default is setting the Bessel filter correctly.

timjarsky (approved):

t-b:
> the checks prevent a user from setting an optional analysis param to NaN or blank.

Yes, that is intended. If you want to set it, you need to supply a valid value. The constant `LPF_BYPASS` has the frequency which sets bypass. This is also mentioned in the help message.

PR #1395, opened 2022-05-24
Fix labnotebook description wave not being refetched if changed

t-b:
This was found as part of testing #1380. This fixes the "BUG: ED_FindIndizesAndRedimension" messages. This bug should not influence normal users as you have to acquire data with one MIES version and then acquire data with a newer MIES version again into the same file. Nevertheless worth fixing.

### DIFF
--- a/Packages/MIES/MIES_ExperimentDocumentation.ipf
+++ b/Packages/MIES/MIES_ExperimentDocumentation.ipf
@@ -489,7 +489,7 @@ static Function [WAVE colIndizes, variable rowIndex] ED_FindIndizesAndRedimensio
 	numKeyCols = DimSize(key, COLS)
 	lastValidIncomingKeyRow = DimSize(incomingKey, ROWS) - 1
 
-	Make/FREE/U/I/N=(DimSize(incomingKey, COLS)) indizes = NaN
+	Make/FREE/D/N=(DimSize(incomingKey, COLS)) indizes = NaN
 
 	WAVE/T/Z desc
 	if(logbookType == LBT_LABNOTEBOOK)
--- a/Packages/MIES/MIES_ExperimentDocumentation.ipf
+++ b/Packages/MIES/MIES_ExperimentDocumentation.ipf
@@ -491,7 +491,7 @@ static Function [WAVE colIndizes, variable rowIndex] ED_FindIndizesAndRedimensio
 
 	Make/FREE/D/N=(DimSize(incomingKey, COLS)) indizes = NaN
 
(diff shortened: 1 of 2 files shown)


Packages/MIES/MIES_ExperimentDocumentation.ipf
```diff
@@ -491,7 +491,7 @@ static Function [WAVE colIndizes, variable rowIndex] ED_FindIndizesAndRedimensio
 
 	Make/FREE/D/N=(DimSize(incomingKey, COLS)) indizes = NaN
 
-	WAVE/T/Z desc
+	WAVE/T/ZZ desc
```
timjarsky, line 494: cool!

Packages/MIES/MIES_ExperimentDocumentation.ipf
```diff
@@ -489,7 +489,7 @@ static Function [WAVE colIndizes, variable rowIndex] ED_FindIndizesAndRedimensio
 	numKeyCols = DimSize(key, COLS)
 	lastValidIncomingKeyRow = DimSize(incomingKey, ROWS) - 1
 
-	Make/FREE/U/I/N=(DimSize(incomingKey, COLS)) indizes = NaN
+	Make/FREE/D/N=(DimSize(incomingKey, COLS)) indizes = NaN
```
timjarsky, line 492: @t-b So IP9 converts NaN in unsigned integer wave type, not all wave types, is that right?

timjarsky:
@t-b have you tested this by trying it with different MIES versions or should I do that?

Packages/MIES/MIES_ExperimentDocumentation.ipf
t-b: Jeep, getting into more and more IP9 only features.

Packages/MIES/MIES_ExperimentDocumentation.ipf
t-b: You can not store a NaN in an integer wave (you can only store it in floating point waves), so there has to be some conversion going on. And as the code is as it was difficult to grasp, I've decided to use a double wave instead which can store NaN.

t-b:
> @t-b have you tested this by trying it with different MIES versions or should I do that?

I've done that. Works.

timjarsky:
> > @t-b have you tested this by trying it with different MIES versions or should I do that?
> 
> I've done that. Works.

Thanks!!

timjarsky (approved):

PR #1396, opened 2022-05-24
Make async channels work with NI hardware

t-b:
That is a bummer!

We actually did not support NI hardware for async channels. I found that by accident when I looked at some other CI failure. What put me off was that we had some ITC errors in the NI hardware tests, see [1].

The test case AsyncAcquisitionLBN works with NI hardware but had this output.

```
<testcase name="AsyncAcquisitionLBN:NI in UTF_BasicHardwareTests.ipf (4)" classname="AsyncAcquisitionLBN:NI" time="15.161">
<system-out>
Entering test case "AsyncAcquisitionLBN:NI" The ITC XOP returned the following errors: ITCError=0x82503001, ITCXOPError=0 System is not initialized (0x82503001);Place: User Layer; Connection: DSP; Error Type: Ready; Sequence Number: 0x0; Device Type: ITC1600; OS: Any; Application: Regular; Local Sequence Number: 0x1 Some hints you might want to try! - Is the correct ITC device type selected? - Is your ITC Device connected to a power socket? - Is your ITC Device connected to your computer? - Have you tried unlocking/locking the device already? - Reseating all connections between the DAC and the computer has also helped in the past. Entering reentry "AsyncAcquisitionLBN_REENTRY" Trying to resurrect missing sweeps from device Dev1 Sweep 0 is missing its sweep wave Sweep 0 is missing its config wave Trying to recontruct sweep 0 Reconstructed successfully. Leaving test case "AsyncAcquisitionLBN:NI"
</system-out>
</testcase>
```

So turns out the test case actually does not work but failed to complain. I've now added a couple of safety nets so that we are getting complained at in the future if something similiar would happen.

[1]: http://bamboo.corp.alleninstitute.org/browse/MIES-PLAYIN-698/artifact/shared/JUNIT-logfile-%5B5%5D/JU_2022_05_23_102810_2022-05-23_11-56-30.xml.

### DIFF
--- a/Packages/MIES/MIES_AsynchronousData.ipf
+++ b/Packages/MIES/MIES_AsynchronousData.ipf
@@ -12,7 +12,7 @@
 /// @brief Check if the given asynchronous channel is in alarm state
 Function ASD_CheckAsynAlarmState(variable value, variable minValue, variable maxValue)
 
-	return value >= maxValue || value <= minValue
+	return IsNaN(value) || value >= maxValue || value <= minValue
 End
 
 /// @brief Read the given asynchronous channel and return the scaled value
--- a/Packages/MIES/MIES_AsynchronousData.ipf
+++ b/Packages/MIES/MIES_AsynchronousData.ipf
@@ -21,12 +21,13 @@ Function ASD_ReadChannel(device, channel)
 	variable channel
 
 	string ctrl
-	variable gain, deviceChannelOffset, rawChannelValue
+	variable gain, deviceChannelOffset, rawChannelValue, hardwareType
 
 	NVAR deviceID = $GetDAQDeviceID(device)
 	deviceChannelOffset = HW_ITC_CalculateDevChannelOff(device)
 
-	rawChannelValue = HW_ReadADC(HARDWARE_ITC_DAC, deviceID, channel + deviceChannelOffset)
+	hardwareType = GetHardwareType(device)
+	rawChannelValue = HW_ReadADC(hardwareType, deviceID, channel + deviceChannelOffset)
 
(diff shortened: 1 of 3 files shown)


Packages/MIES/MIES_AsynchronousData.ipf
```diff
@@ -21,12 +21,13 @@ Function ASD_ReadChannel(device, channel)
...
 
 	string ctrl
-	variable gain, deviceChannelOffset, rawChannelValue
+	variable gain, deviceChannelOffset, rawChannelValue, hardwareType
 
 	NVAR deviceID = $GetDAQDeviceID(device)
 	deviceChannelOffset = HW_ITC_CalculateDevChannelOff(device)
 
-	rawChannelValue = HW_ReadADC(HARDWARE_ITC_DAC, deviceID, channel + deviceChannelOffset)
+	hardwareType = GetHardwareType(device)
```
timjarsky, line 29: @t-b With this change async channels will work on NI hardware now?

t-b: Yes. The test `AsyncAcquisitionLBN` checks that. And also future similiar problems are mitigated as:
- HW_ITC_ReadADC: Returns NaN on errors calling the ITC XOP2 routines. I've added a workaround so that we don't need an updated ITCXOP2, as I thought these changes would be rather involved. Turns out it was not so difficult [1].
- ASD_CheckAsynAlarmState: Treats an ADC value of NaN as failure.

[1]: So I fixed it already in https://github.com/AllenInstitute/ITCXOP2/pull/20 and https://github.com/AllenInstitute/ITCXOP2/pull/21. The ITCXOP now also follows the "modern" approach of using cmake instead of Visual Studio project files. I've been using cmake for every XOP since directly after the ITCXOP2 ;)

PR #1400, opened 2022-06-03
Output message for MCC bug with instructions

### LINKED ISSUE #1357: MCC configuration issue requires Windows user wiping
From https://aiephys.slack.com/archives/G01CZ45S95X/p1651614478318899:

> attached is an image of an error that we occasionally encounter when configuring MIES. The only way that we have found to fix this issue so far is to wipe the Windows user account settings and then when they login and start up MIES the next time the issue goes away.  I uploaded a file that contains this message to your FTP and it is called mies_gain_config_issue_2022_05_03_140008. Just wanted to see if you could take a look and see if there is anything that you can find on your end that might solve this. Let me know if you have any questions or if I should open a MIES issue for this, thanks.

grafik

t-b:
Close #1357.

### DIFF
--- a/Packages/MIES/MIES_DAEphys.ipf
+++ b/Packages/MIES/MIES_DAEphys.ipf
@@ -2716,7 +2716,13 @@ static Function DAP_CheckHeadStage(device, headStage, mode)
 			needResetting = 1
 		endif
 
-		if((!CheckIfClose(DAGain, DAGainMCC, tol=1e-4) && clampMode != I_EQUAL_ZERO_MODE) || (clampMode == I_EQUAL_ZERO_MODE && !CheckIfSmall(DAGainMCC)))
+		if(clampMode == I_EQUAL_ZERO_MODE && !CheckIfSmall(DAGainMCC))
+			printf "(%s) The DA gain for the DA channel %d in I=0 clamp mode is expected to be zero but the configured gain is %g.\r", device, DACchannel, DAGainMCC
+			return 2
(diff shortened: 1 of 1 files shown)


Packages/MIES/MIES_DAEphys.ipf
```diff
@@ -2716,7 +2716,13 @@ static Function DAP_CheckHeadStage(device, headStage, mode)
 			needResetting = 1
 		endif
 
-		if((!CheckIfClose(DAGain, DAGainMCC, tol=1e-4) && clampMode != I_EQUAL_ZERO_MODE) || (clampMode == I_EQUAL_ZERO_MODE && !CheckIfSmall(DAGainMCC)))
+		if(clampMode == I_EQUAL_ZERO_MODE && !CheckIfSmall(DAGainMCC))
+			printf "(%s) The DA gain for the DA channel %d in I=0 clamp mode is expected to be zero but the configured gain is %g.\r", device, DACchannel, DAGainMCC
```
timjarsky, line 2720: @t-b In I=0, the MCC returns a DA gain of zero? It makes sense that the amp would have a gain of 0, but does it matter what the DAC DA gain is?

t-b: I've just rewritten the existing check here. I can remove it if you prefer that. I think it also makes sense to remove it as were are in all other places checking that the MIES settings are good not that the MCC has the correct settings by itself.

timjarsky (approved):

PR #1417, opened 2022-06-24
Add_support for JSON waveNotes

MichaelHuth:
Utility functions are added that add support for JSON in wave notes.

SetNumberInJSONWaveNote
SetStringInJSONWaveNote
SetWaveInJSONWaveNote
GetNumberFromJSONWaveNote
GetStringFromJSONWaveNote
GetNumericWaveFromJSONWaveNote
GetTextWaveFromJSONWaveNote

These functions take a non-empty json path instead of a key as argument.
In addition to the classic e.g. SetStringInWaveNote function, there are
also functions to add waves to the wave note. The waves are serialized to
JSON text, thus it is intended to be used for small amounts of data.

Limitations:
Added waves must be numerical or textual and have at least one data point.

When reading this data back the type must be selected through the getter function,
GetNumericWaveFromJSONWaveNote or GetTextWaveFromJSONWaveNote.

JSON wave notes are appended to the classic wave notes. Classic wave note getter
functions will return the correct data.

It is not supported that additional classic wave notes are written once a JSON
wave note was appended.

@t-b We should discuss how much compatibility with the classic wave notes we need. In principle the classic wave note functions can be adapted to split the JSON part off at the marker, change only the first part and then reassemble the note again. Though this would slow down these functions a lot.


t-b:
>@t-b We should discuss how much compatibility with the classic wave notes we need. In principle the classic wave note functions can be adapted to split the JSON part off at the marker, change only the first part and then reassemble the note again. Though this would slow down these functions a lot.

I don't think we need compatibility between the two approaches. If you want to switch from classic to JSON, or vice versa, we can convert once with upgrading to a new wave version.

Review:
- Code looks good. Nicely done. Some minor nits: 
- `SetStringInJSONWaveNote`: I don't think we need the `ASSERT_TS(!IsNull(str), "got null string as data")` there. As we never check that a passed string parameter is not null.
- Can we remove the /Z for the wave parameter? You are always asserting out anyway.
- `GetJSONWaveNoteAsJSON`: It should mention that the caller is responsible to release the JSON
- Could you we move the JSON wave note code into a separate file with maybe a `JWN_` prefix?

t-b (changes requested):
See comment

t-b (approved):

PR #1437, opened 2022-07-22
Minor fixes for amplifier handling

t-b:
In response to #1435 I did some playing around and found some minor things which are fixed here.


timjarsky:
@t-b works as expected without an amp.

timjarsky (approved):

PR #1442, opened 2022-07-29
Add test with no amplifier and other related fixes.

### LINKED ISSUE #1441: Checks before DAQ without amplifier
@timjarsky  So I did some digging as you said that the Hardware->"DAC Channel and Device Assocations"  works equally well without amplifier.

And I found fae2379a (DAP_CheckHeadStage: Add checks for DA/AD gain and unit, 2016-05-11) which only does the equality checks if we have an amplifier connected.

So I would now change that we don't require an amplifier for these checks?

t-b:
Close #1441.

### DIFF
--- a/Packages/MIES/MIES_WaveDataFolderGetters.ipf
+++ b/Packages/MIES/MIES_WaveDataFolderGetters.ipf
@@ -2059,7 +2059,7 @@ threadsafe Function/WAVE GetLBNidCache(numericalValues)
 	return wv
 End
 
-static Constant SWEEP_SETTINGS_WAVE_VERSION = 34
+static Constant SWEEP_SETTINGS_WAVE_VERSION = 35
 
 /// @brief Uses the parameter names from the `sourceKey` columns and
 ///        write them as dimension into the columns of dest.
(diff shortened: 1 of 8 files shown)


Packages/MIES/MIES_WaveDataFolderGetters.ipf
```diff
@@ -2455,6 +2458,18 @@ Function/Wave GetSweepSettingsKeyWave(device)
 	wv[%Units][55]     = LABNOTEBOOK_BINARY_UNIT
 	wv[%Tolerance][55] = LABNOTEBOOK_NO_TOLERANCE
 
+	wv[%Parameter][56] = "Double precision data"
+	wv[%Units][56]     = LABNOTEBOOK_BINARY_UNIT
+	wv[%Tolerance][56] = LABNOTEBOOK_NO_TOLERANCE
+
+	wv[%Parameter][57] = "Save amplifier settings"
```
timjarsky, line 2465: @t-b what's the motivation for documenting the `Save amplifier settings` and `require amplifier` checkboxes? Won't we know if these have been checked by the data that's present in the lab notebook?

timjarsky (approved):

t-b:
Test AutoTP_OptimumValues is broken.

PR #1447, opened 2022-08-03
DAP_ButtonProc_ClearChanCon: Also clear the channel clamp mode wave

t-b:
Since 9293eac6 (DA_Ephys: Add a button to the hardware tab allowing to
clear the channel/amplifier connections, 2015-12-16) we allow to clear the
headstage/channel association.

But in cb206d3c (GetChannelClampMode: Extend to also hold headstage
information, 2019-03-08) we introduced the channel clamp mode wave which
also holds the headstage/channel association.

This needs to be cleared as well.


t-b:
@timjarsky Is that approved or should I fix something?

timjarsky (approved):

PR #1448, opened 2022-08-03
SF_OperationTP: Fix TP property calculation

### LINKED ISSUE #1273: Add test for tp with stock and user entry
Once 0c95d8a5 (SF_OperationTP: Read out the clamp amplitude from the tags, 2022-02-09) is merged, via #1214, and #1217.

- Use `PSQ_QC_Stimsets` stimset
- Formula:
```
tp(inst, channels(DA), sweeps(), [1, 2, 3])

and

tp(inst, channels(DA), sweeps(), [0, 2, 3])
```

and check that both are equal.

### LINKED ISSUE #1444: PSQ_AccessResistanceSmoke : Returned access resistance looks odd
From https://aiephys.slack.com/archives/G01CZ45S95X/p1659473221870449:

> I was checking the output of our break in sweep (EXTPBREAKN180424) which uses the PSQ_AccessResistanceSmoke analysis function and it appears that there is something wrong with the value for the access resistance that is being returned. In the attached image I have circled in red the access resistance value that shows up in the MIES test pulse window (6.6 MOhms) and the value that is being returned from PSQ_AccessResistanceSmoke (9667673.716012085 Ohms or 9.7 MOhms),I will also upload the pxp file to your FTP (2022_08_02_104637.pxp) for your reference. I am not sure why these values would be different but can you take a look when you get a chance? Also, if you prefer I can open an issue on GitHub for this, just let me know, thanks.

grafik

t-b:
The duration is in points and not in ms as the code assumed.

Bug introduced in the initial version in 044d386 (SweepFormula add tp
operation, 2021-12-09).

Close #1273
Close #1444

### DIFF
--- a/Packages/MIES/MIES_Structures.ipf
+++ b/Packages/MIES/MIES_Structures.ipf
@@ -295,7 +295,7 @@ Structure TPAnalysisInput
 	WAVE data
 	variable clampAmp
 	variable clampMode
-	variable duration
+	variable duration // [points]
 	variable baselineFrac
 	variable tpLengthPoints
 	variable readTimeStamp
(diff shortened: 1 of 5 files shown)


t-b:
@timjarsky I"ve added the reference test with data from a model cell.

timjarsky (approved):

Issue #953, opened 2021-05-12
Rethink or document labnotebook entry source type range conditions


t-b:
This was found with a buggy version of #943.

The following analysis function

```igorpro
Function Buggy(panelTitle, s)
	string panelTitle
	STRUCT AnalysisFunction_V3& s

	variable sweepNo

	if(s.eventType != POST_DAQ_EVENT)
		return 0
	endif

	WAVE numericalValues = GetLBNumericalValues(panelTitle)

	Make/FREE/D/N=(1, 1, LABNOTEBOOK_LAYER_COUNT) values = NaN
	values[][][INDEP_HEADSTAGE] = 1	

	Make/FREE/N=(3, 1)/T keys = ""

	keys[0][0] =  "blah 1"
	sweepNo = s.sweepNo - 1
	ED_AddEntriesToLabnotebook(values, keys, sweepNo, panelTitle, UNKNOWN_MODE)

	WAVE/z junk = GetLastSetting(numericalValues, sweepNo, "ADC", DATA_ACQUISITION_MODE)
	print junk

	keys[0][0] =  "blah 2"
	sweepNo = s.sweepNo - 1
	ED_AddEntriesToLabnotebook(values, keys, sweepNo, panelTitle, DATA_ACQUISITION_MODE)

	WAVE/z junk = GetLastSetting(numericalValues, sweepNo, "ADC", DATA_ACQUISITION_MODE)
	print junk

	return 0
End
```

outputs

```
  '_free_'[0]= {0,NaN,NaN,NaN,NaN,NaN,NaN,NaN,NaN}
  NULL wave
```

This means that the second `GetLastSetting` fails to find the "DAC" entry.

The reason for this behaviour can be found in b07dd1b5 (FindRange: Fix an edge case with sweep rollback and the entry source types, 2017-05-12) which explains that we search for a consecutive block backwards in the sweep range which has the given entry source type.

So when the LBN looks like

![grafik](https://user-images.githubusercontent.com/1185677/118001813-5d0b3e80-b347-11eb-8da4-0b0f381d5db6.png)

we can't find entries with entry source type 0 (DATA_ACQUISITION_MODE) anymore.

t-b:
Some more explanation
---------------------

Our labnotebook stores entries per sweep. For distinguishing entries which exist for DAQ and TP we have introduced the entry source type in 1b21f99c (Bugfix: Disambiguate labnotebook entry search, 2016-10-08).

[...]

PR #1457, opened 2022-08-09
Track analysis function calls

### LINKED ISSUE #1081: Store number of analysis function calls per event in labnotebook
For post-mortem debugging it is interesting to see which analysis functions were called for which events.

- New keys in `GetSweepSettingsWave()`:  `Analysis Event [...]` (all events mentioned in EVENT_TYPE_ANALYSIS_FUNCTIONS)
- Move clearing the sweep settings wave from DC_Configure to DAP_CheckSettings before the AS_PRE_DAQ state transition.
- Increment the call count in AFM_CallAnalysisFunctions

t-b:
(Recreated from #1203)

- [ ] Finalize
- [ ] Commit cleanup
- [ ] Tests: Ditch TrackAnalysisFunctionCalls for V3 and translate the tests to GetAnalysisFunctionCallCountsFromLBN
- [ ] One entry per event in LBN with `count`

Revise after chat:

~~- [ ] Change UNKNOWN_MODE for writing to 2, introduce new variable for "not-caring-about-entry-source" on reading, adapt cache for these changes~~
~~- [ ] Use this new entry source type for call count to avoid LBN querying issues with UNKNOWN_MODE entries being between DATA_ACQUISITION_MODE entries.~~
~~- [ ] The code in 96519036de080d84e78c6f310bcee70e3c25cadc seems odd. With calling DC_ClearWaves from AS_PRE_SWEEP_CONFIG we are now actually calling it later than before.~~

Close #1081.

### DIFF
--- a/Packages/MIES/MIES_AcquisitionStateHandling.ipf
+++ b/Packages/MIES/MIES_AcquisitionStateHandling.ipf
@@ -70,6 +70,8 @@ Function AS_HandlePossibleTransition(string device, variable newAcqState, [varia
 
 	if(oldAcqState == AS_PRE_SWEEP && newAcqState == AS_MID_SWEEP)
 		ED_MarkSweepStart(device)
+	elseif(newAcqState == AS_PRE_SWEEP_CONFIG)
+		DC_ClearWaves(device)
 	endif
 
 #ifdef AUTOMATED_TESTING
--- a/Packages/MIES/MIES_AcquisitionStateHandling.ipf
+++ b/Packages/MIES/MIES_AcquisitionStateHandling.ipf
@@ -74,6 +74,14 @@ Function AS_HandlePossibleTransition(string device, variable newAcqState, [varia
 		DC_ClearWaves(device)
 	endif
 
+	if(newAcqState == AS_PRE_DAQ)
+		AFM_ClearCallCount(device)
+	elseif(oldAcqState == AS_ITI && newAcqState == AS_PRE_SWEEP_CONFIG \
+		   || oldAcqState == AS_POST_DAQ && newAcqState == AS_INACTIVE)
+		ED_WriteAnalysisFunctionCallCount(device)
+		AFM_ClearCallCount(device)
+	endif
+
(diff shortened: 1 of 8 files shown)


t-b:
I've pushed the latest version. But I don't think this is going anywhere sensible in the near future.

As already mentioned in the above todos we have to change the way the labnotebook querying works as to not bury other labnotebook entries.

In

image

the lower zero block for sweep 19 is not connected to the upper zero block as there is one entry source type with NaN (from an analysis function) in between.

See #953 why it is like it is.

Now we could make NaN entry source type to act like glue which connects blocks of the same entry source type. I can't think of an obvious way why this could break existing labnotebooks, but of course we have millions of possibilites of things going wrong.

I would be more in favour of doing if we would *completely* kill sweep rollback together with that change, as that makes reasoning about it easier.

timjarsky:
> I've pushed the latest version. But I don't think this is going anywhere sensible in the near future.
> 
> As already mentioned in the above todos we have to change the way the labnotebook querying works as to not bury other labnotebook entries.
> 
> In
> 
> image
> 
> the lower zero block for sweep 19 is not connected to the upper zero block as there is one entry source type with NaN (from an analysis function) in between.
> 
> See #953 why it is like it is.
> 
> Now we could make NaN entry source type to act like glue which connects blocks of the same entry source type. I can't think of an obvious way why this could break existing labnotebooks, but of course we have millions of possibilites of things going wrong.
> 
> I would be more in favour of doing if we would _completely_ kill sweep rollback together with that change, as that makes reasoning about it easier.

@t-b Instead of multiple entries, can a single string with all the analysis function call meta data be written with each sweep?

t-b:
> @t-b Instead of multiple entries, can a single string with all the analysis function call meta data be written with each sweep?

That does not help I'm afraid. A single entry at an inopportune time could break things.

t-b:
I've added some more info in https://github.com/AllenInstitute/MIES/issues/953#issuecomment-1217958888. We should discuss that on the next chat.

PR #1469, opened 2022-08-18
tools/initial-repo-config.sh: Make it work when called from subdirectories

t-b:
Will merge once CI passes.


t-b:
@MichaelHuth FYI.

PR #1483, opened 2022-09-08
Update IUTF to revised version, test with MIES test run

MichaelHuth:
Do not merge.

This PR is for testing if the IUTF revision runs (still) all MIES tests properly, and maybe even a bit faster.


t-b:
Same result as main. Thanks for checking.

t-b:
Execution time for the NI hardware tests is here and in another PR both 86min. RunAllTests.pxp is 2min.

Issue #1484, opened 2022-09-15
60 hz quantification


timjarsky:
Create a formula to quantify 60hz noise in data. Probably something like FFT to generate power spectrum, return amp @ 60hz expressed relative to baseline noise. Unit could be dB or %.

timjarsky:
I'm uncertain how best to determine the baseline noise region in the power spectrum.

t-b:
Use cases:
- Look at power spectrum
- Analyze features from power spectrum, look at multiple sweeps and compare the noise levels (the multiple sweeps should not be averaged)

Sample data: Any current clamp/voltage clamp sweep with baseline

FFT should use padding for speed reasons.

Start with plain power spectrum and see if that is enough already.

Operation:
- `powerspectrum(data(...))`
- choosable y units: %, db, real units via inverse fft (??!)

t-b:
- Reuse `DoFFT`

Issue #332, opened 2019-10-22
SweepFormula: Add a way to access the average wave


t-b:
We could allow `average()` instead of `sweeps()` in data.

t-b:
Not planned for now.

PR #1493, opened 2022-09-30
SF: Add powerspectrum operation

MichaelHuth:
Goal: 60 Hz noise quantification.

close https://github.com/AllenInstitute/MIES/issues/1484


timjarsky:
@MichaelHuth @t-b I gave `powerspectrum` a try. It works nicely. Thank you.

Would it be possible to add an optional parameter that returns the average power spectrums from all sweeps defined in `data`? 

I believe there is an open sweep formula issue for averaging sweeps. Maybe it's this one: https://github.com/AllenInstitute/MIES/issues/332, but if I understand how it might work, the sweeps would be averaged together before taking the power spectrum. In this case, I want to average the power spectrums. 

A second optional parameter I think would be helpful would be to zero the data before taking the power spectrum.

t-b:
Chat with TJ:
- [ ] Investigate if zeroing the input before FFT makes sense
- [ ] Add optional parameter to average different sweeps (but of course *not* channels)

MichaelHuth:
- optional parameter through string "avg" and "noavg".
- if source data is not from data() then when "avg" is set, average over all
- design averaging over channels as subroutine

timjarsky:
> * design averaging over channels as subroutine

I believe this should be averaging over sweeps, not channels

t-b:
> > * design averaging over channels as subroutine
> 
> I believe this should be averaging over sweeps, not channels

This seems to be a non-native english problwem. We, Michael and me, chatted about that today and we will *not* mix data from different channels.

timjarsky:
> > > * design averaging over channels as subroutine
> > 
> > 
> > I believe this should be averaging over sweeps, not channels
> 
> This seems to be a non-native english problem. We, Michael and me, chatted about that today and we will _not_ mix data from different channels.

@t-b That makes sense. Thanks!

MichaelHuth:
@timjarsky Here is another revision of the powerspectrum operation for feedback.

The syntax is like
```
powerspectrum(
data(cursors(A,B),select(channels(AD),sweeps(),all)),
avg,
zero,
Hanning,
1000)

```
Argument 0: input data, e.g. results from data operation

optional Argument 1: "avg" or "noavg", default "noavg" enables averaging over sweeps

optional Argument 2: "zero" or "nozero", default "nozero" enables zeroing of sweep data

optional Argument 3: Window function for FFT or "none", default "Hanning"
valid are the Igor Pro known window functions, one of
_Bartlet;Bartlett;Blackman367;Blackman361;Blackman492;Blackman474;Cos1;Cos2;Cos3;Cos4;Hamming;Hanning;KaiserBessel20;KaiserBessel25;KaiserBessel30;Parzen;Poisson2;Poisson3;Poisson4;Riemann;SFT3F;SFT3M;FTNI;SFT4F;SFT5F;SFT4M;FTHP;HFT70;FTSRS;SFT5M;HFT90D;HFT95;HFT116D;HFT144D;HFT169D;HFT196D;HFT223D;HFT248D;_

optional Argument 4: max frequency on the x-axis, default: 1000 (1kHz)
The max frequency is capped at the max reasonable frequency returned from the FFT, typically ~125 kHz, probably lower if you use more channels or ITC.

The y-axis shows dB now.

When over sweeps is averaged the legend shows "averaged" as sweep number.

I would like to note here, that for the default TP powerspectrum no window function is used, that produces strong artifacts.
You can compare with "none" as window function.

timjarsky:
@MichaelHuth @t-b, thank you for the updates. Overall it's working well and is already useful.

I couldn't successfully define a non-default max frequency on the x-axis. The formula below returned a blank plot. Maybe I don't have the syntax quite right?

```powerspectrum(
data(
E0,
select(channels(AD0),sweeps(),all, ic)
),avg,Hanning,500
)
```

Y-axis scaling: rather than dB, it would more intuitive to have the unit on the y-axis be something that relates more directly to the measured unit (mV or pA). What do you think about scaling the power spectrum so that the area under the curve is equal to the variance of the signal (variance is the standard deviation squared)? Could we support multiple Y-scalings (i.e, keep dB)?

My following comment is a feature request related to an everyday use case. Users will often want to quantify the strength of power line noise (60Hz here, 50Hz Europe). Is there a way we could express the ratio of the powerline noise to the noise at neighboring frequencies (e.g., +/- 10 hz?). So a ratio of 1 would indicate the power line band isn't contributing to the frequency content.  ratios>1 would show that the power line frequencies contaminate the signal.

MichaelHuth:
@timjarsky Regarding the formula: The "zero" argument is missing after the "avg" argument.

MichaelHuth:
@timjarsky We can add another argument to choose how the result should be calculated to target specific y-axes. Though, we should not have too many arguments. So e.g. can we drop some argument and use a default value instead. I could imagine that always zeroing is a possible option as it reduces the peak at 0 Hz from a constant background. (This assumes that the theoretical baseline is at zero)

If we don't calculate dB the base units for the y-axis would be pA^2 or mV^2.

Regarding the scaling for the y-axis such that area(powerspectrum) == var(powerspectrum) I suppose you refer to Parseval's theorem to normalize the spectrum by its total energy?

For the 50 Hz / 60 Hz analysis we could add another argument. However, it should be defined how the levels are determined. e.g. fitting a gauss with a fixed peak location at 60 Hz and take the ratio of peak amplitude to amplitude at peak + 10 Hz? (in the linear term only offset, no slope)

timjarsky:
> @timjarsky We can add another argument to choose how the result should be calculated to target specific y-axes. Though, we should not have too many arguments. So e.g. can we drop some argument and use a default value instead. I could imagine that always zeroing is a possible option as it reduces the peak at 0 Hz from a constant background. (This assumes that the theoretical baseline is at zero)

Yes, I agree always zeroing makes sense.

> 
> If we don't calculate dB the base units for the y-axis would be pA^2 or mV^2.

Okay
> 

> Regarding the scaling for the y-axis such that area(powerspectrum) == var(powerspectrum) I suppose you refer to Parseval's theorem to normalize the spectrum by its total energy?

That's correct. 

> 
> For the 50 Hz / 60 Hz analysis we could add another argument. However, it should be defined how the levels are determined. e.g. fitting a gauss with a fixed peak location at 60 Hz and take the ratio of peak amplitude to amplitude at peak + 10 Hz? (in the linear term only offset, no slope)

I think my pref for the reference/control frequency would be to take the average of the peak at bracketing frequencies. E.g, 50 and 70 Hz for 60 Hz line. Fitting with a gauss sounds optimal, but I worry about fit failures. A simpler method might be to take the y-value at the relevant frequencies.

MichaelHuth:
I have made another revision. The syntax is adapted to
`powerspectrum(array data[, string unit[, string average[, variable ratioFrequency[, variable cutOffFrequency[, string windowFunction]]]]])`

it would be interesting to see how well the quantification works, e.g.
`powerspectrum(data(cursors(A,B),select(channels(AD),sweeps(),all)),dB,avg,60)` for 60 Hz.

I have added some more documentation and also updated the SF help notebook with details.
Note that the ratio is determined on the basis of the set y-axis units.

timjarsky:
@MichaelHuth I tried the quantification on data that doesn't have line noise contamination, and I get; ```error: singular matrix or other numeric error```  when debugging is enabled.

If I push through or disable the debugger, the value returned is 1;

image

Any of the data I've previously shared from IVSCC should allow you to test it. The naming format for IVSCC data is mouse line followed by the date. Here is an example
![image](https://user-images.githubusercontent.com/5693697/194586044-2020a31f-5b58-439a-8082-e1d87814ad78.png

MichaelHuth:
After a chat with Thomas, we have decided to return 0 if the fit failed. We also changed the fit to be a combination of a linear and a gaussian term. This gives more accurate results.

When debugging is enabled through the debug panel then the fitted curve is plotted as well.
The fitted curve is plotted if the fit was successful. The ratio result can still be 0 as the other conditions as mentioned in the docu apply for a successful ratio result. For successful ratio results the fit results are printed to the history.
Look for a line starting with `PS ratio, peak position, baseline, peak amplitude`.

timjarsky:
@MichaelHuth have all the changes been pushed? I'm not seeing the history content or the curve plotting with debugging enabled. 

image
image

t-b:
@timjarsky "Debugging enabled" here means the debug panel and MIES_SweepFormula.ipf checked.

timjarsky:
@t-b gotcha. Thanks. It's working nicely!

MichaelHuth:
I have added a small fix for sweeps that contain NaNs at the end. These are not allowed for the averaging and FFT, so I cut them off.

timjarsky (dismissed):

t-b:
Review:

ab2dac93 (Split DoFFT function and rename it to DoPowerSpectrum, 2022-10-05)

- [ ] DoPowerSpectrum and DOFFTImpl should be moved to MIES_Utilities.ipf

176f72b8 (SF: Add powerspectrum operation, 2022-09-30)

- [ ] You are using only

```
#ifdef DEBUGGING_ENABELD
...
```

but that should be 

```
#ifdef DEBUGGING_ENABLED
	if(DP_DebuggingEnabledForCaller())
  ...
```

as we are using file specific debugging.

- [ ] Please add a short comment to explain the fit constraints.

- [ ] SF_AverageDataOverSweeps is still using isAssociated naming although we
      wanted to reserve that for associated/unassociated channels which are not
      attached to a headstage

- [ ] Missing tests

2f9d3e24 (SF: Adapt plotter legend for averaged sweeps, 2022-10-05)
8061788e (SF: Add Tracecolor in plotter for sweep averaged data, 2022-10-05)

Looks good.

59c8a87f (SF: Add documentation for power spectrum operation, 2022-10-07)
28b2d7ce (SF: Add docu for operation powerspectrum to SF help notebook, 2022-10-07)

Nice!

t-b (changes requested):
See comment

t-b:
LGTM.

t-b (approved):

Issue #1498, opened 2022-10-03
Fix failing bamboo job with ITC hardware


t-b:
Since around February 2022 the ITC tests are mostly failing in CI. See http://bamboo.corp.alleninstitute.org/reports/generateReport.action?reportKey=com.atlassian.bamboo.plugin.system.reports%3AratioOfSuccess&selectFields=reportKey&labelTarget=&buildIds=153157784&selectFields=buildIds&groupByPeriod=AUTO&selectFields=groupByPeriod&dateFilter=RANGE&selectFields=dateFilter&dateFrom=01%2F01%2F2020&dateTo=31%2F12%2F2022&save=Submit&atl_token=20bfbb9788561001a2d14b27b31b40bc663b6fc7 (requires VPN access)

Graph:

![grafik](https://user-images.githubusercontent.com/1185677/193652451-c61ed2a6-967e-4c75-b2f4-5f8d67118614.png)

We would like to get that fixed as it is bad.

- We do get quite often a crash with ITC hardware in CI. We also have the PDB files on the CI server. This is tracked at WaveMetrics as issue 4168. The stacktrace looks like:

```
2022-05-02 09:30:38: BUG: Unexpected magic value in IgorStringClass::v().
Stack Trace:
	QApplicationPrivate::notify_helper
	QObject::event
	HandleXOPWinMessage
	HandleXOPWinMessage
	SetScrapData
	SetScrapData
	BlockMove
	BlockMove
	BlockMove
	BlockMove
	SetScrapData
	SetScrapData
	SetScrapData
	SetScrapData
	SetScrapData
	SetScrapData
	BlockMove
	BlockMove
	BlockMove
	BlockMove
	SetScrapData
	HandleXOPWinMessage
	HandleXOPWinMessage
```

Which does not make any sense at all as none of our XOPs use `SetScrapData`. This function does also not call itself.

We don't get these crashes with NI hardware.

- The timepoint in spring 2022 where the number of sucessfull builds drops does not coincide with any big things like IP8->IP9
- The ITC XOP was last updated in d1830b66 (XOPS-IP9-64bit/itcXOP2-64.xop: Update it, 2022-05-26).

t-b:
I ran HardwareBasic.pxp overnight with ITC hardware. It crashed in ITCstartAcq2 via

![grafik](https://user-images.githubusercontent.com/1185677/193804281-0a8c6d46-cbce-4bfe-a125-87bbb33689a6.png)

[...]

PR #1504, opened 2022-10-05
Hardware Tests: Stop DAQ/TP after each test case

t-b:
This should prevent some crashes when DAQ does not finish properly and the
UTF background monitor timeout kicks in and forcefully stops DAQ/TP.

Found in a crash dump debug session with @MichaelHuth, see also https://github.com/AllenInstitute/MIES/issues/1498#issuecomment-1268741641.


Packages/tests/UTF_HardwareHelperFunctions.ipf
```diff
@@ -117,6 +117,14 @@ Function TEST_CASE_END_OVERRIDE(name)
 	string dev, experimentNWBFile, baseFolder, nwbFile
 	variable numEntries, i, fileID, nwbVersion, expensiveChecks
 
+	// be sure that DAQ/TP is stopped before we do anything else
+#ifndef TESTS_WITH_NI_HARDWARE
```
timjarsky, line 121: @t-b so if it is **not** TESTS_WITH_NI_HARDWARE, we stop DAQ/TP when the testing timer times out?

Also, the testing timer is in Bamboo, not the .pxp test file, is that correct?

timjarsky (approved):

Packages/tests/UTF_HardwareHelperFunctions.ipf
t-b: This is not the bamboo timer. This is the testing framework timeout for the reentry function. See also https://docs.byte-physics.de/igor-unit-testing-framework/advanced.html#test-cases-with-background-activity.

The hardware test cases are always in two parts:
- setup, starting DAQ/TP
- checking results 

The "checking results" step is in the functions name `XXX_REENTRY`. This part is called once DAQ/TP is finished *or* when the user settable timeout hits. We have set this timeout to 10m. Once that timeout is hit the `XXX_REENTRY` function is called and *then* we call TEST_CASE_END_OVERRIDE. And at this point DAQ/TP could still be ongoing, so we have to stop it before we do anything else.

t-b: Yes this part is only executed for ITC hardware.

PR #1521, opened 2022-10-21
SetControlInEvent: Fix documentation and add check function

t-b:
I've tweaked the documentation, fixed the error message wording and also added a check function so that it complains already in the wave builder. This required a bit of a general change as we now also pass parameters not returned in the `_GetParams` function to the check function.


timjarsky (dismissed):

t-b:
Will merge once CI passes.

PR #1529, opened 2022-10-28
Add analysis parameters table

### LINKED ISSUE #1350: Add table with all analysis parmeters
- [x] Include analysis function
- [x] Optionality (after #1345)
- [x] Type

t-b:
The table is saved in the repo but we have a check so that it does not get stale.

It looks like

grafik

The non-abbreviated names are clickable.

- [x] Add legend for analysis function abbreviations
- [x] Add help as last column
- [x] Sort by name only

Close #1350.


t-b:
@timjarsky Updated.

timjarsky (dismissed):

Issue #1247, opened 2022-01-07
SweepFormula: Add new TP mode


t-b:
- Allow user to choose between single (`exp_XOffset`) and double exponential. See `TP_AutoFitBaseline` for how it calls CurveFit.
- Extend existing TP operation
- Works on a single TP at a time
- Fit range: Only the trailing baseline. If the next epoch after the TP has amplitude zero enlarge the fit range by at most the trailing baseline stolen from the next epoch, cropped to the next epoch length of course.
- returns `tau [ms]` slow (aka large) by default, allow user to return the other one as well
- Allow returning amplitudes (using y units of the data) as well
- Same approach for IC/VC
- Visualize fit if debugging is enabled for MIES_SweepFormula.ipf

Use one of the IVSCC files as reference, e.g. Rbp4-Cre_KL100;Ai14-543835.09.09.01.pxp.

t-b:
Chat with Tim to clarify objectives and more fine-grained specs.

t-b:
I've chatted with Tim about the question how to integrate the new mode.

And we think

```
tp(fit(...), [array selectData[, array ignoreTPs]])
``

is the way to go. In this case we will also have the operations base(), ss(), inst(). The help for those operations should be part of `tp` and the topics should only link there. The current numeric constants and strings for `mode` must also be removed so that we don't create a compatibility mess.

MichaelHuth:
@timjarsky Some questions came up regarding the fitting:

- What is exactly meant by "the next epoch after the TP"? There are several candidates here:
    - the epoch that follows TP_B1 (aka trailing baseline) on the same tree level or if not present on a lower tree level
    -   the epoch that follows TP (aka test pulse) on the same tree level or if not present on a lower tree level

- to determine if TP_B1 is shortened, is it correct to use TP_B0 as reference interval?

- if TP_B1 is shortened and I extend the range within the following epoch then it is not relevant if the following epoch has a non-zero amplitude within the range?

- Should the x0 constant for the exp_XOffset and dblexp_XOffset kept at the default value (left edge of fit range)

- When dblexp_XOffset if fitted and amplitude should be returned, which amplitude should it be? (e.g. the higher amplitude of both?)

- TP_AutoFitBaseline has a criteria where the residuals are evaluated and a fit quality threshold is applied. Should such criteria be also applied here? (the same threshold?)

timjarsky:
> @timjarsky Some questions came up regarding the fitting:
> 
[...]

PR #1542, opened 2022-11-04
SF: Adapt tp operation for new fit mode

### LINKED ISSUE #1155: Auto TP: Make baseline fitting routine available as sweep formula

MichaelHuth:
close https://github.com/AllenInstitute/MIES/issues/1247
close #1155

### DIFF
--- a/Packages/MIES/MIES_Constants.ipf
+++ b/Packages/MIES/MIES_Constants.ipf
@@ -1925,6 +1925,10 @@ StrConstant SF_DATATYPE_MIN = "Min"
 StrConstant SF_DATATYPE_RANGE = "Range"
 StrConstant SF_DATATYPE_EPOCHS = "Epochs"
 StrConstant SF_DATATYPE_TP = "TestPulse"
+StrConstant SF_DATATYPE_TPSS = "TestPulseMode_SteadyState"
+StrConstant SF_DATATYPE_TPINST = "TestPulseMode_Instantaneous"
+StrConstant SF_DATATYPE_TPBASE = "TestPulseMode_Baseline"
+StrConstant SF_DATATYPE_TPFIT = "TestPulseMode_Fit"
 StrConstant SF_DATATYPE_POWERSPECTRUM = "Powerspectrum"
 /// @}
 
--- a/Packages/MIES/MIES_Epochs.ipf
+++ b/Packages/MIES/MIES_Epochs.ipf
@@ -47,6 +47,8 @@ static StrConstant EPOCH_SN_TRIG_CYCLE_INCOMPLETE = "I"
 static StrConstant EPOCH_SN_TRIG_HALF_CYCLE = "H"
 static StrConstant EPOCH_SN_UNACQUIRED = "UA"
 
+static Constant EPOCH_GAPS_WORKAROUND = 0
+
 /// @brief Clear the list of epochs
 Function EP_ClearEpochs(string device)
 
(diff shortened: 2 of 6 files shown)


MichaelHuth:
@timjarsky This is in a state for initial testing and tryout. The new command is:
`tp(mode, ...)` // the first argument changed

for mode you have to use one of the following operations:
`tpbase()`
`tpss()`
`tpinst()`
`tpfit(exp,amp,250)`

where `exp` is for exp_Xoffset and `doubleexp` would be for dblexp_XOffset fit function.
The second argument is for the return value displayed, options are:

`tau` // if doubleexp max tau, otherwise identical to tausmall

`tausmall` // if doubleexp min tau, otherwise identical to tau

`amp` // if doubleexp then amplitude of maximum abs amplitude

`minabsamp` // if doubleexp then amplitude of minimum abs amplitude

`fitq` // sum(residuals^2) / numXPoints

I have not tested User TPs yet.
Display of full fit result in debug mode is not implemented yet.

timjarsky:
@MichaelHuth I get a compile error; expected assignment operator
```
				WAVE/Z/T nextEpoch = EP_GetNextEpoch(numericalValues, textualValues, sweepNo, XOP_CHANNEL_TYPE_DAC, dacChannelNr, epBaselineTrail, 1)
```

MichaelHuth:
@timjarsky I forgot to push a part of the changes epochs code. Sorry for that. It should work now.

Also special output is in when debugmode for sweepformula is enabled. In Debugmode I recommend not so select too many sweeps, otherwise the output can get a bit overwhelming.

timjarsky:
Hi @MichaelHuth 

In debug mode, the fit never extends past the end of TP_B1 (see below). Does the end of the fit trace indicate that the baseline epoch following TP_B1 is not being used for the fit? 

image
image

MichaelHuth:
@timjarsky Could you upload your example case as experiment file? The start of the following epoch at 78.005 is a bit suspicious.

timjarsky:
[TPfittesting.zip](https://github.com/AllenInstitute/MIES/files/10016202/TPfittesting.zip)
zipped .pxp file. Data was generated from the PR @MichaelHuth

MichaelHuth:
Diagnosis:

The issue is that we calculate the Length the TP uses at two locations differently. This breaks the consecutive epochs.
For the TP wave the length is retrieved from `TP_UpdateTPSettingsCalculated` with
`calculated[%pulseLengthPointsDAQ] = trunc(TPSettings[%durationMS][INDEP_HEADSTAGE] / (DAP_GetSampInt(device, DATA_ACQUISITION_MODE) * MICRO_TO_MILLI))`

The other location where this plays a role is if `Inserted TP` is enabled in the DA_Ephys panel. Then the ms are calculated with
`TP_CalculateTestPulseLength` and written to the GUI control `Onset delay (ms)` and reduced to 5 digits after the . because `%g` is used. This value is then used to calculate the effective onset delay in `DC_ReturnTotalLengthIncrease` with
`onsetDelayAutoVal      = round(GetValDisplayAsNum(device, "valdisp_DataAcq_OnsetDelayAuto") / (samplingInterval * MICRO_TO_MILLI))`

The result is that the onset delay that defines the start of the stimset is not the same as the end of the inserted TP. This gives a few missing ms in the epochs and the stimset epoch does not start when the TP epoch ends - thus it is not the next epoch.

t-b:
@timjarsky I'm chatting with Michael about a fix. I'll do the fix so that in the future we have no gaps in the epochs again.

I presume you also want to analyse existing data with gaps?

timjarsky:
> @timjarsky I'm chatting with Michael about a fix. I'll do the fix so that in the future we have no gaps in the epochs again.
> 
> I presume you also want to analyse existing data with gaps?

@t-b I'm not sure. For now, I might prefer to make an issue for handling gaps rather than solve it now.

t-b:
>  I'm not sure. For now, I might prefer to make an issue for handling gaps rather than solve it.

It's relatively easy to ignore the gaps, so how about now?

timjarsky:
@t-b sure

t-b:
https://github.com/AllenInstitute/MIES/pull/1554 has a fix for the epoch "gap".

MichaelHuth:
I added a workaround for the gap that can be enabled by setting the constant EPOCH_GAPS_WORKAROUND in MIES_Epochs.ipf to 1.

I also added tests.

t-b:
@MichaelHuth Can you fix CI? Thanks. Linting is `tools/check-code.sh`.

timjarsky:
@MichaelHuth @t-b 
(1)
I expect combining avg with tp to return a single average value, that's the average of each tp formula return for each sweep.

Instead, it returns the same values as TP formula alone but plotted vs. sweeps.

image
image

(2)
In voltage clamp, the following error occurs sometimes:
image
Here is the data file
[QN22.26.030.14.07B.03.zip](https://github.com/AllenInstitute/MIES/files/10149336/QN22.26.030.14.07B.03.zip)

(3)
Since the user can set the time for the exp fit, should we remove the 500 ms hard cap? e.g., tpfit(exp,amp,1000)

MichaelHuth:
> @MichaelHuth @t-b (1) I expect combining avg with tp to return a single average value, that's the average of each tp formula return for each sweep.
> 
> Instead, it returns the same values as TP formula alone but plotted vs. sweeps.

The issue here is the following:

The operation avg average the data within a sweep. So if you do an avg(data(...)) then it averages the points from each sweep independently. In developer speech data returns multiple independent result waves and avg is applied to each one.

Now tp returns multple data waves as well, for each tp times sweep. The result with exp is a single point and thus, avg averages over each of these single point waves independently. But you want now something else, you want to average over the data of multiple result waves. This would need a different average operation that works in the other direction.

Regarding the different plotting: tp suggests to the plotter to plot against the tp number on the x-axis. After avg is applied this suggestion is cleared, as it would in the general case make no sense any more. Thus, the default kicks in for data that was somehow sources by sweeps that it is plotted agains the sweep number.

MichaelHuth:
@timjarsky regarding 2:

TP operates this way "If a single sweep contains multiple test pulses then the data from the test
pulses is averaged before analysis. The test pulses in one sweep must have the same duration." (copied from docu)
So the issue here is that there are 4 TPs in the sweep and the first tp epoch is 100 ms, whereas the others are 60 ms long.
The operation does not average TPs with different length as this will likely give no senseful result.
If you adapt the expression to
```
TP(
tpss(),
select(channels(AD),sweeps()),0
)
```
to exclude the first TP it will give a result.

timjarsky:
> 
> The issue here is the following:
> 
> The operation avg average the data within a sweep. So if you do an avg(data(...)) then it averages the points from each sweep independently. In developer speech data returns multiple independent result waves and avg is applied to each one.

Hi @MichaelHuth Thomas reminded me we already have an issue for this #1552.

timjarsky (approved):

t-b:
Review:

Please rebase against latest main so that NI passes.

52e8875d (SF: Add operation tpss, tpinst, tpbase, tpfit for setup of tp mode, 2022-11-04)

Good!

eb524741 (SF: Add new tpfit mode, 2022-11-07)

Unneeded ` + 1` for EnsureLargeEnoughWave. This seems to be a persisting issue. Does the documentation need improving?

Wrong indentation:
```
if(debugMode)
	return output
endif
```

Unrelated changes are part of the commit! All the epoch additions should be in its own commit. Tests are missing for EP_GetNextEpoch. And that function also does not handle multiple next epochs which can happen with user epochs, I'm fine with asserting in this case if you think this is the easiest approach.

b829e470 (SF: Change TP annotation to be aware of invalid sweep numbers, 2022-11-10)

The commit message should explain why this change is needed. How can that bug happen in real life?

ccc45138 (SF: Add tests for tpss, tpbase, tpinst, tpfit, 2022-11-11)

Good.

3898d0e0 (SF: tp operation tpfit add fit length information to output in testing mode, 2022-11-14)

Please add newlines around `#ifdef /#endif` blocks as these are a new logical unit per our coding guidelines.

7189a16d (SF: Adapt tests for tp operation with new tp mode parameter, 2022-11-14)
bc471524 (SF: Add tests for tp operation with tpfit mode, 2022-11-14)
ad6c05ed (EP: Add workaround for epoch gaps for EP_GetNextEpoch, 2022-11-18)

The commit message leaves the reader clueless why gaps can happen. As the MR in question is not yet merged, I think it is in this case useful to refer to https://github.com/AllenInstitute/MIES/pull/1554.

EP_GetNextEpoch: Is the new optional parameter `checkGaps` named correctly? For me this looks more like `ignoreGaps`.

I think `EP_GetGaps` can be made static.

18c22363 (EP Tests: Add Tests for EP_GetNextEpoch, 2022-11-18)

Tests are good, commit splitting not, see the comment for eb524741.

t-b (changes requested):
See comment

t-b (approved):

PR #1543, opened 2022-11-04
Bugfix/1543 fix analysisbrowser loading sweeps from igor

MichaelHuth:
In https://github.com/AllenInstitute/MIES/commit/97b125244d10733d7da77e733b15eed2a420f87b it was fixed that AB_LoadDataWrapper changed the current DF. However the caller must use then the known DF to access the loaded data.
That was not done correctly in AB_LoadSweepFromIgor and is fixed now.

Also other locations where the callee saved the current DF for the AB_LoadDataWrapper have the unnecessary DF saving/restore removed.

- [x] Add test using PGC


Packages/tests/Basic/UTF_AnalysisBrowserTest.ipf
```diff
@@ -0,0 +1,30 @@
...
+
+	win = AB_OpenAnalysisBrowser()
+	PathInfo home
+	REQUIRE_EQUAL_VAR(V_flag, 1)
+	PGC_SetAndActivateControl(win, "setvar_baseFolder", str=S_path)
+	PGC_SetAndActivateControl(win, "button_base_folder_scan")
+	WAVE expBrowserSel = GetExperimentBrowserGUISel()
+	expBrowserSel[0][0][0] = 81
+	PGC_SetAndActivateControl(win, "button_load_sweeps")
+	win = SWEEPBROWSER_WINDOW_TITLE + "1"
```
t-b, line 26: The constant SWEEPBROWSER_WINDOW_TITLE does not exist anymore in latest main. It is now called SWEEPBROWSER_WINDOW_NAME. But I would actually get the list of graphs using WinList, check that it contains only one and use that.

t-b (approved):

PR #1544, opened 2022-11-04
bugfix: When upgrading the LBN the DimLabels were converted too late

MichaelHuth:
With IP9 the characters allowed for Dimlabels are stricter. For this in 487be1f1f02d390abdf389c78ce4b12761366801
a upconversion was introduced. However, it happened too late in the LBN upgrade chain, such that old DimLabels generated an RTE.

The upconversion was moved to an earlier location in the LBN upgrade process.


t-b:
Forgot about that. Thanks!

t-b (approved):

PR #1535, opened 2022-10-29
Fix hooks functions for SweepBrowser

t-b:
This fixes the problem with the missing "restore" button once the settings history was closed.

It also fixes the problem that the really helpful SF hooks did not work with the SweepBrowser.

I also cleaned the code up a bit.

PR #1546, opened 2022-11-06
Make browser from SF plot retrievable

### LINKED ISSUE #1526: Retrieve data browser from SF plot
With multiple data browsers open, it can be challenging to keep track of which data browser is linked with the sweep formula plot window.

Options: 
1) add a button to the SF plot to bring the DB to the front
2) link SF plot to DB window - this would need to be tied to buttons that allow the user to shrink the various panels of the DB

t-b:
- [x] Merge only after #1535

Close #1526.


timjarsky (approved):

t-b:
Will merge once CI passes.

PR #1554, opened 2022-11-17
Make the inserted TP length calculation consistent

t-b:
- [x] Add test differing dDAQ delay for epoch info
- [x] Nicify commits
- [x] fix onset delay calculation for GUi
- [x] Fix GetTotalOnsetDelayFromDevice
- [x] Investigate test failures


t-b:
@MichaelHuth Please have a look.

MichaelHuth (changes requested):
It is working and code looks good to me.

Still, there is one thing that should be unified:
We have a test pulse pulse length calculation in https://github.com/AllenInstitute/MIES/blob/8c1a53b5c54656f1199cd2c739de8fd9eaf056df/Packages/MIES/MIES_TestPulse.ipf#L1499 for the creation and in
TP_CalculateTestPulseLength for the whole length and in TP_CalculateBaselineFraction for the BL fraction.

It would be good if this is more condensed in the code at one location or maybe even one function.

I am also fine to postpone that to an issue.

t-b:
@MichaelHuth THx. Done.

MichaelHuth (dismissed):
LGTM

t-b:
The failing NI test, SF_InsertedTPVersusTP, is fixed in #1559.

timjarsky (approved):

PR #1566, opened 2022-12-03
Revert ITCXOP

### LINKED ISSUE #1551: Make PR with old ITCXOP and see if it makes a difference

t-b:
Let's see if CI passes in this case better. It is reverted to 96465c53 (XOPs: Update ITC XOP, 2021-12-03).

Close #1551.


t-b:
Crashed, see http://bamboo.corp.alleninstitute.org/browse/MIES-MPNL50-TWIO-1.

PR #1568, opened 2022-12-04
Draft: Make async framework faster

t-b:
- [x] Wait for fixed IP version reported as issue 4305
- [ ] Finalize
- [ ] Ask MH for review

### DIFF
--- a/Packages/MIES/MIES_Async.ipf
+++ b/Packages/MIES/MIES_Async.ipf
@@ -103,7 +103,7 @@ End
 threadsafe static Function ASYNC_Thread()
 
 	for(;;)
-		DFREF dfr = ThreadGroupGetDFR(0, 10)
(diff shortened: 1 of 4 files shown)


t-b:
Closing for now.

t-b:
This still crashes with IP9:

```
IGORVERS:9.06
BUILD:56657
COMMIT:4a46fb8a46
```

t-b:
Try out with MoveDataFolder and the old approach. Use fixed name for subfolder.

t-b:
@MichaelHuth I can't get it working. Also MoveDataFolder crashes. I would give this up now, if you don't have a genius-style idea.

t-b:
I don't think using a global DF makes sense as creating and removing that requires the databrowser to update. And that is quite slow.

Issue #1577, opened 2022-12-06
Show trace info tags should be on by default


t-b:

t-b:
Still buggy in my and MH's machine.

PR #1578, opened 2022-12-06
ShowTraceInfoTags: Make it really work

t-b:
Turns out the function only suceeds if a graph window is on top. So let's just created one and dispose it afterwards.

Close: #1577


timjarsky:
@t-b, should this show tags in the analysis browser after loading sweeps or after selecting trace info tags from the graph menu?

t-b:
It should show tags/tooltips right away without you having to select Graph->Show Trace Info tags.

timjarsky (approved):

t-b:
I can't break it here.

PR #1584, opened 2022-12-08
SF: bugfix for epochs operation not returning name in latest version

MichaelHuth:
- With the introduction of shortnames for epochs the epochs operation returned the full tags information instead of the name. However according to the documentation it should only return the name.


t-b:
Code looks good, but can we have a test which checks that for epoch info without shortnames we do return the full tag?

t-b (changes requested):
See comment

t-b (approved):

PR #1594, opened 2022-12-14
bugfix for SetStringInWaveNote generating pending RTE

MichaelHuth:
- Retrieving numEntries by calling numpnts created a pending RTE for a null wave argument. The RTE was hidden in the tests because the function is declared threadsafe.
- solution: First check argument with ASSERT_TS before calling numpnts

since a1bc2f34b1e24417a7def6ae76d2f4b38b6a0f4a


t-b (approved):

t-b:
Nice one!

PR #1602, opened 2022-12-16
Nicify epoch tooltips

t-b:


timjarsky:
image

@t-b What do you think about putting the shortname at the top of the list and adding the time unit to the start and end times?

t-b:
@timjarsky Jeep, I can do that.

t-b:
Shortname is the top entry in the list.
If there is no shortname output `ShortName: NA`, otherwise just have the shortname in one line and make it bold.

timjarsky:
very nice;

timjarsky:
very nice;
image

PR #1604, opened 2022-12-19
PSQ_PipetteInBath/PSQ_AccessResistanceSmoke: Use new tp mode operations

t-b:
Forgotten in 08484992 (SF: Adapt tests for tp operation with new tp mode parameter, 2022-11-14).

@MichaelHuth Looks like https://github.com/AllenInstitute/MIES/pull/1542#issuecomment-1343431853 was not yet done. I've fixed the obvious test failures. But TestHelpNotebook still fails and the Notebook help is also outdated. Can you take over? Thanks.


t-b:
This was a misunderstanding.

MichaelHuth:
@t-b docu for new tp operation, tpbase, tpss, tpinst and tpfit added. Here the test against the help runs through.

t-b:
@MichaelHuth Thanks!

Issue #1119, opened 2021-10-25
New mode for apfrequency sweep formula function


t-b:
diff --git a/Packages/MIES/MIES_SweepFormula.ipf b/Packages/MIES/MIES_SweepFormula.ipf
index 9563ee95..aa09dd54 100644
--- a/Packages/MIES/MIES_SweepFormula.ipf
+++ b/Packages/MIES/MIES_SweepFormula.ipf
```diff
@@ -2108,6 +2108,23 @@ static Function/WAVE SF_OperationApFrequency(variable jsonId, string jsonPath, s
                case SF_APFREQUENCY_FULL:
                        Make/N=(numSets)/D/FREE outD = levelPerSet[p] / (DimDelta(data, ROWS) * DimSize(data, ROWS)) * 1e3
                        break
+               case SF_APFREQUENCY_INSTANTANEOUS_VS_TIME:
+                       // calculate SF_APFREQUENCY_INSTANTANEOUS
+                       // but remember x values of spike pairs
+                       //
+                       // Spikes
+                       // 1 2 3 4
+                       // -------
+                       //
+                       // 1-2: same x values from 1
+                       // 2-3: same x values from 2
+                       //
+                       // and then user wants to plot spikes vs x values
+                       // ...
+                       //
+                       // use two function modes for this
+                       // SF_APFREQUENCY_INSTANTANEOUS_PAIR vs SF_APFREQUENCY_INSTANTANEOUS_PAIR_TIMES
+                       break
                case SF_APFREQUENCY_INSTANTANEOUS:
                        Make/N=(numSets)/D/FREE outD
```

PR #1607, opened 2022-12-27
Extend apfrequency operation with method instantaneous pair

MichaelHuth:
close https://github.com/AllenInstitute/MIES/issues/1119

The new method number for apfrequency is 3.


MichaelHuth:
@timjarsky Before I add tests and docu, could you please verify that the calculation works as intended by the brief issue?
I added kHz as label is the input wave units are ms, though it's quite a stretch to define a frequency from two data points.

timjarsky:
@MichaelHuth The new method is looking good. Can you add the unit to the y-axis? I also need help understanding the x-axis unit. It should reflect the AP number in some way.
image

MichaelHuth:
@timjarsky I think the kHz should actually be on the y-axis. (I will fix this). About the x-axis in the issue is stated
> and then user wants to plot spikes vs x values
The x values are the time of the left spike of the pair in ms.
Does that make sense? From your comment it sound like x should just count the (interleaved) pairs.

timjarsky:
Hi @MichaelHuth,

Let's keep the x-axis as spike time but add an option to have the x-axis be spike count.

I'd also like to add an option to have the Y values be time or frequency (default to frequency).

One final option is to have the Y-values be normalized to the first interval time/frequency (default to non-normalized).

Thank you!

t-b:
@MichaelHuth It might make sense to have these new optional parameters for all methods.

MichaelHuth:
@timjarsky So the third option would actually be two options? 
1. divides all results by the first spike pair interval time
2. divide all results by the first pair frequency

timjarsky:
Yes, you can think of it that way. I was first defining whether the user wants to work with frequency or time, then letting the user decide if they want the data normalized or not.

MichaelHuth:
I added two more optional parameters:
time/freq~ and normalize/nonormalize~
~ = default

time/freq has only an effect for the result values on method 3 (pairs) as I currently don't see a senseful application on the other methods that return integrated values.

normalize works on all methods and requires at least 2 peaks per sweep to be applied. The normalization divisor depends on the time/freq setting.

I have fixed also the Y-axis label and simplified the X-axis label under the assumption that the input is always sweep data with ms a x wave units.

timjarsky:
@MichaelHuth, thank you for adding the new features. I would still like to add an x-axis option that plots the frequency/time (Y) values against spike count instead of spike time.

Minor: when using the normalized option, y-axis labels could be something like: "normalized frequency" instead of (Hz), which is confusing.

timjarsky:
@MichaelHuth In addition to axis label changes you may also want to utilize the legend labels to make plots like this interpratable.

image

timjarsky:
@MichaelHuth It would also be helpful if the Y-axis label included the method string.

MichaelHuth:
Good for another try. I added after normalize another argument that can either be `time` or `count`. With `time` as default.

The bigger part was to make traces and annotations from different formulas in the same plot distinguishable.
I added that as general approach but it is currently setup only for apfrequency.
So with your example from the image it should now adapt the annotation and show what argument is different and change the marker/line style. For the same arguments the style in one plot stays the same for multiple formulas combined with the "with" keyword.

timjarsky:
@MichaelHuth, how are the normalized values calculated with method `full`?

I was expecting the value to be normalized to the first sweep.

(third from the top plot)
image

MichaelHuth:
The normalization factor does not depend on the method, only if you use `time` or `freq`.
For formula three you use `freq`, so
```igorpro
peakTimeFreq = 1.0 / ( (peaksAt[1] - peaksAt[0]) * MILLI_TO_ONE) // frequency in Hz from the first two peaks
result = result / peakTimeFreq // apply normalization factor
```

timjarsky:
> The normalization factor does not depend on the method, only if you use `time` or `freq`. For formula three you use `freq`, so
> 
> ```igor
> peakTimeFreq = 1.0 / ( (peaksAt[1] - peaksAt[0]) * MILLI_TO_ONE) // frequency in Hz from the first two peaks
> result = result / peakTimeFreq // apply normalization factor
> ```

Thanks for the explanation @MichaelHuth  So, I think normalization should depend on the method. For methods where a single value is returned for each sweep, the normalization should be across all sweeps.

timjarsky:
@MichaelHuth , i want to modify what I said above. Rather than have the normalization depend on the method, i think it makes sense to give the user to specify if they want the normalization to occur within  the sweep or across sweeps.

MichaelHuth:
@timjarsky Good for another test. The allowed values for the normalization argument are now: `nonormalize`, `normoversweeps`, `normwithinsweep`, where `nonormalize` is default.

normwithinsweep works as before.

normoversweeps uses the first peak difference from the first sweep for the normlization (time or freq) and applies that to all sweeps.

timjarsky:
> normoversweeps uses the first peak difference from the first sweep for the normlization (time or freq) and applies that to all sweeps.

@MichaelHuth 
I don't understand how norm-over-sweeps is normalizing. Should the normalized value of the first sweep be 1?

I'm getting 
image
with 
```
apfrequency(
data(E1,
select(channels(AD0),sweeps())),
0,0,freq,normoversweeps)
```

timjarsky:
> For `normoverweeps`:
> 
> * For all sweeps t_1,1 is used as normalization value and all t_x,y are divided by it.
> 
> In case sweep 1 has no pair, then a possible normalization value from the first pair of sweep 2 is used a.s.o.
> 
> For a single sweep both calculations lead to the same normalized result. The result values are determined by the method. In a general case the sweeps don't have the same number of pairs.

Thanks for the detailed explanation @MichaelHuth. 

In the example I posted, and based on the snippet above, I would expect the first value to be 1, but it's not. Why is that?

timjarsky:
@MichaelHuth Sorry if my last post could be more precise. For each method ```0` for "full", `1` for "instantaneous" , `2` for apcount, or 3 for InstVsTime```, ```normoversweeps``` normalization should be relative to the result that's specific to each method. For example, the ```full``` method returns the average frequency (number of spikes divided by the pulse duration), so the normalization should be relative to the max (or min; can we offer the user a choice?) frequency observed across all sweeps. Does that make sense to you?

MichaelHuth:
@timjarsky Ok. One more thing: If the sweep contains only a single peak, no frequency or time intervall can be determined. The original algorithm takes then just the position (time) of level transition as result. Since that is actualy no time intervall or frequency I handle these results differently and skip them for normalization. Now if you have sweeps with single peaks and multiple peaks then you get mixed results, where single peaks give positions and multiple peaks frequency or time intervals.
I can keep it this way, but I think the mixing of different results is a problem as currently there is no direct indicator for it. Only the vertical axis label would show mixed data of normalized/non-normalized while some normalization was specified.

I think it would be better to return no results for apfrequency methods that require pairs but have only a single peak.
Does that make sense?

timjarsky:
> @timjarsky Ok. One more thing: If the sweep contains only a single peak, no frequency or time intervall can be determined. The original algorithm takes then just the position (time) of level transition as result. Since that is actualy no time intervall or frequency I handle these results differently and skip them for normalization. Now if you have sweeps with single peaks and multiple peaks then you get mixed results, where single peaks give positions and multiple peaks frequency or time intervals. I can keep it this way, but I think the mixing of different results is a problem as currently there is no direct indicator for it. Only the vertical axis label would show mixed data of normalized/non-normalized while some normalization was specified.
> 
> I think it would be better to return no results for apfrequency methods that require pairs but have only a single peak. Does that make sense?

@MichaelHuth I agree that it makes sense to return no result **for sweeps** used in apfrequency methods that require pairs but have only a single peak.

I appreciate that you added normalization methods without adding parameters. e.g., ```normoversweepsmin``` and ```normoversweepsmax``` Thank you!

timjarsky (approved):

timjarsky (dismissed):

t-b:
@MichaelHuth 

Docu build says:
```
/home/ci/Packages/MIES/MIES_SweepFormula_Helpers.ipf:627: warning: The following parameter of SFH_TransferFormulaDataWaveNoteAndMeta(WaveRefWave input, WaveRefWave output, string opShort, string newDataType, WaveText argSetup=defaultValue, variable keepX=defaultValue) is not documented:
  parameter 'argSetup'
/home/ci/Packages/MIES/MIES_SweepFormula_Helpers.ipf:627: warning: The following parameter of SFH_TransferFormulaDataWaveNoteAndMeta(WaveRefWave input, WaveRefWave output, string opShort, string newDataType, WaveText argSetup=defaultValue, variable keepX=defaultValue) is not documented:
  parameter 'argSetup'
```

t-b:
@MichaelHuth Nice. Thanks!

timjarsky (approved):

PR #1617, opened 2023-01-11
Always use sweeps for the tpfit x-axis

### LINKED ISSUE #1615: TP operation with tpfit does not use a custom x axis
```
runwithOpts(testcase = "REpeatSets_1")
```

SF:
```
tp(tpfit(doubleexp, tausmall, 500), select(channels(AD0), sweeps(), all))
```

gives
image

t-b:
Close #1615.

@MichaelHuth Please review.


MichaelHuth (approved):
Ok, the only thing is that if you remove the value check in the test then it stays untested if a given fit calculates the corect value.

timjarsky:
@t-b this works in my hands; to results are displayed vs sweep by default

Do you think this also work?

```
tp(tpfit(doubleexp, tausmall, 500), select(channels(AD0), sweeps(), all))

/

tp(tpfit(doubleexp, tau, 500), select(channels(AD0), sweeps(), all))
```

this does:

```
[1,2,3,4]
/
[1,2,3,4]
```

t-b:
> Ok, the only thing is that if you remove the value check in the test then it stays untested if a given fit calculates the corect value.

Yes your are totally right. I think we need to change to preacquired data for checking analysis values.

PR #1629, opened 2023-02-03
SFH_GetArgumentAsNumeric/SFH_GetArgumentAsText: Allow to check the parameter for a list of allowed values

t-b:


t-b:
@MichaelHuth Please have a look

MichaelHuth (approved):
Looks good.

I just think that the overhead from always creating the error message is a bit high.

Do you think it would be a good idea to combine ASSERT with our own printf from the IUTF?
So we could have the condition check before we create the output message.

PR #1633, opened 2023-02-10
BSP_SFHelpWindowHook: Interrupt the user less when typing

t-b:
Currently the coloring of the notebook commands requires to change the notebook selection. This can interfere quite easily with the user typing. So let's skip coloring the keywords if the contents did not change and only do it on deactivation.

It really is annoying.


t-b:
@timjarsky Ready!

PR #1653, opened 2023-03-17
EPSP prep v6

t-b:
@MichaelHuth Please have a look.

### DIFF
--- a/Packages/tests/Basic/UTF_SweepFormula.ipf
+++ b/Packages/tests/Basic/UTF_SweepFormula.ipf
@@ -60,6 +60,73 @@ static Function/S CreateFormulaGraphForBrowser(string browser)
 	return win
 End
 
+/// Add 10 sweeps from various AD/DA channels to the fake databrowser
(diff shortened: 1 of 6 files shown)


timjarsky (approved):

Packages/tests/Basic/UTF_SweepFormula.ipf
```diff
@@ -2472,76 +2486,20 @@ End
...
-			config[j][%ChannelType]   = XOP_CHANNEL_TYPE_DAC
-			config[j][%ChannelNumber] = channelNumber
-		endfor
-
-		// create sweeps with dummy data for sweeps() operation thats called when omitting select
-		MoveWave sweep, dfr:$GetSweepWaveName(sweepNumber)
-		MoveWave config, dfr:$GetConfigWaveName(sweepNumber)
-		MIES_DB#DB_SplitSweepsIfReq(win, sweepNumber)
-
-		// channels setup DA: 8, 6, NaN, 4, NaN, 2, NaN, 0, NaN
```
MichaelHuth, line 2523: I think the comment should also be transferred, as it helps to understand the test.
At first glance this difference between ADC and DAC is not as obvious.

MichaelHuth (approved):
Just a minor comment regarding the loss of the comment.
The speedup in the helper getters is nice, though we have to use them then also in other operations...

Issue #276, opened 2019-09-20
SweepFormula: add data as variables


ukos-git:
Similar to JSON logic, the data should get separated from the operations. This ensures that we do not call the getter two times if for example, calculating with it.

`derivative(data(cursors(a,b), channels(AD), sweeps())) * data(cursors(a,b), channels(AD), sweeps())`

will then reduce to

`derivative(var0)*var0`

with the data object

`var0=data(cursors(a,b), channels(AD), sweeps());`

 It can also mean that the functions are a lot easier to read. and you can name the data.

To get to that point, string values need to get checked if a variable is registered for them. Probably related to #274 

I don't see significant problems implementing that.

ukos-git:
http://jsonlogic.com/

t-b:
- data will be defined in separate notebook
- will make it faster

ukos-git:
From the json logic documentation about the `var` operation:

logic:

```json
{ "and" : [
  {"<" : [ { "var" : "temp" }, 110 ]},
  {"==" : [ { "var" : "pie.filling" }, "apple" ] }
] };
```
data:
```json
{ "temp" : 100, "pie" : { "filling" : "apple" } }
```

should return `true` aka `1`

t-b:
Example:
```
currSel = select(channels(AD9), [31,33,34], all)
myrange = [2000,4000]
kernel = epspKernel(myrange, currSel)

epsp(kernel, 0.6, 3, (-6)), 0.04, 550, 0)

and

epspStats(myrange, currSel, tau, accept, hist)
```

Questions:
- Do we need to mark named entities? With `$`?
- Needs to be known across all `with` and `and` parts
- No partial expression support

PR #1664, opened 2023-03-28
Add variable support to SweepFormula

MichaelHuth:
Example:

```igorpro
s = select(channels(AD),sweeps(),all)
c = cursors(A,B)

data($c,$s)
```

Example2:
```igorpro
s = select(channels(AD),sweeps(),all)
c = cursors(A,B)
d = data($c,$s)

$d

```

@timjarsky @t-b initial version

close https://github.com/AllenInstitute/MIES/issues/276


timjarsky:
> Example:
> 
> ```igor
> s = select(channels(AD),sweeps(),all)
> c = cursors(A,B)
> 
> data($c,$s)
> ```
> 
> Example2:
> 
> ```igor
> s = select(channels(AD),sweeps(),all)
> c = cursors(A,B)
> d = data($c,$s)
> 
> $d
> ```
> 
> @timjarsky @t-b initial version
> 
> close #276

@MichaelHuth, wow, that was fast! Unfortunately, I have a busy day today, but I should be able to try it out tomorrow. Thank you!!

t-b:
Thanks for the quick PR. I've played around with it and it works in all good cases.

Some bad cases could be catched with a more strict regexp I think. How about something like

```
^(?i)\\s*([A-Z]{1}[A-Z0-9_]*)\\s*=(.+)$
```

This not only adds front and back locking but also the generic "ignore case" and `.` instead of `[\\s\\S]`.

With the current regexp something like

```
132s = select(channels(AD),sweeps(),all) # comment
```

gets the variable name `s`.

timjarsky:
@MichaelHuth this is so much better!

```
d = data(E1,select(channels(AD0),sweeps()))

$d

and

apfrequency($d,0,0,freq,normoversweepsmin)

with

apfrequency($d,0,0,freq,normoversweepsmax)

with

apfrequency($d,0,0,freq,normoversweepsavg)

and

apfrequency($d,3,0,freq,normoversweepsmax)

with

apfrequency($d,3,0,freq,nonorm)

and

apfrequency($d,3,0,time,nonorm,time)

and

apfrequency($d,3,0,freq,norminsweepsmax,time)
```

vs.

```
data(E1,
select(channels(AD0),sweeps()))

and


apfrequency(
data(E1,
select(channels(AD0),sweeps())),
0,0,freq,normoversweepsmin)

with

apfrequency(
data(E1,
select(channels(AD0),sweeps())),
0,0,freq,normoversweepsmax)

with

apfrequency(
data(E1,
select(channels(AD0),sweeps())),
0,0,freq,normoversweepsavg)


and

apfrequency(
data(E1,
select(channels(AD0),sweeps())),
3,0,freq,normoversweepsmax)

with

apfrequency(
data(E1,
select(channels(AD0),sweeps())),
3,0,freq,nonorm)

and

apfrequency(
data(E1,
select(channels(AD0),sweeps())),
3,0,time,nonorm,time)

and

apfrequency(
data(E1,
select(channels(AD0),sweeps())),
3,0,freq,norminsweepsmax,time)
```

t-b:
Works as advertised for me as well. @timjarsky Can Michael finish it?

t-b:
@MichaelHuth 

Why does

```
d = data(E1,select(channels(AD0),sweeps()))
e = $d,3,0,freq,norminsweepsmax, time

apfrequency($e)
```

not work?

MichaelHuth:
What happens is that $e is evaluated to an array, but it has mixed data type content.
This results in `$e = [NaN,3,0,NaN,NaN,NaN]`.

Also when the operation
apfrequency starts parsing its arguments, it sees a single argument. The executor will then replace the $e string with the array from the variable. Effectively apfrequency then sees as first argument (where data should be) the parameter array from $e.

timjarsky (dismissed):

t-b:
@MichaelHuth THanks for the explanation, so I guess there is no easy way to fix that.

t-b:
Review:
- UTF_SweepFormulaHardware.ipf does not compile
- `a = unknown`

triggers

```
  BUG_TS: Encountered pending RTE: 734, DefineGuide;Illegal dual guide position.
```

178a0d95 (SF: Add utility function ParseFormulaToJSON, 2023-03-28)
066313d8 (SF: Allow $ character in strings parsed by SF_FormulaParser, 2023-03-28)

Jeep.

6883d68c (SF: Move kill of working DF outside of SF_FormulaPlotter, 2023-03-28)

Nice.

1b583539 (SF: Add parse and execute for variable names, 2023-03-28)

- Is it possible to just resize the existing wave to zero rows instead of

```
+    WAVE/WAVE varStorage = GetSFVarStorage(dfr)
+    KillOrMoveToTrash(wv = varStorage)
+    WAVE/WAVE varStorage = GetSFVarStorage(dfr)
```

b83d8840 (SF: Add variable support to formula executor, 2023-03-28)

```
+    WAVE/Z/T varAssignments = $""
+    [varAssignments, code] = SF_GetVariableAssignments(preProcCode)
```

IP9 now supports

```
[WAVE/T varAssignments, code] = SF_GetVariableAssignments(preProcCode)
```

The /Z is implicit (unfortunately, already reported).

Querying varStorage should be moved into its own function SF_FormulaExecutor is way too big already.

965fee7e (SF: Add support to check also variables with the check button, 2023-03-29)

SF_GetVariableAssignments: Our usual convention is to return $"" instead of an empty wave.

10601dfb (SF: Use try/catch approach for checking formula input, 2023-03-29)

Fine.

09f52c9f (SF: Add support for with keyword when checking formulas, 2023-03-29)

Good find. We use small caps "graph" and "formula" in the JSON tab so small caps variable would be better though.

49a543f4 (SF: Tests for variable support in SF, 2023-03-29)

- The test should also check that the correct number of entries and correct name was added to varStorage. It might make also sense to split the passing tests into an own test and use multi data tests for that, but I'll leave that to you.
- Variable case insensitivity should also be tested

- One test where you check that the executor replaced it correctly would also be nice.

- Test also that

```
b = 1

a$b vs 0
```

works and keeps working.

a921e1a3 (SF: Add variable documentation to SF help notebook, 2023-03-30)
0ae46241 (SF: Add variable SF feature to MIES documentation, 2023-03-30)

Both nice!

MichaelHuth:
@t-b Could you give me the test results?

t-b:
@MichaelHuth Sure, UTF_SweepFormula#TestInputCodeCheck fails.

t-b:
LGTM.

t-b (approved):

PR #1672, opened 2023-03-31
Warn about empty wavebuilder epochs

### LINKED ISSUE #8: Prevent the user from creating single point epochs in the WaveBuilder
We probably need to set the limit a bit higher than 100mus as we want to have a valid wave even after decimation. Maximum decimation is a factor of 160 (16 from sampling multiplier and 10 from DAQ).

t-b:
Close #8.


t-b:
The failing tests, e.g. ExportAndLoadOfCustomWaves, are due to the fact that when creating a new stimset with the stimset API, ST_CreateStimSet, we save an stimset with the initial parameters. But the initial parameters have one epoch with duration zero.

```
•print ST_CreateStimSet("setA", CHANNEL_TYPE_DAC)
  The epoch 0 has a duration of zero. Please either remove that epoch or make its duration non-zero.
```

Maybe we should default to initial paramters which create a valid stimset?

timjarsky:
@t-b this catches empty epochs in the WB. Should it also catch empty epochs at data acquisition time?

t-b:
Solution: ST_CreateStimset should set the first epoch to a length of 1ms to make it valid.

t-b:
Fixed. I now also warn about empty epochs before DAQ. And also care about unassociated DA/TTL channels by adding the check into DAP_CheckStimsets. The hack to check that in the epoch generation is therefore not necessary anymore.

timjarsky:
@ t-b epochs that are too short for acquisition still show up as epochs in the data browser. It doesn't happen for epochs at the beginning of a sweep but for those in the middle. Is that expected?
image

[1672TestingDemo.zip](https://github.com/AllenInstitute/MIES/files/11558406/1672TestingDemo.zip)

t-b:
> @ t-b epochs that are too short for acquisition still show up as epochs in the data browser. It doesn't happen for epochs at the beginning of a sweep but for those in the middle. Is that expected?

No but...
We don't create epoch info for epochs which would be just a single point since 068dbd3a (DC_AddEpoch: Don't add single point epochs, 2021-06-10). So every epoch if it would be really too small is dropped. I guess that your first epoch is like that. I have a look tomorrow and confirm that.

Issue #1669, opened 2023-03-31
Adapt SF_ExecuteFormula to support variables


MichaelHuth:
- add optional parameter to SF_ExecuteFormula to enable variable evaluation before (remaining) formula is evaluated. (default = on), e.g. `noVariables`
- Move to function:
```
formula = SF_PreprocessInput(formula)
jsonId = SF_ParseFormulaToJSON(formula)
WAVE/Z result = SF_FormulaExecutor(databrowser, jsonId)

``` 

- Prefer `SF_ExecuteFormula` over directly calling `SF_FormulaExecutor` except for SF internal usage. Adapt `SF_GetMultipleResults` in tests

PR #1678, opened 2023-04-03
Adapt SF_ExecuteFormula to support variables

MichaelHuth:
- adapted SF_ExecuteFormula
- Removed GetSingleResult/GetMultipleResults from SF tests in favor of SF_ExecuteFormula
- simplified dfr usage in SF

close https://github.com/AllenInstitute/MIES/issues/1669


t-b:
Failing test:
```
SweepFormulaHardware#SF_ButtonTest:NI in UTF_SweepFormulaHardware.ipf (396)                                 
```

with 
```
ERROR: Encountered "AbortOnValue" Code 1 in test case SweepFormulaHardware#SF_ButtonTest" (UTF_SweepFormulaHardware.ipf)```

**Comment by t-b** [comment]
Review:

- [ ] What is the reason for the removal of `GetMainWindow` in

```
 Function/DF SFH_GetWorkingDF(string win)

-    DFREF dfr = BSP_GetFolder(GetMainWindow(win), MIES_BSP_PANEL_FOLDER)
+    DFREF dfr = BSP_GetFolder(win, MIES_BSP_PANEL_FOLDER)

     return createDFWithAllParents(GetDataFolder(1, dfr) + SFH_WORKING_DF)
 End
```
?

And why not just call `SF_GetBrowserDF` which should actually use `GetMainWindow` as in 

```
Function/DF SF_GetBrowserDF(string graph)

    return BSP_GetFolder(GetMainWindow(graph), MIES_BSP_PANEL_FOLDER)
End
```

The reason we do have these `GetMainWindow` calls sprinkled out in these getter functions is that we want to relieve the caller from having to think about at what exact subwindow the window string is actually pointing to. We only care that the base window is correct here.

- [ ] Can we have a non-negated optional argument for `SF_ExecuteFormula`? Something like `useVariables`? That is easier to grasp than the negated version `noVariables`. I'm totally fine with whatever default you see fitting for that variable.

t-b:
Another thing:
- If you move `SFH_GetArgument` and `SFH_ParseArgument` into MIES_SweepFormula.ipf you can make both `SFH_ParseArgument` and `SF_FormulaExecutor` static :)

t-b (changes requested):
See comment

MichaelHuth:
> Review:
> 
> * [ ]  What is the reason for the removal of `GetMainWindow` in
> 

The reason is that `BSP_GetFolder` itself calls GetMainWindow, so there is no need to call it twice.
https://github.com/AllenInstitute/MIES/blob/2453e2567d7192e68dcb54fdbfc0ec8d88b7273f/Packages/MIES/MIES_BrowserSettingsPanel.ipf#L380

t-b (approved):
LGTM.

Issue #1552, opened 2022-11-13
SweepFormula: Revise average operation


t-b:
- optional parameter to choose averaging inside timeseries or across sweeps (datawaves returned by select())
- don't add averaging across channels for now
- **not** for other statistical operations
- use trace colors for average wave from databrowser
- make it dashed

PR #1682, opened 2023-04-05
SF: Change average operation to support two modes

MichaelHuth:
- avg(array data[, string mode])
- mode can be in or over
- in: average each element of a data set -> output single point waves
- over: average over all elements of a data set -> output single wave
- averaged wave is shown in sweep average color as in data browser and style dashed

close https://github.com/AllenInstitute/MIES/issues/1552


MichaelHuth:
@timjarsky please test if it works as intended. e.g. `avg(data(cursors(A,B), select(channels(AD),sweeps(),all)),over)`. If it is good, I add the tests and docu.

The averaging over the input data is set to ignore NaNs.

timjarsky:
@MichaelHuth works beautifully! Thank you!! My only nitpick is that the average trace is somewhat hard to see.

image

t-b (changes requested):
See comment.

MichaelHuth:
> even more correct would be if(IsFinite(...))

The !IsNaN check is targetted to know if there was a JSON entry or not. I add a utility function to check for a valid trace linestyle instead.

MichaelHuth:
> missing !! in front of JWN_GetNumberFromWaveNote( to force it to boolean

That is actually more tricky here, because a non existing JSON entry returns NaN and !!NaN is 1 because NaN is nonzero for the ! operator. However, if you do an if(NaN) then it is considered zero aka false.
```
	if(NaN)
		print "It is NaN"
	endif
	if(!!NaN)
		print "It is !!NaN"
	endif

```
I change the code to not rely on these Igor Pro shenanigans.

t-b:
> The !IsNaN check is targetted to know if there was a JSON entry or not. I add a utility function to check for a valid trace linestyle instead.

Nice!

> That is actually more tricky here, because a non existing JSON entry returns NaN and !!NaN is 1 because NaN is nonzero for the ! operator. 

Good point. Thanks for fixing my idea. Your new solution reads well!

Rest looks good as well.

timjarsky (approved):

Issue #620, opened 2020-08-31
JSON configuration: How to inform users about outdated configuration files


t-b:
Consider the DAEphys panel.

When we add a new control to the panel, users with existing configuration files don't have that control in their config. The question is how can we tell the user what is missing from the configuration file? Creating two json files and just presenting the diff output is probably too complicated.

t-b:
Another issue is when we rename or remove controls. We inform the user that the control does not exist anymore, but from then on she is left on her own.

Ideas:
- Create a new JSON config file which has the now non-existing controls removed
- Or we create a new JSON config file from the GUI and the merge the old JSON config file into that overriding the values from the GUI. And use that as new file.

t-b:
This came up again recently.

The question was how to incorporate new JSON entries defined via `CONF_DefaultSettings` without having to manually edit the JSON file.

When you load a DAEPhys JSON config and then after loading save it again this is what you get in the new config:
- All non-existing-anymore controls, which were still referenced in the original config, are not present anymore
- All new controls are present
- The experiment configuration block defined in `CONF_DefaultSettings` is written with the default values and not the used ones. New entries are added.
- Any idea of the rig file (if used) is forgotten

timjarsky:
Add a menu option(s) to update config files for a standard configuration and window configuration (the user specifies the file path in the case). Overwrite the old files.

MichaelHuth:
My idea would be that we extend our SyncJSON function.

Then we sync the current config in the old config such that the current config JSON is traversed and
* if it exists in both then the entries stay in the old config (needs JSON_Exists)
* if it exists only in the current config, then the entries are added to the old config (needs JSON_GetJson / JSON_AddJson)
* if it exists in the old but not in the current config, then it is removed from the old config (needs JSON_Remove)

For that it would be very convenient to read and write JSON objects from jsonPaths. While we have a function JSON_AddJson / JSON_SetJson we don't have JSON_GetJson (or JSON_Dump with a given path).

PR #1694, opened 2023-04-25
Proper support for DAEphys configuration file upgrade with rig file support.

MichaelHuth:
- [x] Suggest suitable new configuration file name
- [x] Copy previous rig file over
- [x] Tests
- [x] Docu

close https://github.com/AllenInstitute/MIES/issues/620


t-b:
@MichaelHuth Can't check right now, but I'm pretty sure that one of the test experiments does not compile.

t-b:
Review:

89984250 (Util: Add function that copies a simple JSON element from src to target, 2023-04-25)

Looks good. But please remove the empty line after `ASSERT` in the default branch.

58754201 (CONF: Add constant for filename separator in filename list in window userdata, 2023-04-25)

Good point!

da15c29c (CONF: Change DAEphys save handling if window was previously restored, 2023-04-25)

Commit message:
intro -> into

CONF_TransferPreviousDAEphysJson:
Please add `string entry` per our coding conventions.
Same for CONF_RemoveRigElementsFromDAEphysJson.

CONF_RemoveRigElementsFromDAEphysJson:
- You pass `ignoreErr=1` to JSON_Remove. In which situation would the removal fail and you want to ignore that?

f6bdd842 (CONF: Replace settings templates with working setting templates, 2023-04-25)

Nice find. Do we now have tests which check that these keep working?

0091a813 (Util: Extend SaveTextFile to optionally return the name of the saved file, 2023-04-26)

Okay.

87dc1cb5 (CONF: Saving DAEphys configuration now refes to the previous conf file name, 2023-04-26)

Works!

83b3db1e (CONF: If configuration save would overwrite a file show dialog, 2023-04-26)

I'm not sure here. If we go the dialog route we would also have to use
GetOpenZFlag, as we must *never* show a dialog in CI.

But stepping back, why not just generate a unique filename with
UniqueFileOrFolder and avoid the overwrite and dialog problem all together?

That change also breaks the RestoreDAEphysPanel test if run twice.

4eae503a (Tests: Add GetTestStimsetFullFilePath helper function, 2023-04-27)

Good point introducing that. Can that function also be used in
FixupJSONConfig_IGNORE? And can GetTestStimsetFullFilePath assert out if the
file does not exist?

c4946ed6 (Tests: Add test that loads the template configuration, resaves and reloads, 2023-04-27)

As already mentioned that needs adaptation for both hardware types and changed stimset path. And should also be not in UTF_VeryBasicHardwareTests.ipf but more in UTF_DAEphys.ipf. Or create UTF_ConfigurationHW.ipf and move all configuration tests there (every configuration test uses FixupJSONConfig_IGNORE).

3ba67be0 (CONF: Add documentation for changed DAEphys configuration save behavior, 2023-04-27)

I've tried the whole upgrade enchilada. And it works! Nicely done. One minor
thing is that the new name is all lower case now. Can that be fixed?

```
$ ls -l1
-rw-r--r-- 1 thomas 197121 24852 Apr 27 21:12 1_DA_Ephys.json
-rw-r--r-- 1 thomas 197121   331 Apr 27 21:12 1_DA_Ephys_rig.json
-rw-r--r-- 1 thomas 197121 23487 Apr 27 21:14 1_da_ephys_new.json
-rw-r--r-- 1 thomas 197121   331 Apr 27 21:14 1_da_ephys_new_rig.json
-rw-r--r-- 1 thomas 197121 10503 Apr 27 21:12 2_DataBrowser.json
```

MichaelHuth:
> One minor thing is that the new name is all lower case now. Can that be fixed?

Actually this can not be fixed. I noticed that too and the issue here is the following:

We have the file name of the old file that was saved in the user data in lower case.
This file exists. I would like to get now the correct case from the existing file, but
Igor Pro never reads that properly again. The main options are:
* GetFileFolderInfo: Retrieves the path in the correct case, but leaves the file name case as given by the parameter string.
* Open: Returns in S_fileName the same case as originally given in the file name string
* FStatus: Returns in S_fileName the same case as originally given in the file name string when Open was called

Only when I use Open with a Dialog the casing as present in the file system is returned. But this is no option here.

I would rate this behavior as bug, because all Functions that are directly file system related should return the actual case-correct file name from the file system.

t-b:
> I put a workaround in now, that is that I do not apply LowerStr when Loading the config, such that the original case is saved in the user data. (but thats not actually solving the problem)

That should do the trick for us. Thanks for the digging.

t-b (dismissed):

t-b:
Rebased to get GH actions CI.

t-b:
Rebased and fix compilation due to merge of #1701.

timjarsky:
@MichaelHuth I got the following error when executing a config file

```
  Configuration restore aborted at file C:Users:timj:OneDrive - Allen Institute:Documents:git repos:MIES:Packages:Settings:DA_ephys_1600.json.(ITC1600_Dev_0) The sweep number ordering is not strictly ascending.
```

I had already acquired data with the device, although I'm not sure that's relevant as I don't fully understand the error message.

[Config_1694.zip](https://github.com/AllenInstitute/MIES/files/11558752/Config_1694.zip)

t-b:
> ` Configuration restore aborted at file C:Users:timj:OneDrive - Allen Institute:Documents:git repos:MIES:Packages:Settings:DA_ephys_1600.json.(ITC1600_Dev_0) The sweep number ordering is not strictly ascending.`

That should not happen but is also unrelated. If I understand the issue correctly this should also be fixed with #1700.

@MichaelHuth Can you fix conflicts and rebase? Thx.

t-b:
@timjarsky Can you try again?

t-b:
@MichaelHuth And again...

timjarsky (dismissed):

Packages/MIES/MIES_Configuration.ipf
```diff
@@ -90,6 +90,19 @@
...
+///
+/// *_rig.json configuration files store settings that are specific for a workplace.
+/// When restoring a DAEphys panel the settings from the rig file are joined with the settings in the DAEphys configuration file.
+/// Having entries for the same setting in both files is invalid and an assertion will be thrown.
+///
+/// Saving a DAEphys panel that was originally restored from a configuration file:
+/// - entries from the "Common configuration data" block are updated with the values from the previous configuration,
+///   if they already existed. New entries are set at default values.
+/// - entries defined in an associated rig file are removed from the DAEphys panel configuration because they would appear in
+///   both files otherwise
```
timjarsky, line 102: ~~otherwise~~

timjarsky:
@t-b I made minor comments on the documentation. No need to send it back to me for review.

t-b:
@MichaelHuth See comments from Tim. Please assign me once done.

PR #1696, opened 2023-04-26
WB_AssembleSetName: Allow longer names

t-b:


timjarsky:
@t-b when the name is too long, and the user tries to save the stim set, the stim set disappears and isn't saved without warning. The preferred behavior is to prevent saving, maintain the stim set in the WB, and warn the user.

t-b:
> The preferred behavior is to prevent saving, maintain the stim set in the WB, and warn the user.

This is what I'm doing now.

timjarsky (approved):

PR #1702, opened 2023-05-03
Fix for GUI getters returning string through S_value could be null

MichaelHuth:
GUI getters GetPopupMenuString, GetSetVariableString, GetValDisplayAsString were not aware that ControlInfo can return S_value that is null if the controls value was never set before.

This caused for theses cases a RTE due to the null string being returned. The RTE was not caught and thus pending.


t-b:
THanks for the fix.

Code looks good but the tests should use `CHECK_NO_RTE()` instead of `PASS()`.

t-b:
Doh merged before morning coffee. Well I think we can live with `PASS.

Issue #1708, opened 2023-05-09
IVSCC: Core 1 crash


t-b:
> This has happened multiple times on rig 4 in the last week.

[Downloads.zip](https://github.com/AllenInstitute/MIES/files/11431227/Downloads.zip)

t-b:
That one is a crash we have seen a couple of times now. Already reported as #4173 to WM.

We do need a recipe to get further.

t-b:
It's unfortnuately a known crash we don't know how to replicate.

@timjarsky:
- Anything happening specially when the crash happens?
- Can you upload the logfiles via MIES Panels->Advanced->Upload log files and post the ticket number?
- Any accessibility features turned in Windows?

t-b:
Ticket: 63fadb84-f59d-46a2-b8d7-e3f65d508b04

t-b:
Version:

```
Release_2.6_20220606-720-g11874eca8
Date and time of last commit: 2023-01-13T17:42:47+01:00
Submodule status:
 e46385bee5c53a0daa1eca0c1d6af44d19677c13 Packages/IPNWB (e46385b)
 d229d451f35e88b8c38edb8c26533acfbc3fd10f Packages/doc/doxygen-filter-ipf (heads/main)
 873dee901d747ceedfce01e66c058c8791a3b046 Packages/unit-testing (873dee9)
```

t-b:
The crash happened at `Date: 2023-05-08 13:49:16`.

We have the following two log entries from BUG/BUG_TS in the vincinity

```
[...]

PR #1715, opened 2023-05-18
Improve logging on ITC errors

t-b:
That would have been helpful when investigating #1708.


MichaelHuth:
I realized we already have a GetFileSize utility function. I used that instead of FStatus.

t-b:
Thanks for your commit.

Can you:
- Split the commit into smaller parts as it currently touches three different places
- Keep FStatus as that is faster than GetFileFolderInfo
- Integrate the changes into your currrnt PR as I'll close this PR. My changes are included in #1747.

Thanks and sorry for the late reply.

t-b:
Will be integrated via #1748.

PR #1731, opened 2023-05-19
Switch to JSON wave serialization for store in all cases

t-b:


timjarsky:
@t-b In trying to test this PR, I got the `Name parameter must be textual` error, but it also happens in main with `store([1,2,3],[4,5,6])`

timjarsky:
@t-b JSON formatting looks good, but I guess this still won't solve storing PSX histogram results because store operation doesn't yet support `multiple data waves as input.` Is that right?

t-b:
> @t-b In trying to test this PR, I got the `Name parameter must be textual` error, but it also happens in main with `store([1,2,3],[4,5,6])`. So, not sure if it should be fixed as part of this PR, but it seems like it should work, right?

Your holding it wrong. The first parameter to store must be the name, as in 

```
store(abcd, [[1,2,3],[4,5,6]])
```

and that gives here

```json
{
    "data": {
        "raw": [
            1,
            4,
            2,
            5,
            3,
            6
        ]
    },
    "date": {
        "modification": 0
    },
    "dimension": {
        "size": [
            2,
            3
        ]
    },
    "type": "NT_FP64"
}
```

> JSON formatting looks good, but I guess this still won't solve storing PSX histogram results because store operation doesn't yet support multiple data waves as input. Is that right?

Good point. I opened an DB, enabled OVS and selected two sweeps and the executed

```
store(abcd, data([A, B]))
```

which gives `Multiple dataSets not supported yet for store()`. I think I can fix that now by outputting an array of JSON encoded waves. Thanks for finding that one!

timjarsky:
@t-b oops, sorry for the silly mistake. 

JSON question: is it the 
```
"dimension": {
        "size": [
            2,
            3
        ]
```

that tells us the raw data is made up of two arrays that are three rows long?

t-b:
> that tells us the raw data is made up of two arrays that are three rows long?

yes!

t-b:
@timjarsky Ready for another try :)

t-b:
I've also revised https://github.com/AllenInstitute/ZeroMQ-XOP#wave-serialization-format to have some examples including text wave and wave reference wave.

timjarsky:
works gud! @t-b

timjarsky (approved):

PR #1732, opened 2023-05-22
bugfix: GetAllFilesRecursivelyFromPath use reliable name retrieval

MichaelHuth:
- IndexedDir and IndexedFile return an RTE if the directory or folder can not be accessed. The functions also returns an emtpy string in that case. Thus, as soon as such a folder was encountered further traversal was cancelled. A typical folder is $RecycleBin in Windows where that applies.
- Now the functions are used to retrieve all files or folders at a time. Also RTE conditions are ignored.

### DIFF
--- a/Packages/MIES/MIES_Utilities.ipf
+++ b/Packages/MIES/MIES_Utilities.ipf
@@ -2811,11 +2811,11 @@ End
 Function/S GetAllFilesRecursivelyFromPath(pathName, [extension])
 	string pathName, extension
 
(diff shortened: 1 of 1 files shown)


t-b:
Review:

Typo:
- "The functions also returns" -> "The functions also return"

- The commit message does not explain why we haven't used the shown approach since the beginning.
- I think the "ListMatch" in the return statement is no superfluous or?

t-b (approved):
LGTM.

Issue #1714, opened 2023-05-18
Logfiles are getting too big for date extraction


t-b:
Coming here from #1708.

Version:

```
Release_2.6_20220606-720-g11874eca8
Date and time of last commit: 2023-01-13T17:42:47+01:00
Submodule status:
 e46385bee5c53a0daa1eca0c1d6af44d19677c13 Packages/IPNWB (e46385b)
 d229d451f35e88b8c38edb8c26533acfbc3fd10f Packages/doc/doxygen-filter-ipf (heads/main)
 873dee901d747ceedfce01e66c058c8791a3b046 Packages/unit-testing (873dee9)
```

In IVSCC logfiles I'm seeing

```
Log.jsonl.0Ffw7Y:{"action":"report","exp":"2023_05_08_154443","id":"7c5fab5","msg":"Encountered pending RTE: 823, PadString;Parameter is out of 
range","source":"BUG_TS","stacktrace":"Stacktrace:\rCONF_AutoLoader(...)#L318 
 [MIES_Configuration.ipf]\rCONF_RestoreWindow(...)#L442
 [MIES_Configuration.ipf]\rCONF_RestoreDAEphys(...)#L677
 [MIES_Configuration.ipf]\rPGC_SetAndActivateControl(...)#L230
 [MIES_ProgrammaticGUIControl.ipf]\rDAP_ButtonProc_TPDAQ(...)#L4296
 [MIES_DAEphys.ipf]\rTPM_StartTestPulseMultiDevice(...)#L122
 [MIES_TestPulse_Multi.ipf]\rTPM_StartTPMultiDeviceLow(...)#L45 
 [MIES_TestPulse_Multi.ipf]\rAssertOnAndClearRTError(...)#L4365
 [MIES_Utilities.ipf]\r","ts":"2023-05-08T15:44:44.638-07:00"}
```

which says in short

```
Encountered pending RTE: 823, PadString;Parameter is out of range
```

It can be from various places but I strongly suspect that it is from UploadLogfiles/FilterLogfileByDate/LoadTextFile where PadString is used.

The reason is probably that the logfile is larger than 2GB and then we can't create an IP string from it.

The solution here is multifold:
- Assert in LoadTextFile when PadString errors out
- Add logfile rotation on a weekly basis

The last step must happen before UploadLogFilesDaily() tries to read the too large logfile.

t-b:
[...]

PR #1748, opened 2023-05-25
Correct handling of large log files

MichaelHuth:
- Make it possible to load > 2 GB log files and performant evaluation
- Add archiving of uploaded logs to chunks.
- Better upload handling in chunks and some progress output
- Use new functions also for log data in NWB files

close https://github.com/AllenInstitute/MIES/issues/1714
close https://github.com/AllenInstitute/MIES/pull/1715


t-b:
Review:

1fddb99a (Util: Add function LoadTextFileToWave that supports > 2 GB files, 2023-05-25)

Missing tests.

a113a908 (Util: Update LoadTextFile function description, noting the < 2 GB limit, 2023-05-25)

Jeep.

c2342d80 (Util: Move JSON Parsing in seperatate function in FilterByDate, 2023-05-25)

Fine.

5bbd611d (Util: FilterLogfileByDate replace LoadTextFile with LoadTextFileToWave, 2023-05-25)

Makes sense.

7354b205 (Rewrite FilterByDate using BinarySearch, 2023-05-25)

That's very terse. FindFirstLogEntryElementByDate/FindLastLogEntryElementByDate should have documentation.

The commit message does also not explain any potential differences of the new implementation vs the old one. As you have removed part of the old test `// non-matching entries are included` this is especially important.

The case

```
9 | 8009 |   | if(!IsNaN(ts))
-- | -- | -- | --
  | 0 | 8010 |   | break
  | 9 | 8011 |   | endif
```

is not covered in the tests for FindFirstLogEntryElementByDate/FindLastLogEntryElementByDate.
Double whitespace after equal sign in `ts =  ParseISO8601TimeStamp`.

a89b81e4 (Util: New LogToStrings that splits a logData text wave to size limited strings, 2023-05-25)

Missing documentation and tests. Is there anything MIES specific in these functions? If not they should be moved to MIES_Utilities.ipf.
Why does LogToStrings not return a wave reference wave? The current approach seems to work, but it's quite confusing and also slow with the TExtWaveToList in LogToStringsPutPart.j

7175c89a (Util: Add function base64EncodeSafe with proper size check, 2023-05-25)
be83f1f7 (ParseISO8601TimeStamp: Add eraly return if argument is empty string, 2023-05-25)

eraly -> early

6e5fc54c (GenerateJSONTemplateForUpload: Add optional argument to set timeStamp, 2023-05-25)

okay

b0c8c808 (UploadLogFiles: Upload in chunks, 2023-05-25)

Bad commit message as it does not explain anything.
The new code ignores the verbose flag. Why?

Add a separate function for calculating the base64 overhead.

```
		fSize = GetFileSize(file)
		if(IsNaN(fSize))
			continue
		endif
```

That drops non-existing files a la `XXX-does-not-exist`. It is also quite expensive to query the filesize just for the print out.

Can we have one upload only for the common case of no splitting?

a3932026 (Remove unused function FilterLogfileByDate, 2023-05-25)

This should be merged with the commit not requiring the function as otherwise future readers don't understand why this function is now obsolete finding this commit.

53c074d3 (Add archiving of uploaded parts of the Log files, 2023-05-26)

ArchiveLogFile must *not* duplicate the custom zeromq flags!!

```
+        endif
+        if(FileExists(newFullFilePath))
```

-> elseif

65470c45 (Revise NWB_AppendLogFileToString to use new FilterByDate, 2023-05-27)

I think we chatted about that, but you can assume that the data from one day is below 2GB.

```
+        now = GetISO8601TimeStamp(localTimeZone = 1)
+        firstDate = ParseISO8601TimeStamp(now[0, 9] + "T00:00:00Z")
```

It's heavily confusing to have `localTimeZone` in one line and `Z` in the other.

47ff67ba (Add constant for log file line ending, 2023-05-27)

Makes sense.

358caed2 (CONF: Make uploading of logfiles on Configuration restore more resilient, 2023-06-01)

Good point, although you might want to clear the RTE before calling BUG.

11754da5 (ASSERT on too large files in LoadTextFile for more resilience, 2023-06-01)
4907b6a0 (Add ASSERTions for RTEs of GetFileFolderInfo in common utility functions, 2023-06-01)

Both are good.

MichaelHuth:
> Why does LogToStrings not return a wave reference wave? The current approach seems to work, but it's quite confusing and also slow with the TExtWaveToList in LogToStringsPutPart.j

I have rewritten the functions such that the splitting function returns a wave reference wave.
If the caller works further with a string as logData TextWaveToList is now called in these callers.
The only way I see to mitigate TextWaveToList calls could be for cases where the resulting data is written to disk. Instead of writing a string through FBinWrite, a text wave through wfprintf.
We could also consider to use wfprintf in TextWaveToList for the case of 1D waves, if that is faster (not tested).

> The new code ignores the verbose flag. Why?

Nothing has changed in the new code regarding the verbose flag.

> Can we have one upload only for the common case of no splitting?

If nothing is split there is only a single json created and uploaded.

t-b:
Needs conflict resultion and rebase for Historic Data CI.

t-b:
Very nice! Thanks a lot for the effort!

t-b (approved):

PR #1761, opened 2023-06-01
Fix CI: Allow install binary to be located in sub-directories

Garados007:
The CI was broken and no installer could be found for the installation step. It is unknown why this has worked before but is broken right now. This fix prevents this bug from happening again.


Garados007:
@t-b You can merge it after the pipeline succeeds. Otherwise, reassign it to me. I will see when I have time for this.

I don't know why these errors are appearing right now. All previous runs seem to work. 🤷

t-b (approved):

Issue #197, opened 2019-06-07
Ping website on MIES device locking to count users


t-b:
- [x] Add new setting "User Ping" to package settings defaults to `on`
- [x] Add toggable menu entry to turn it permanently off
- [x] Add JSON config setting to turn on/off
- [x] Only do at most once a day, don't do in CI
- [x] Upload info on device lock via GenerateJSONTemplateForUpload/AddPayloadEntries/UploadJSONPayload.
- [ ] Clarify if doing it in the startup hook would be better, as we would then also count users which would not have hardware

Uploaded info in addition to the defaults:
- available hardware types
- Full IP version
- MIES version

t-b:
- [ ] Look into technical options to find out the location of each user

timjarsky:
@ t-b tracking users who don't use hardware is a good idea, but it also makes sense to track users who have hardware. I'm unsure how to do this in a way that doesn't result in two pings.

t-b:
@timjarsky The current implementation pings on every IP startup. I think we can handle the additional load.

PR #1762, opened 2023-06-01
Ping website

MichaelHuth:
- only run once per day, not in CI
- Uploads Igor Version incl. BUILD number, Hardware Device List, MIES Version
- Added to StartOrNewHook

close https://github.com/AllenInstitute/MIES/issues/197

### DIFF
--- a/Packages/MIES/MIES_AnalysisBrowser_SweepBrowser.ipf
+++ b/Packages/MIES/MIES_AnalysisBrowser_SweepBrowser.ipf
@@ -403,6 +403,8 @@ Function/DF SB_OpenSweepBrowser([variable mode])
 		ASSERT(mode == BROWSER_MODE_USER || mode == BROWSER_MODE_AUTOMATION || mode == BROWSER_MODE_ALL, "Invalid mode")
 	endif
 
+	UploadPingPeriodically()
+
 	Execute "DataBrowser()"
 
 	mainWin = GetMainWindow(GetCurrentWindow())
--- a/Packages/MIES/MIES_Configuration.ipf
+++ b/Packages/MIES/MIES_Configuration.ipf
@@ -153,6 +153,8 @@ static StrConstant EXPCONFIG_EXCLUDE_CTRLTYPES = "12;9;10;"
 
 static StrConstant EXPCONFIG_SETTINGS_AMPTITLE = "0,1;2,3;4,5;6,7"
 
+static StrConstant EXPCONFIG_JSON_GLOBALPACKAGESETTINGBLOCK = "Global Package Settings"
+
 static StrConstant EXPCONFIG_JSON_HSASSOCBLOCK = "Headstage Association"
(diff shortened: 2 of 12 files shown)


t-b:
Review:

a77bf3ae (Add userping entries to package settings, 2023-06-01)

Jeep.

e6c97167 (Add dynamic menu entry to switch daily ping, 2023-06-01)

MIES_PackageSettings.ipf is meant for generic package management. Therefore all functions in there don't do MIES specific stuff. I therefore think that PS_SwitchUserPingSetting() needs to be in `MIES_MIESUtilities.ipf` as `UpgradeSettings` does also live there. Or we create a new file with MIES specific package settings functions.

934bcd4b (PS: Add Setter/Getter function for upload ping properties, 2023-06-01)

same comment as previous commit.

ea14b87f (CONST: Add SECONDS_PER_DAY constant, 2023-06-01)

Fine.

6b86bd3a (Add UploadPingDaily functionality, 2023-06-01)

UploadPingDaily uses two versions of the current timepoint `now` and `currentTS`. PS_SetUserPingTimestamp should probably accept `now`.

The uploaded entries should be more structured. So instead of multiple files, version.txt should be kept, upload one JSON formatted file. DAP_GetDACDeviceList is also not the right function to call as it does call `DAP_SuggestIssueForAdditionalNIHardware()` if it could not find any hardware. Instead call DAP_GetITCDeviceList/DAP_GetNIDeviceList.

Why do you use `isBinary = 1`?

bbc31ead (PHP: Add saving of user IP in php upload script, 2023-06-01)

Code looks good. I've also test it, and it works.

I have not found the implementation for the checked todo:

> Add JSON config setting to turn on/off

MichaelHuth:
> Why do you use `isBinary = 1`?

I used the same setting as for the logfile upload. I think it is not required here though.

t-b:
> > I have not found the implementation for the checked todo:
> > > Add JSON config setting to turn on/off
> 
> It is added in [a77bf3a](https://github.com/AllenInstitute/MIES/commit/a77bf3ae46fa59c86a269932f2cf14c0719784da)

- I meant **JSON config** not **Package settings**
- Can UploadPing structure the sent data even more? Have `"DAQHardwareDevices": {"NI" : []", "ITC"}` and also use separate objects for IgorInfo(0) and IgorInfo(3). Also we want to not have any igor style KV pairs in the sent json.

t-b:
Add to json config as global package settings:

```
        "Global Package Settings": {
          "userping": {
            "enabled": true
          }
        }

```

(same level as Position MCCs), write default with true via CONF_DefaultSettings(). On new User

t-b (dismissed):

t-b:
@MichaelHuth The menu option and the print outs still say "Daily". Otherwise looks good!

t-b (approved):

t-b:
I'v deployed the new version of the upload script.

Issue #1766, opened 2023-06-01
JSON Configuration does not work with wavebuilder panels


t-b:

MichaelHuth:
The menu calls `CONF_RestoreWindow("", usePanelTypeFromFile = 1)`, but the WB panel has no panelType. With `CONF_RestoreWindow("")` the restore works. Should we introduce a panelType for the WB panel?

PR #1769, opened 2023-06-02
Automatize panel type handling on window configuration restore

MichaelHuth:
- Previously we had the optional parameter usePanelTypeFromFile to hint the restore function that the file should have a panel type and a new window of that type if opened before restore.
- This is changed now, such that this optional parameter is discarded. Instead we check if a panel type information is present in the file and create a new window if so.

close https://github.com/AllenInstitute/MIES/issues/1766

Please note, that the wave builder has no panel type. Thus, for restore no new wavebuilder panel is opened automatically as the function can not know. The wave builder is treated as a generic window, where the user must have that window as front window when attempting the configuration restore.

### DIFF
--- a/Packages/MIES/MIES_AnalysisBrowserMacro.ipf
+++ b/Packages/MIES/MIES_AnalysisBrowserMacro.ipf
@@ -114,6 +114,7 @@ Window AnalysisBrowser() : Panel
 	SetWindow kwTopWin,userdata(ResizeControlsInfo)=A"!!*'\"z!!#EIJ,hu%zzzzzzzzzzzzzzzzzzzzz"
 	SetWindow kwTopWin,userdata(ResizeControlsInfo)+=A"zzzzzzzzzzzzzzzzzzzzzzzzz"
 	SetWindow kwTopWin,userdata(ResizeControlsInfo)+=A"zzzzzzzzzzzzzzzzzzz!!!"
+	SetWindow kwTopWin,userdata(Config_PanelType)="AnalysisBrowser"
 	SetWindow kwTopWin,userdata(JSONSettings_StoreCoordinates)="1"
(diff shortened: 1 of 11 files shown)


t-b:
@MichaelHuth 

> Please note, that the wave builder has no panel type. Thus, for restore no new wavebuilder panel is opened automatically as the function can not know. The wave builder is treated as a generic window, where the user must have that window as front window when attempting the configuration restore.

I think this makes it a non-solution as the user expects the window to be created. Let's chat about this on monday.

t-b:
@MichaelHuth Code looks mostly good, but it does not work as I get:

```
  Message: "Configuration file entry for panel type has an unknown type (WaveBuilder)."
```

When loading a wavebuilder config file. Maybe we also need a test for that?

The test `TCONF_CheckMacrosForPanelType()` should also use data generators as in 

```igorpro
Function/WAVE GetMiesMacrosWithPanelType()
	WAVE/T allMiesMacros = GetMIESMacros()

	Make/FREE/T panelsWithoutType = {"ID_Headstage_Panel", "ID_Popup_Panel", "DP_DebugPanel", "SBE_ExportSettingsPanel"}

	WAVE/T matches = GetSetDifference(allMiesMacros, panelsWithoutType)

	SetDimensionLabels(matches, TextWaveToList(matches, ";"), ROWS)

	return matches
End

/// @brief Checks if every (qualified) panel has a panel type set
///
/// UTF_TD_GENERATOR GetMiesMacrosWithPanelType
static Function TCONF_CheckMacrosForPanelType([string str])

	string win, panelType

	Execute/Q str + "()"
	win = GetCurrentWindow()
	panelType = GetUserData(win, "", EXPCONFIG_UDATA_PANELTYPE)
	INFO("Panel: " + win)
	CHECK_PROPER_STR(panelType)
End
```

t-b (approved):
LGTM.

PR #1780, opened 2023-06-09
Various fixes

t-b:
@MichaelHuth Can you have a look?


t-b:
Needs some more fixes.

Packages/tests/UTF_HelperFunctions.ipf
```diff
@@ -781,3 +781,60 @@ Function/S GetTestStimsetFullFilePath()
...
+
+	if(!loadSweeps)
+		return [abWin, ""]
+	endif
+
+	WAVE/T expBrowserList = GetExperimentBrowserGUIList()
+	WAVE expBrowserSel = GetExperimentBrowserGUISel()
+
+	WAVE/Z indizes = FindIndizes(expBrowserList, colLabel = "file", prop = PROP_NON_EMPTY)
+	INFO("Tryling to load files @%s", s = files)
```
MichaelHuth, line 829: Tryling -> Trying

Packages/tests/UTF_HelperFunctions.ipf
```diff
@@ -781,3 +781,60 @@ Function/S GetTestStimsetFullFilePath()
...
+	PathInfo home
+	REQUIRE_EQUAL_VAR(V_flag, 1)
+
+	Duplicate/FREE/T files, filesWithPath
+	filesWithPath[] = S_path + files[p]
+
+	abWin = AB_OpenAnalysisBrowser(restoreSettings = 0)
+	CHECK_PROPER_STR(abWin)
+
+	for(file : filesWithPath)
```
MichaelHuth, line 814: Missing "string file".

I would also suggest a different naming here, because we have "files" and "filesWithPath".
How about "fullFilePath" ?

Packages/tests/UTF_HelperFunctions.ipf
```diff
@@ -781,3 +781,60 @@ Function/S GetTestStimsetFullFilePath()
...
+		REQUIRE(FileExists(file))
+		MIES_AB#AB_AddElementToSourceList(file)
+	endfor
+
+	MIES_AB#AB_AddExperimentEntries(abWin, filesWithPath)
```
MichaelHuth, line 819: This works for the current tests, but I think generally it would be better to trigger the "Refresh" button instead because then you also get the proper state of the AB GUI.

MichaelHuth (changes requested):
The rest looks good.

Packages/tests/UTF_HelperFunctions.ipf
t-b: Fixed.

Packages/tests/UTF_HelperFunctions.ipf
t-b: Good point. Done!

PR #1781, opened 2023-06-09
.github/workflows/build-pr.yml: Revise NWB packaging

t-b:


Garados007 (approved):
I see no issues with this PR but you will never get the historic data run finished inside this PR. There are only these options:

1. Disable the requirement temporary.
2. Rebase this PR on the commit that contains the historic data run in its PR workflow. 

I also propose replacing the emoji for the historic data run with 📆 (only if its not contained in `TestWithoutHardware`).

t-b:
> I see no issues with this PR.

Nice!

> but you will never get the historic data run finished inside this PR.

I know, I'll rebase.

> I also propose replacing the emoji for the historic data run with 📆 (only if its not contained in TestWithoutHardware).

It is part of HistoricData.

Issue #1595, opened 2022-12-15
Using epochs who's edges are partially outside the acquired data doesn't work.


t-b:
FTP: 2022-03-08%20PSC%20example%20data/2022_03_08_140256.pxp
Sweep 0
SF:

```
data(
"Stimset",
select(channels(AD),sweeps())
)
```

gives

```
Specified ending range not inside sweep 0.
```

This is because the stimset epoch is a bit larger than the acquired data. I don't know why it the case here, but it can always be the case with premature stopping a sweep and using a MIES version prior to 8c6e5da6 (EP_WriteEpochInfoIntoSweepSettings: Handle unacquired data, 2021-07-13).

Instead of complaining should we maybe clip the range? Only for epoch names or in general?

EDIT: Some more info for this case:

![image](https://user-images.githubusercontent.com/1185677/209996302-b8915b43-ff51-496a-9ce5-dd12e817f4e7.png)

So the stimset epoch ends at 12041.67969, but the stimset ends at 12041.64, the sampling interval is 0.04. So in this case the stimset decimation, DAQ with fixed frequency 25kHz, resulted in the deviation. For this case the solution could be to clip the epoch. Or adapt the epoch ranges to sample points on storing.

MichaelHuth:
I already thought this might come up as we have in SF a strict check that the evaluated range from the epochs is within the sweep.

I would say we clip it only for epochs and skip if the result has no data.

@timjarsky What would make sense here?

t-b:
@timjarsky Plesae also assign an appropriate project swim lane (todo/staging) if you think this should be done.

timjarsky:
> I already thought this might come up as we have in SF a strict check that the evaluated range from the epochs is within the sweep.
> 
> I would say we clip it only for epochs and skip if the result has no data.
> 
> @timjarsky What would make sense here?
@t-b @MichaelHuth 
I'm not sure if "clip-and-keep" is the right approach. For example, if DAQ was stopped early, I don't think we'd want to include the epoch with truncated data. So, it may be better to skip the epoch in that sweep. Does that make sense?

t-b:
Solution: Ignore differences of less than the sampling interval and just clip the epoch in SF.

t-b:
[...]

PR #1789, opened 2023-06-16
SF: bugfix: Add epoch range clipping for SFH_GetSweepsForFormula

MichaelHuth:
After
8c6e5da (EP_WriteEpochInfoIntoSweepSettings: Handle unacquired data, 2021-07-13) we corrected the epoch range calculation that epochs are properly clipped to the sweep range. Older data can contain epochs that have ranges beyond the sweep range.

For that data, we needed a workaround that clips epochs or skips outlier epochs:

- epochs outside the sweep range are skipped
- epochs starting within the sweep and ending outside are clipped before retrieving the sweep data in that range.

Note: the original old epoch data in the LBN is not touched.

close https://github.com/AllenInstitute/MIES/issues/1595

### DIFF
--- a/Packages/MIES/MIES_SweepFormula_Helpers.ipf
+++ b/Packages/MIES/MIES_SweepFormula_Helpers.ipf
@@ -501,8 +501,8 @@ Function/WAVE SFH_GetSweepsForFormula(string graph, WAVE range, WAVE/Z selectDat
 				rangeEnd = limit(rangeEnd, -inf, lastx)
 			endif
 
-			SFH_ASSERT(rangeStart == -inf || (IsFinite(rangeStart) && rangeStart >= leftx(sweep) && rangeStart < rightx(sweep)), "Specified starting range not inside sweep " + num2istr(sweepNo) + ".")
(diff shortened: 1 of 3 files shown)


t-b:
@MichaelHuth Is that ready for review?

t-b:
Review:

- Comment "we did not capped" -> "we did not cap"
- Is `rangeStart >= rightx(sweep)` really correct? Why not check for `rangeStart >= lastx` which would throw away all epochs which start at the last point, those would be empty anyway.
- Can you add a SF test for this pxp as historic data test? UTF_HistoricData.ipf has the instructions. If something is unclear please ask, this is pretty new.
- The change `rangeEnd > leftx(sweep)` in the assertion is hidden, if that is justified please move it into its own commit.

MichaelHuth:
Remarks for UTF_HistoricData.ipf setup of historic data:
- calling CompressFile does ASSERT because it calls GetInputPath that uses an IUTF assertion, which works only when run through a test run
- The mentioned GetFiles() function does not exist, I guess GetHistoricDataFiles should be adapted.
- Mention the "MIES-HistoricData" sub folder for the ftp upload

t-b:
THanks for the remarks and the update.

t-b (approved):

Issue #1779, opened 2023-06-08
Improve and speed up TextWaveToList


MichaelHuth:
Since IP9 wfprintf has no size limit for elements of a text wave.

`NumericWaveToList` already uses the wfprintf approach and we can add that too for `TextWaveToList` for the 1D and 2D case for IP9+

The idea is to move the core function of `NumericWaveToList` to an implementation function that accepts a format string, in our case either "%g" or "%s". The function is then called from `NumericWaveToList` and `TextWaveToList`. (Only for IP9+, 1D and 2D waves and if stoponempty, maxElements is not set)

PR #1790, opened 2023-06-19
Speed up TextWaveToList using wfprintf

MichaelHuth:
- for IP9+ because wfprintf only supports long strings since
- NumericWaveToList and TextWaveToList use now the same functions
- for 1D and 2D waves only

close https://github.com/AllenInstitute/MIES/issues/1779

### DIFF
--- a/Packages/MIES/MIES_SweepFormula_Helpers.ipf
+++ b/Packages/MIES/MIES_SweepFormula_Helpers.ipf
@@ -852,7 +852,7 @@ Function [WAVE/T keys, WAVE/T values] SFH_CreateResultsWaveWithCode(string graph
 
 	WAVE/Z selectData = SF_ExecuteFormula("select()", graph, singleResult=1, useVariables=0)
 	if(WaveExists(selectData))
-		values[0][%$"Sweep Formula sweeps/channels"][INDEP_HEADSTAGE] = NumericWaveToList(selectData, ";")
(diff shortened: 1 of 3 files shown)


MichaelHuth:
It turns out that NumericWaveToList has the row/column separator swapped for 2D waves. The test for the function checks against the swapped order :-/

At least ExportIntoNWBSweepBySweep and TestSweepFormulaCodeResults from the Hardwaretests fail, if I swap it to the correct order. For now I kept the different order of the separators, but we should discuss that in our next chat.

t-b:
Review:

46b35d75 (Bugfix: TextWaveToList not adding separator properly on single element waves, 2023-06-20)

Nice one!

b904dddd (Util: Speed up TextWaveToList using wfprintf, 2023-06-20)

The commit message says

"for IP9+ because wfprintf only supports long strings since" but I would think this needs to be reworded as you have remove "#ifdef IgorVersion"

The fix for NumericWaveToList falls a bit under the rug here in the speeding up. CAn we have a dedicated commit fixing NumericWaveToList and after that make it faster. This commit would then also include the next one.

ebebf237 (Workaround for previous NumericWaveToList bug, 2023-06-20)

See above.

b5143a49 (Util: Speed up NumericWaveToList, 2023-06-20)

Nice!

t-b:
Nice. Thanks.

t-b (approved):

PR #1800, opened 2023-07-05
Bugfix AcquireData_NG ignored TTL channels 4-7 in setup for NI HW

MichaelHuth:
Also adapted the TTL test UnassociatedChannelsAndTTLs to check against the correct setup.

### DIFF
--- a/Packages/tests/UTF_HardwareHelperFunctions.ipf
+++ b/Packages/tests/UTF_HardwareHelperFunctions.ipf
@@ -1480,9 +1480,9 @@ Function AcquireData_NG(STRUCT DAQSettings &s, string device)
 		endif
 
 		if(i >= NUM_ITC_TTL_BITS_PER_RACK                 \
-		   && (GetHardwareType(device) == HARDWARE_NI_DAC \
-		       || HW_ITC_GetNumberOfRacks(device) < 2))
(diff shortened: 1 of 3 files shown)


Packages/tests/UTF_HardwareHelperFunctions.ipf
```diff
@@ -1479,8 +1479,8 @@ Function AcquireData_NG(STRUCT DAQSettings &s, string device)
 			continue
 		endif
 
-		if(i >= NUM_ITC_TTL_BITS_PER_RACK                 \
-		   && (GetHardwareType(device) == HARDWARE_NI_DAC \
+		if(i >= NUM_ITC_TTL_BITS_PER_RACK                  \
+		   && (GetHardwareType(device) == HARDWARE_ITC_DAC \
 		       || HW_ITC_GetNumberOfRacks(device) < 2))
```
t-b, line 1484: `||` -> `&&`? We want to skip the TTL channels if we have:
- a ttl bit of rack one
- and itc hardware
- and only one rack

MichaelHuth: good catch

t-b (approved):

Issue #1763, opened 2023-06-01
JSON Configuration: Allow configuration of unassociated AD/DA channels


t-b:
Test "UnassociatedChannelsAndTTLs"

MichaelHuth:
I have tested around and it seems to work properly.

I ran the test, then saved the DAEphys configuration, closed it and restored the configuration.

Then I acquire one sweep as the test does and the result is identical.

Here the recipe:
```
runwithopts(testcase="UnassociatedChannelsAndTTLs")
CONF_SaveWindow("C:projects:daephys.json")

```
close the panel

`CONF_RestoreWindow("C:projects:daephys.json", usePanelTypeFromFile=1)`

Acquire again.

@t-b if that works for you too, I would say we can close this issue.

timjarsky:
@t-b @MichaelHuth I can confirm this works.

timjarsky:
@t-b  I ran into this again when testing TP insertion into unassociated DA channels. I'm finding that the amplifier hardware settings are getting instantiated on configuration. So, when checking DA2, for example, headstage 2 gets activated rather than only (unassociated) DA2. This happens despite having cleared the associations for that headstage: 
![image](https://github.com/AllenInstitute/MIES/assets/5693697/61f22e00-cb5f-4fc8-9405-cd7019cad26b)

I also see a potentially related issue with `save window configuration` where MIES asks the user to select an existing file rather than save it with a new name.

t-b:
There is something odd with the config for unassoc.

> I also see a potentially related issue with save window configuration where MIES asks the user to select an existing file rather than save it with a new name.

I saw that as well, moved to #1801.

PR #1804, opened 2023-07-06
CONF: Clear HS association on DAEphys window restore for HS w/o amp

MichaelHuth:
- Previously the on restore the amplifiers were set plain by headstage. The default configuration of the DAEphys panel is that unless "Clear Associations" is clicked all HS are associated. Also MIES implements that associated HS need to have an amplifier set.

  On saving the DAEphys window there is no GUI control that state explicitly if a HS is associated or not.

  Solution: We assume that the headstages without a set amplifier are unassociated and clear the association for these HS.

since introduction of configuration module

close https://github.com/AllenInstitute/MIES/issues/1763
close https://github.com/AllenInstitute/MIES/issues/1811

### DIFF
--- a/Packages/MIES/MIES_Configuration.ipf
+++ b/Packages/MIES/MIES_Configuration.ipf
@@ -510,7 +510,7 @@ Function CONF_RestoreWindow(string fName, [string rigFile])
 		if(ClearRTError())
 			ASSERT(0, errMsg)
 		else
-			printf "Configuration restore aborted at file %s.", fName
+			printf "Configuration restore aborted at file %s.", fullFilePath
 			Abort
 		endif
 	endtry
--- a/Packages/MIES/MIES_Configuration.ipf
+++ b/Packages/MIES/MIES_Configuration.ipf
@@ -680,18 +680,32 @@ Function/S CONF_RestoreDAEphys(jsonID, fullFilePath, [middleOfExperiment, forceN
 		winHandle = num2istr(GetUniqueInteger())
 		SetWindow $device, userdata($DAEPHYS_UDATA_WINHANDLE) = winHandle
 		isTagged = 1
-		device = CONF_JSONToWindow(device, restoreMask, jsonID)
-		isTagged = 0
-		SetWindow $device, userdata($DAEPHYS_UDATA_WINHANDLE) = ""
+
+		WAVE/T winNames = CONF_GetWindowNames(jsonID)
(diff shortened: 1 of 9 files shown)


timjarsky:
@t-b Based on our chat yesterday, may I be unassigned from this PR for now?

t-b:
d83d8743 (Tests: Fix of FixupJSONConfigImplRig to replace only existing amp serials, 2023-07-07)
f121edca (CONF: Clear HS association on DAEphys window restore for HS w/o amp, 2023-07-06)

Already reviewed.

186ddcd3 (PGC: Add optional argument noRangeCheck to PGC_SetAndActivateControl, 2023-07-21)

No tests because??? I'm also not convinced that this is the right approach. Why not teach the popup menu control have `NONE` instead?

4f13d36e (CONF: For DAEPhys save DA/AD gain/unit/channels for all active HS, 2023-07-21)

Missing tests. The changed handling of the title entry should also be its own commit.

0ba284d7 (CONF: For DAEPhys load the AD/DA channels are restored for all active Headstages, 2023-07-21)

I also don't understand

>     - if originally no DA/AD channel was set, then NaN was saved as channel number,
>      this is restored properly to an unset channel number utilizing the noRangeCheck
>      argument of PGC_SetAndActivateControl
>      The same approach as in DAP_UpdateChanAmpAssignPanel is applied.

does it make sense to write NaN as DA/DA channel number out?

The following window configuration was stored with HS 7 active and no DA/AD channels associated. And loading that as config fails with

```
  (Dev1) Please select a valid DA and AD channel in "DAC Channel and Device Associations" in the Hardware tab.
  Configuration restore aborted at file .
```

Note the missing filename.

[NoAmplifier.zip](https://github.com/AllenInstitute/MIES/files/12123713/NoAmplifier.zip)

Missing tests and please also check the wording in the commit message.

c71f89b1 (DAEPhys: Increased width of AD/DA gain setVar fields, 2023-07-21)

Looks good.

I also saw that the hiding and restoring of the DAEphys panel has changed in this PR. Now it is shown when the controls are configured on load.

t-b:
@timjarsky Feel free to test already as the changes Michael needs to do are not functionality wise.

timjarsky (dismissed):

t-b:
@MichaelHuth Please rebase and assign me.

t-b (approved):

Issue #1735, opened 2023-05-22
Release 2.7


t-b:
With the newly added GH actions CI in #1656 it would now be nice to also have CI on the release branch. The release/2.6 branch is based on Release_2.6_20220606. This MIES version does not yet have the enhanced test layout from #1475. So backporting the GH actions CI to that release branch would be quite an effort.

So let's start to think about the next release and what should be part of it.

PRs:

- [x] #1700 
- [x] #1694 
- [x] #1691 
- [x] #1672
- [x] #1731 
- [x] #1732
- [x] #1715
- [x] #1689
- [x] #1747
- [x] #1748
- [x] #1746
- [x] https://github.com/AllenInstitute/MIES/pull/1772
- [x] https://github.com/AllenInstitute/MIES/pull/1791
- [x] https://github.com/AllenInstitute/MIES/pull/1804
- [x] https://github.com/AllenInstitute/MIES/pull/1807
- [x] #1468
- [x] https://github.com/AllenInstitute/MIES/pull/1822

Issue:

- [x] #1753
- [x] #1595
- [x] https://github.com/AllenInstitute/MIES/issues/1836

Once all that is done:
- [x] Write release notes and create PR
- [x] Revise release instructions as we are not using bamboo anymore
- [x] Merge PR
- [x] Tag
- [x] Check that CI works on release branch
- [x] Ping Rusty and Brian Lee about new release. Tell them about configuration upgrade path.

PR #1839, opened 2023-08-04
Prepare Release 2.7

t-b:
- [x] Add changes from branch feature/1468-sweepformula-epsp
- [x] Revise release instructions as we are not using bamboo anymore
- [x] Read over it again
- [x] Cleanup commits
- [x] Check if version string in documentation is correct
- [x] Test installer and check it a bit including version string
- [x] Check if all installed files are needed (disclaimer: no)

See #1735


t-b:
Will merge once CI passes.

PR #1860, opened 2023-08-25
Expose psx deconv filter parameters

### LINKED ISSUE #1855: PSX: Expose filter parameters in PSX_DeconvoluteSweepData
- [x] Add wrapper operation like psxrisetime
- [x] Use plain Hz instead of fractional sampling frequency
- [x] Early version for TJ to playaround
- [x] Default to current defaults
- [x] Do we have to check that the number of filter coefficients are odd?

Additional:
- [x] Use check function `checkFunc = IsStrictlyPositiveAndFinite` for filterLow and filterHigh

t-b:
Close #1855.


t-b:
@timjarsky Ready for a first try.

The new parameters:

```
psx(quack,$kernel, 2.5, 550, 0) # , 10, psxRiseTime(), psxDeconvFilter())
psxDeconvFilter(low, high, order)
```

I've done some hoops to default to the exact old values for the deconvolution filtering.

timjarsky:
@t-b When applying very different low and high values to psxDeconvFilter(low, high, order), I do not see any changes in the deconvolved trace. It looks like changes to psxDeconvFilter don't force an update to psx. It seems like it should, no?

t-b:
@timjarsky Thanks for testing. Indeed that was still broken. I've commited a fix.

timjarsky:
@t-b updating the SD param in psx kernel isn't updating the output of psx.

timjarsky:
@t-b, do we have a list of param that force an update to psx and those that don't?

t-b:
Sorry for the late reply.

> @t-b updating the SD param in the psx operation isn't updating the output of psx.

I can't reproduce that. Using the file `2022_03_08_140256.pxp` I've always used and

```
sel=select(channels(AD9),[31,32,35,37,39],all) 
kernel = psxKernel(E0,$sel, 0.45, 1, (-6))
sd=1.5

psx(quack,$kernel, $sd, 550, 0, 10, psxRiseTime(), psxDeconvFilter())

and

psxstats(quack, E0, $sel, tau, all, count)
```

I get 2567 events with SD = 1.5 and 1321 with 2.5.

> do we have a list of param that force an update to psx and those that don't?

Kind of. PSX_OperationSweepGathering is responsible for generating or fetching sweepDataFiltOff/sweepDataFiltOffDeconv. It get's a JSON document with all psx parameters generated in PSX_Operation. From the JSON we then filter in PSX_GetPSXParameters which parameters are used for generating the cache key and which not.

timjarsky (approved):

timjarsky:
@t-b i confirmed that the failure to update was because psxPrep was not provided an updated SD parameter.

t-b:
@timjarsky Ready again. Now with documentation.

timjarsky (approved):

PR #1866, opened 2023-08-31
DAP_UpdateChanAmpAssignPanel: Don't use PGC_SetAndActivateControl for control setting

### LINKED ISSUE #1868: Pressure controls are messed up when changing headstage
From Tim:

> All the pressure devices for each headstage are set to the last pressure device listed in the updated config file. After manually correcting the auto-updated config file, every second headstage still has an incorrect pressure device.

>  I think I figured out at least part of the issue, the pressure controls don't update when a different headstage is selected in the DA ephys panel hardware tab

t-b:
This reverts commit c17cefc94 (DAP: Use PGC_SetAndActivateControl in DAP_UpdateChanAmpAssignPanel, 2023-07-24).

The reason this does not work is that using PGC calls also the GUI control procedure. But we don't want to call the GUI procedure here as we are syncing from CAA wave to the GUI.

Found as setting pressure devices doesn't work anymore. The hardware pressure controls and the associations controls share the same GUI control procedures DAP_SetVarProc_CAA and DAP_PopMenuProc_CAA and thus using PGC on associations control also triggers a pressure control update.

@MichaelHuth The reverted commit introduced in #1804 broke the pressure device selection. Do you remember why the commit was required?

`runwithOpts(testsuite = "UTF_ConfigurationHardware.ipf")` passes in this branch locally as well.

Close #1868

### DIFF
--- a/Packages/MIES/MIES_DAEphys.ipf
+++ b/Packages/MIES/MIES_DAEphys.ipf
@@ -4825,6 +4825,8 @@ static Function DAP_UpdateChanAmpAssignPanel(device)
 
 	HeadStageNo = str2num(GetPopupMenuString(device,"Popup_Settings_HeadStage"))
 
+	// function does not use PGC_SetAndActivateControl on purpose, see #1872
+
 	// VC DA settings
 	channel = ChanAmpAssign[%VC_DA][HeadStageNo]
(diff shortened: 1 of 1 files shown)


MichaelHuth:
This was changed to get rid of directly using the GUI operations. The CAA feedback was circumvented by working on a copy of the CAA wave. Though it looks like there are more dependencies.

t-b:
Add comment to code why it the way it is.

t-b:
@timjarsky Ready!

timjarsky (approved):

Issue #1867, opened 2023-08-31
Initial NWB export takes forever when having a large ZeroMQ logfile


t-b:
We have a report that configuring the DaEphys panel takes a couple of minutes with the new MIES version (2.7).

I made a shot in the dark and renamed the 2.4GB zeromq logfile. And now the issue is gone.

Maybe we should archive the unintereseting parts of the logfile in NWB_AppendLogFileToString?

@MichaelHuth Would that make sense? If yes can you take a stab at this?

t-b:
Asked reporter about sample logfile.

t-b:
@MichaelHuth The FTP has in the folder `issue-1867` a sample logfile (zipped).

t-b:
@MichaelHuth Did you look into the posted log file?

MichaelHuth:
Yes. The average size increase is 2.7 MB/day.

t-b:
Thanks!

PR #1870, opened 2023-09-01
Archive logfile(s) on successful device locking

MichaelHuth:
- if there are large log files the nwb export takes long as the whole logfile is always loaded. Archiving the logfile shortens the logfile length to logdata from the same day.

close https://github.com/AllenInstitute/MIES/issues/1867


t-b:
Code looks good. Please have a look at CI and the log file I posted in the issue.

t-b (dismissed):

t-b:
Works as advertised!

Review:

e33fa667 (Util: Use wave getter function for retrieving log file names, 2023-09-05)

Good.

50eef3f7 (Util: Factor out splitting of ISO9660 timestamp into components, 2023-09-05)

Nice!

d07a9a43 (DAP: Archive log files on successfull locking of a device, 2023-09-05)

- The commit message does not mention why we are doing this.
- ARCHIVE_ONCE should be defined at CalledOnceNames. And maybe AlreadyCalledOnce should have

```
/// 
/// @param name identififer, see also @ref CalledOnceNames
```

in the documentation as that is not obvious.

Can we move DAP_ArchiveLogFilesOnLocking to MIES_MiesUtilities.ipf and maybe
rename? Keeping ArchiveLogFile static sounds like a good idea.

t-b (changes requested):
See comment.

t-b:
Very nice! And the revised commit message is also a well done writeup.

t-b (approved):

PR #1887, opened 2023-09-08
.github/workflows: Cancel redundant workflow runs

t-b:
Evern since the move to github actions in 70daa61b (CI: add GH actions for main and release builds, 2023-04-20) we just executed the workflows for every branch push.

This can quickly result in the workflow queue being clogged. There is also no point in executing older versions of the same branch when a newer version is available.

The version here supports multiple workflows [1]. And also cancels, on purpose, redundant workflows on main and the release branches. See [2] for a future way to don't do that on special branches.

[1]: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-only-cancel-in-progress-jobs-or-runs-for-the-current-workflow
[2]: https://stackoverflow.com/questions/68418857/how-to-cancel-existing-runs-when-a-new-push-happens-on-github-actions-but-only

### DIFF
--- a/.github/workflows/build-pr.yml
+++ b/.github/workflows/build-pr.yml
@@ -31,6 +31,9 @@ on:
 defaults:
   run:
     shell: bash
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
 jobs:
 
   BuildInstaller:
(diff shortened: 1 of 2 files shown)


t-b:
Will merge once CI passes.

PR #1892, opened 2023-09-12
SF_GetDefaultFormula: Don't use cursors

t-b:
The initial SweepFormula design had a prominent place for the cursors
operation. The idea was to allow users to extract data from arbitrary
ranges. But as soon as we had epoch support back in 9cfbd57 (SweepFormula
add epochs command, 2021-09-03) we realized that one wants to usually just
extract an epoch and not an arbitrary range.

The special cases which is exploited in the cursors(A, B) code is that
when there are no cursors present it returns the whole range.

But we can as easy return the whole range via [0, inf].


t-b:
Fixed conflicts.

timjarsky:
@t-b in this branch 

```
print SF_GetDefaultFormula()
```
returns
```
trange = [0, inf]
sel = select(channels(AD), sweeps())
dat = data($trange, $sel)
```

but both the data and sweep browser SF windows still display the old formula.

```
trange = cursors(A,B)
sel = select(channels(AD),sweeps())
dat = data($trange, $sel)

$dat
```

t-b:
@timjarsky 

> but both the data and sweep browser SF windows still display the old formula.

Thanks for testing. Fixed.

timjarsky (approved):

Issue #1725, opened 2023-05-18
Logfile bugs:  "Missing DAScale stepsize LBN entry"


t-b:
Example:
```
{
  "action": "report",
  "exp": "C57BL6J-629713.05.01.02",
  "id": "99ac17d",
  "msg": "Missing DAScale stepsize LBN entry",
  "source": "BUG",
  "stacktrace": "Stacktrace:
   AD_CheckProc_Toggle(...)#L1221 [MIES_AnalysisFunctions_Dashboard.ipf]
   AD_Update(...)#L61 [MIES_AnalysisFunctions_Dashboard.ipf]
   AD_FillWaves(...)#L284 [MIES_AnalysisFunctions_Dashboard.ipf]
   AD_GetResultMessage(...)#L169 [MIES_AnalysisFunctions_Dashboard.ipf]
   AD_GetSquarePulseFailMsg(...)#L413 [MIES_AnalysisFunctions_Dashboard.ipf]
   ",
  "ts": "2022-11-30T14:15:41.375-08:00"
}
```

Experiments:
```
"C57BL6J-629713.05.01.02"
```

Files:
```
.\W10DT1120232-07-svc-ivscc-2023-02-01T16_08_29Z\Log.jsonl.bPpYJf
```

Code:
```
Packages/MIES/MIES_AnalysisFunctions_Dashboard.ipf=static Function/S AD_GetSquarePulseFailMsg(numericalValues, sweepNo, headstage)
Packages/MIES/MIES_AnalysisFunctions_Dashboard.ipf:      BUG("Missing DAScale stepsize LBN entry")
```

t-b:
@ru57y34nn Can you get the experiment file "C57BL6J-629713.05.01.02"?

t-b:
And again with adaptive testing:

FTP: `AdaptiveSupraPXPs01282025/Sst-IRES-Cre (1).pxp`

PR #1894, opened 2023-09-13
PSQ_DAScale: Add adaptive suprathreshold

### LINKED ISSUE #1532: Add new suprathreshold mode to PSQ_DAScale; spike frequency monitor [Adaptive Threshold]
The existing suprathreshold mode gives fixed, user-defined, current injection values above the previously determined rheobase (PSQ_Rheobase). I want to add a suprathreshold mode that monitors spike frequency and continues to increase the injected current until the spike frequency delta falls below a user-defined parameter. 

Typically, cell spiking responses rise with current injection, but eventually increasing the amount of injected current doesn't change the spike frequency. The goal is to have the analysis function stop injecting current at or near this threshold. 

Example I-F plot showing the plateau in spike frequency with increasing current injection (The experimental conditions are not important here).
image

t-b:
Close #1532.

- [x] Document new SF operations and flesh out into their own PR
- [x] Implement passing supra sweep checks
- [x] PSQ_DS_GatherFrequencyCurrentData should use the start position from supra but the length from the E1 epoch in adaptive
- [x] epoch E1 in adaptive needs to be shorter that in supra
- [x] ~Move code from PRE_SET_EVENT to POST_SWEEP_EVENT as we don't yet have epoch info for PSQ_DS_GatherFrequencyCurrentData~
- [x] Remove WIP commits
- [x] Discuss better DASCale estimation with Tim, integrate properly, adapt tests
- [x] Add new tests for new optional parameters, adapt test matrix

Todo List from testing on live neuron 2024/02/29:
- [x] Investigate dashboard not updating. Can't reproduce 
- [x] Add more debug output for overshoot correction. Especially for the cases where it does *not* add new DAScale entries.
- [x] ~~Investigate why "dascales left"/"f-I slopes" in the textual LBN looks like single FP data. Partial explanation: The DAScale values ("Stim Scale Factor") read from the supra sweeps' LBN already have the looks-like-SP property, e.g "119.9999988529321"~~ seems to be by-design supra uses `spikeFrequencies[] = str2num(FloatWithMinSigDigits(spikeCount[p] / (pulseDuration[p] * MILLI_TO_ONE), numMinSignDigits = 2))`
- [x] Gather passing sweeps from long rheobase SCI
- [x] Calculcate daScaleminStepWIdth/dascaleMaxStepWith and store in LBN
- [x] Use normalized DAScale step width in overshoot detection
- [x] Use normalized dAScale step widths in PSQ_DS_CalculateDAScale
- [x] Fix tests
- [x] Update test matrix
- [x] Cleanup commits
- [x] Investigate dashboard error "Failure" which does not say why, moved to https://github.com/AllenInstitute/MIES/issues/1725
- [x] Index out of range error in DAQ, see 2024_04_25_091659-compressed.nwb
- [x] cleanup diff from live testing
- [x] check why we don't get an automation browser, https://github.com/AllenInstitute/MIES/pull/2100
- [x] check if we used the hideously large frequencies, no they are not. E.g. sweep 261 has frequency of ~114 but the stimscale factor stays the same for the next sweeps.
- [x] don't do apfrequency when BL QC is failing
- [x] Integrate ad-hoc changes from last thursday
- [x] Flowchart is made for aaca2db3 (PSQ_GetLastPassingDAScale: Don't overwrite key inside loop, 2023-12-12)

<details>
<summary>Original todo list</summary>

Thirty six hints for writing analysis functions
===============================================

- [x] Create an issue with the description of the new analysis function.
   Also define its name and two letter abbreviation (which must not collide with existing ones).

- [x] Always track *any* open items in task lists in the issue or the PR itself.

- [x] Think about what should happen at each event. As a rule of thumb we want
   to do only what is really needed in ``MID_SWEEP_EVENT`` (if that is used at
   all). The common QC entries are determined in the following events:

   - Sweep QC in ``POST_SWEEP_EVENT``
   - Set QC in ``POST_SET_EVENT``
   - Baseline QC in ``MID_SWEEP_EVENT``/``POST_SWEEP_EVENT``.

- [x] Create a list of used labnotebook keys. The question which labnotebook keys
   are required can be answered by thinking about how the dashboard will interpret
   the analysis function run results. Only if it can decide, by looking at
   labnotebook entries, for each possible outcome exactly why the run failed, all
   labnotebook keys were thought of.

- [x] Create a list of used user epochs. User epochs are there to define
   interesting stimset x-ranges for the analysis function. This should be used
   preferrably over other similiar approaches.

- [x] Create a list of analysis function parameters including required/optional
   state and for the latter also the default values.

- [x] Units for labnotebook keys should be added if possible. For physical units we
   tend to prefer base units without prefix, i.e. Ω instead of GΩ.

- [x] Decide if the new labnotebook entries should be headstage dependent or not.
   The existing entries don't do a very good job in guiding you here. An
   ideal choice is that the ``DEPEND``/``INDEP`` type of a entry would not
   have to be changed if the analysis function would need to support more or
   fewer headstages.

- [x] Make a list of additional features and/or changes in common ``PSQ_``/``MSQ_`` functions you
   need.

- [x] Draw a preliminary flowchart, on paper is fine. This serves as a way to think the behaviour through.
   Have a look at existing flowcharts for inspiration.

- [x] Create a stimulus set for testing. The test stimsets can be loaded via
   ``LoadStimsets`` and saved via ``SaveStimsets`` available in
   HardwareTests.pxp.

- [x] At this point you should have a pretty good idea what needs to be done.
   Discuss what you think you need to do with your boss.

- [x] Add a skeleton analysis function, see `here <https://alleninstitute.github.io/MIES/file/_m_i_e_s___analysis_functions_8ipf.html>`__,
   and add all analysis parameters, their help messages and check code.

- [x] Add documentation for labnotebook keys and user epochs to the tables at the
   top of ``MIES_AnalysisFunctions_PatchSeq.ipf``/``MIES_AnalysisFunctions_MultiPatchSeq.ipf``

- [x] Implement the test override entries in
   :cpp:func:`PSQ_CreateOverrideResults`/:cpp:func:`MSQ_CreateOverrideResults` with
   documentation.

- [x] Implement the behaviour for each event. Going from easy to difficult has proven to work.

- [x] Now you should have a first version. Congratulation! Pad yourself on the
   back and take a break, because now the real fun starts.

- [x] Add preliminary dashboard support. We do check for every testcase that
   the dashboard works.

- [x] Create a new test suite and add it to ``UTF_HardwareAnalysisFunctions.ipf``. Be sure to
   base it on the test suite of the last added analysis function to avoid copying
   deprecated approaches.

- [x] Add a first test case were all test override entries result in failed QC.

- [x] As rule of thumb what to check in each test case; be sure to have test
   assertions for all added labnotebook entries (except standard baseline
   entries) and position checking of user epochs.

- [x] Make that first test case pass, this takes a surprisingly long time. The
   function :cpp:func:`LBV_PlotAllAnalysisFunctionLBNKeys` helps for debugging.

- [x] After this first test case passes, reassess the test assertions. Are you testing enough or too much?

- [x] Writeup a test matrix for determining what needs to be
   tested, first version in paper is fine. The columns are the inputs,
   usually these are test overrides and analysis parameters.

   See ``UTF_PatchSeqSealEvaluation.ipf`` for an example.

   We always have the following three test cases:

     - The first has all QC (except ``Sampling Interval QC``) failing.
     - The second has all QC passing.
     - The last one only has ``Sampling Interval QC`` failing.

- [x] Implement all test cases, fixing bugs in the analysis function on the way.

- [x] Run all tests for this analysis function with instrumentation.

- [x] Check the coverage output to see if you still have relevant gaps in testing.

- [x] Add new test cases for filling coverage gaps.

- [x] Repeat the last three points until the coverage is good enough.

- [x] Check if some helper code etc. can/should be fleshed out into its own pull request.

- [x] Be sure to include documentation and tests if your analysis function
   publishes ZeroMQ messages. See ``CheckPipetteInBathPublishing`` and
   :cpp:func:`PSQ_PB_Publish` for an example. Add it also in
   ``CheckPublishedMessage``.

- [x] Tell your boss to test the current state.

- [x] Check and fill any gaps in the documentation:

   - Analysis function comment with ascii art stimulus sets
   - Labnotebook entries
   - User epochs

- [x] Create digital versions of the test matrix and the flowchart. For the
   latter see `here <https://github.com/AllenInstitute/MIES/tree/main/Packages/doc/dot#readme>`__.

- [x] Cleanup commits

- [x] Your done!

</details>


t-b:
@timjarsky I've completed the fix from our chat session last time.

t-b:
@timjarsky Ready for another try.

timjarsky:
> @timjarsky Ready for another try.

Totally missed this message. I'll get on it.

t-b:
Adaptives sweeps looking into:

0th SCI:
  Sweep 17 passed, f-I slope = 0.0

```
s = range(11, 18)
ep = E1

apfrequency(data($ep, select(channels(AD), $s, all)))

vs

mean(data($ep, select(channels(DA), $s, all)))

and

labnotebook("USER_DA Scale f-I slope", select(channels(DA), $s, all), UNKNOWN_MODE)
```

grafik

It looks odd that the calculated f-I slope is so vastly different from what `apfrequency` calculates.

t-b:
@timjarsky Ready for action!

t-b:
And ready again.

t-b:
And ready!

t-b:
And ready!

timjarsky (approved):

PR #1908, opened 2023-09-28
Propagate changed stimsets better

### LINKED ISSUE #1907: wavebuilder "combine" not working with multiple addition
Recipe:

Create ten sin wave stimulus sets at different frequencies. 1 5s sweep per stimulus set.  Add them all together using the combine functionality, e.g., A+B+C......  Wavebuilder fails to add more than 4 stimulus sets together. Larger formulas do nothing.

t-b:
Since the introduction of combine epochs for the wavebuilder in 9b80987
(WaveBuilder: Add new epoch type Combine, 2015-11-12) we need to maintain
a wave with the shorthand<->full name mapping. This mapping is updated in
WB_UpdateEpochCombineList.

Now that function was not called when loading stimsets from NWB and not
after starting the wavebuilder panel which resulted in errors trying to
recreate combine epochs.

This is now fixed. In order to make the fix less hacky we also introduced
WB_UpdateChangedStimsets which also calls
DAP_UpdateDaEphysStimulusSetPopups which needs to be called in the same
cases.

Close #1907.

@timjarsky The issue is a plain list updating issue. The stimset itself is good.


timjarsky:
@t-b I'm getting:
```
Stimset : Could not create the wave from formula of version 1
  WB_FillWaveFromFormula: Error executing the formula "ten_da_0?+twenty_da_0?+forty_da_0?+eighty_da_0?+onesixty_da_0?+threetwenty_da_0?"
```

t-b:
@timjarsky Please try again. I do need a demo if it is still broken.

timjarsky:
@t-b after inducing the issue, then trying combine a single stimset, I get:

``` WB_FillWaveFromFormula: Error executing the formula "sixforty_da_0?"
  Stimset : Could not create the wave from the formula** Make gave error: the number of points in a wave must be between 0 and 214700 million.
```

t-b:
Bug is present if one stimset is named like the end of another stimset. The function `WB_FormulaSwitchToShortHand` does buggy replacement.

t-b:
@timjarsky Please try again.

t-b:
Tests should be fixed.

t-b:
Tests are *still* failing.

timjarsky (approved):

PR #1921, opened 2023-10-21
Fix: Testpulse calculation failed if TP data is NaN

MichaelHuth:
In the case the test pulse data is completely NaN WaveStats can not determine V_minRowLoc and V_maxRowLoc and returns a -1 in both. This index was used unchecked and resulted in a RTE for invalid wave index access.

This is now checked properly and NaN is returned for "instantaneous".

since
c5ed1686 (TP_TSAnalysis: Change the way we calculate the instantaneous resistance, 2022-09-02)


MichaelHuth:
@t-b I could use some input here, as I can not reproduce the failing test with NI hardware. It does not seem to run in the TP analysis function.

t-b (approved):

t-b:
@MichaelHuth I don't think the test failure is real. I've approved and restarted the test.

PR #1923, opened 2023-10-24
Make vs per formula

### LINKED ISSUE #1627: Sweepformula Using "with" keyword and different x-axis -> only last x-axis is setup
When using sweepformula with the "with" keyword and e.g. two formulas where the operation return different x-axis then only the x-axis of the last formula is setup in the plot.

If that is implemented I think the best option is to enable support for that for maximum of two different x-axes and set the second one as top axis. This also needs then a change of plot markers (single point data) to enable a visual difference between the data points that allows the viewer to attribute the data to the respective axis. The decision key could be the x-axis label (wave unit).

Three or more different x-axis will get complicated and at some point also confusing. Might need some warning or other message output.

### LINKED ISSUE #1809: Failing SF code with multiple vs statements
```
1 vs 3

with

2 vs 4
```

creates the following plot which is wrong.

image

```
[1] vs [a]
with
[2] vs [b]
```

fails in

```igorpro
		if(action != SF_ACTION_SKIP && lastAction == SF_ACTION_ARRAY)
			// If the last action was the handling of "]" from an array
			SFH_ASSERT(action == SF_ACTION_ARRAYELEMENT || action == SF_ACTION_HIGHERORDER, "Expected \",\" after \"]\"", jsonId=jsonId)
		endif
```

as action is `SF_ACTION_COLLECT`.

t-b:
Close #1809
Close #1627


timjarsky:
@t-b Nice!

timjarsky (approved):

PR #1925, opened 2023-10-25
Add some code formats

Garados007:
## Format: Add brackets around operation flag arguments

Some operation flag arguments can contain function calls which can
probably conflict with normal operation values.

---

## Format: Add brackets around operation value division

Some operations do allow flags after their values (e.g. Make with
/WAVE). To prevent unintended conflicts are these values surrounded with
brackets when they contain a division.

---

## Format: Prevent variable name "as"

"as" is used inside Igor as a keyword and therefore its not a good
decission to use this as a variable name.


t-b:
@Garados007 LGTM. Please assign me once this is ready for review.

t-b (approved):

PR #1929, opened 2023-11-01
Add some formattings

Garados007:


t-b:
Thanks for the PR.

Commits look mostly good. But I don't see why cb49aa18660cf0f645aa78d126405aa6449043c4 is necessary. main passes without that as well. I guess it is because the formatting makes `10 ^ (` out of it and that is not recognized by the CI check.

I'll merge as is as we can live with that NOLINT.

t-b:
@Garados007 Please assign me if you think this is ready to merge.

PR #1937, opened 2023-11-09
Sutter preparation: Add headstage and clampmode information in config wave

MichaelHuth:
Preparation PR with changes for sutter hardware

- the config wave is upgraded with per hardware channel information of headstage and clampmode


t-b:
Review:

- [ ] Both comments in IsValidConfigWave mentioning the wave versions should be updated as we now have version 3 as well.
- [ ] The new code in GetDAQConfigWave has copied the comment from the v2 upgrade.
- [ ] Now that the config wave is double we don't need to create a FP64 wave anymore in HW_ITC_TransposeAndToDouble.
- [ ] Tests are missing. Some places to check headstage and clamp mode could be: TPDuringDAQ, TPDuringDAQTPAndUnAssoc and UnassociatedChannelsAndTTLs

t-b (approved):

PR #1944, opened 2023-11-30
IsValidSweepWave/IsValidSweepAndConfig: Allow null sweep waves again

t-b:
This was broken when adding NI support in 7712ba36 (IsValidSweepAndConfig Adapted for new data storage format as WREF wave, 2018-08-07).

### DIFF
--- a/Packages/MIES/MIES_MiesUtilities.ipf
+++ b/Packages/MIES/MIES_MiesUtilities.ipf
@@ -6336,17 +6336,21 @@ End
 threadsafe Function IsValidSweepWave(sweep)
 	WAVE/Z sweep
 
+	if(!WaveExists(sweep))
+		return 0
+	endif
+
 	if(IsWaveRefWave(sweep))
-		if(WaveExists(sweep) && DimSize(sweep, ROWS) > 0)
+		if(DimSize(sweep, ROWS) > 0)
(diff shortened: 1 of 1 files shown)


Packages/MIES/MIES_MiesUtilities.ipf
```diff
@@ -6363,13 +6367,15 @@ threadsafe Function IsValidSweepAndConfig(sweep, config, [configVersion])
 		configVersion = DAQ_CONFIG_WAVE_VERSION
 	endif
 
+	if(!WaveExists(sweep))
+		return 0
+	endif
+
 	if(IsWaveRefWave(sweep))
 		return IsValidConfigWave(config, version = configVersion) &&  \
-				 IsValidSweepWave(sweep) &&                           \
```
MichaelHuth, line 6368: The call would additionally check if there is at least one channel in the sweep that can be resolved and contains data.
Should this really be omitted here?

MichaelHuth (commented):
See comment

Packages/MIES/MIES_MiesUtilities.ipf
t-b: THanks, that was an oversight. Fixed.

MichaelHuth (approved):
ok

PR #1967, opened 2024-01-08
Make IgorInfo faster

### LINKED ISSUE #1966: Buffer call to IgorInfo(3) in a global and use that
Calling IgorInfo(3) needs about 25 ms, that is a lot compared to other code. (e.g. at each acquire, when documented to LNB, function 
GetIgorProVersion() )

As the returned value does not change it would sense to buffer the string in a global and use that instead.

t-b:
Close #1966

### DIFF
--- a/.github/workflows/test-igor-rebase-exec-workflow.yml
+++ b/.github/workflows/test-igor-rebase-exec-workflow.yml
@@ -74,18 +74,14 @@ jobs:
         uses: actions/checkout@v3
         with:
           submodules: recursive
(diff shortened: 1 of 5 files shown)


t-b:
Reassigning to me to fix the each commit CI job.

MichaelHuth (approved):
LGTM

the ITC1600 test case fails seem unrelated?

t-b:
> the ITC1600 test case fails seem unrelated?

Yes.

PR #1972, opened 2024-01-15
Sutter preparation: Adaptation of Acquire_NG for sutter hardware

MichaelHuth:
The logic in Acquire_NG was changed to work with Sutter hardware.


t-b (approved):
LGTM.

PR #1973, opened 2024-01-15
Sutter preparation: Test adaptation for epochs tests to work with Sutter

MichaelHuth:
The epoch test had the issue that they were targetted for higher DA frequencies like 100 kHz, but sutter runs with 10 kHz.
Thus, some checks were off by one sample, e.g. the amplitude check.

Adapted determination of amplitude boundary checks and TP boundaries.

### DIFF
--- a/Packages/MIES/MIES_Epochs.ipf
+++ b/Packages/MIES/MIES_Epochs.ipf
@@ -74,11 +74,12 @@ End
 
 static Function EP_CollectEpochInfoDA(string device, STRUCT DataConfigurationResult &s)
 
-	variable i, channel, singleSetLength, epochOffset, epochBegin, epochEnd
+	variable i, channel, singleSetLength, epochOffset, epochBegin, epochEnd, lastP
 	variable stimsetCol, startOffset, stopCollectionPoint, isUnAssociated, testPulseLength
 	string tags
 
 	stopCollectionPoint = ROVar(GetStopCollectionPoint(device))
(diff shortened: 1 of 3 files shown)


t-b:
Review:

5853905f (TP: Factor out determination of points that create testpulse, 2024-01-12)

Good.

6fe487b2 (EP: Fix that end of epochs are limited to end at last data point, 2024-01-12)

I agree with the changes. But we eiter need to add a new test or adapt an existing one, so that the fix in this commit is also reflected in the test suite.

b9dbbdb6 (EP: When adding TP epochs do not redo parts of the calculation, 2024-01-12)

- Missing test or test adaptation
- The commit message does also not mention why we need this change

2b54e709 (Tests: Adaptation to Epoch tests for Sutter, 2024-01-12)

From our conversation I know that this fix is required for lower sampling frequencies. But the commit message does not say it. And also no adapted test.

If possible please add the test or adaptation to each commit so that we can understand that also in a couple of years.

MichaelHuth:
I added more description to the commit messages, I added one test to [d4f1df3](https://github.com/AllenInstitute/MIES/pull/1973/commits/d4f1df33800db0afd8027d74a55616ec3c447179)

The next commits is regarding to the test very close related. So I referred to the test in the commit message.
The last commit is a pure test adaptation. The referred test also verifys that this adaptation works.

t-b:
LGMT. Very nicely explained and a nice test.

t-b (approved):

PR #1982, opened 2024-01-17
Check unassociated DA/AD channels better

### LINKED ISSUE #1977: Add pre-Acquisition check that AD or DA gain is not zero for unassoc channels
Currently we can run a measurement with zero gain, that results in various assertions because the division by the gain factor results in Infs. This is not supported by various after DAQ analysis.

t-b:
Close #1977


MichaelHuth:
LGTM.

timjarsky (approved):

Issue #1976, opened 2024-01-16
Tests: Use keepDataFolder flag in RunWithOpts to skip AdditionalExperimentCleanup in TEST_CASE_END_OVERRIDE


MichaelHuth:
Sometimes a developer would like to diagnose panel states that were setup in a test case. Then it is unlucky if they are closed at the end of the test case.

We can store the keepDataFolder flag of RunWithOpts in a global to prevent the panel to be cleaned up at the end.
This induces possible errors reported from the WaveLeak checks but is good for diagnostics.

PR #1984, opened 2024-01-18
Test config cleanup

t-b:


MichaelHuth:
LGTM but I don't see how that closes #1976

t-b:
> LGTM but I don't see how that closes #1976

2608fc6a5a195bb7c60f877da6eb1c5a4b0020a0 only closes the panels and removes the MIES folder if wave leak tracking is enabled. As wave leak tracking is disabled by default I thought this would fix our normal usecase when debugging.

MichaelHuth:
Ok.

semi related: Do you think we should disable wave tracking by default when keepDataFolder is enabled?

t-b:
> semi related: Do you think we should disable wave tracking by default when keepDataFolder is enabled?

I don't think this is necessary. `keepDataFolder` is for preserving the temporary DF which igortest creates and makes the CDF. I rarely use that.

PR #1985, opened 2024-01-18
Tests/CheckLogFiles: Fix the test

### LINKED ISSUE #1970: Test: CheckLogFiles in UTF_Utils.ipf fails if there is logfile existing at test start.
If at the begin of this test a MIES.jsonl file is existing the test fails.

t-b:
We should only ignore the lines with {}. And only if we only have these lines can we skip the file.

Added also two more info calls for better debugging.

Bug introduced in ed8a41dc (LOG_AddEntry: Readd timestamp, 2023-12-11).

Close #1970

### DIFF
--- a/Packages/tests/Basic/UTF_Utils.ipf
+++ b/Packages/tests/Basic/UTF_Utils.ipf
@@ -7391,11 +7391,12 @@ End
 
 static Function CheckLogFiles()
 
-	// ensure that the ZeroMQ logfile exists as well
-	PrepareForPublishTest()
-
-	string file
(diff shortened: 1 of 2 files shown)


t-b:
Timing issue:

The log line is

```
"{"direction":"Incoming","str":":heartbeat","ts":"2024-01-19T08:54:55.1973026-0800"}"
```

and that is coming form the heartbeat thread. I'll do some thinking how to move forward.

MichaelHuth (approved):
Looks good to me.

PR #1986, opened 2024-01-18
Tests: Call zeromq_stop after the complete test run

### LINKED ISSUE #1975: zeromq prevents multiple instances of Igor running MIES
The main application is here to compare tests with e.g. different hardware.

I am not sure if it would be ok to add `zeromq_stop()` to `TEST_END_OVERRIDE` as simple solution?

This should work as long as tests are not run simultaneously in the Igor instances.

t-b:
This allows to run tests for local debugging purposes in different igor instances.

Close #1975


MichaelHuth:
LGTM. We should only define at some point, when to use static and not for the hooks.

PR #1989, opened 2024-01-23
Add read/write support for wave reference waves in JSON wave notes

t-b:


t-b:
Results from review from @MichaelHuth:
- [ ] Add two variants for reading one for contained numeric and one for contained text
- [ ] Extend tests for text waves and read/write of empty/null waves
- [ ] assert on invalid waves types and contained null waves

t-b:
Ready for review.

MichaelHuth:
Looks good to me.

The extra Make to do the null wave check seems to be not very efficient, though there doesn't seem to be an obvious better method. I thought of FindValue but that does not accept WREF waves.

MichaelHuth (dismissed):

t-b:
> Looks good to me.
> 
> The extra Make to do the null wave check seems to be not very efficient, though there doesn't seem to be an obvious better method. I thought of FindValue but that does not accept WREF waves.

I've removed the make invocation and used our `GetRowIndex` wrapper.

```
$ git range-diff @{u}...HEAD
1:  e28ad6af0 ! 1:  695662449 MIES_JSONWaveNotes.ipf: Add read/write support for wave reference waves
    @@ Packages/MIES/MIES_JSONWaveNotes.ipf: threadsafe Function/WAVE JWN_GetTextWaveFr
     +
     +   JSON_Release(jsonID)
     +
    -+   Make/FREE/N=(numRows) junk = WaveExists(container[p])
    -+
    -+   ASSERT_TS(Sum(junk) == numRows, "Encountered invalid waves in the container")
    ++   ASSERT_TS(IsNaN(GetRowIndex(container, refWave = $"")), "Encountered invalid waves in the container")
     +
     +   return container
     +End

th
```

Issue #1147, opened 2021-11-10
Epoch info is off-by-one for level 3 pulse and baseline start/end times


t-b:
Investigate if this is something we can fix easily or if it depends on real vs minimum sampling interval etc. As this information is gathered in the wavebuilder it might also be a problem there.

Example file on FTP: `2021-10-22 Spike Control Failures/2021_10_20_111205-compressed_resave.nwb`

timjarsky:
Investigate to determine the cause and confirm off-by-one is restricted to level 3

MichaelHuth:
- Check in WB how edges are created in discrete stimset data from edge time
- How does decimation shifts (trunc or round)
- How is epoch time calculcated compared to WB edge time
- Can we reliably calculate the sample before / after the edge in DAQDataWave?

Possibilities:
- Change epoch time creation to be at correct sample points

- Allow to calculate epochs for old non-epoch data in the future
- Revise creation of stimset such that WB stimset has the same sample rate as the target DAQDataWave, remove decimation

timjarsky:
"

  *   Revise creation of stimset such that WB stimset has the same sample rate as the target DAQDataWave, remove decimation

— it's not possible to know the sample rate at the time a stimset is made in the WB. E.g., users can turn off headstage during a RA cycle. One way this could work is if the WB made the stimset at all supported sample rates however this feels like a BIG change in how MIES works.

On Jan 23, 2024 9:24 AM, Michael Huth ***@***.***> wrote:

  *   Check in WB how edges are created in discrete stimset data from edge time
  *   How does decimation shifts (trunc or round)
  *   How is epoch time calculcated compared to WB edge time
  *   Can we reliably calculate the sample before / after the edge in DAQDataWave?

Possibilities:

  *   Change epoch time creation to be at correct sample points

  *   Allow to calculate epochs for old non-epoch data in the future

  *   Revise creation of stimset such that WB stimset has the same sample rate as the target DAQDataWave, remove decimation

t-b:
>> Revise creation of stimset such that WB stimset has the same sample rate as the target DAQDataWave, remove decimation

> it's not possible to know the sample rate at the time a stimset is made in the WB. E.g., users can turn off headstage during a RA cycle. One way this could work is if the WB made the stimset at all supported sample rates however this feels like a BIG change in how MIES works.

[...]

PR #1990, opened 2024-01-25
Make epochs sample point exact

MichaelHuth:
- [ ] Check / Fix current WB segment / signal creation
- [ ] WB: Add indice information to stimset wave note for all epoch-relevant segments  / signal structures
- [ ] Create a routine that converts such stimset index to an DAQDataWave index (after resample)
- [ ] Gather all epoch-relevant signal structures from DC as indices (DAQDataWave based)
...

close https://github.com/AllenInstitute/MIES/issues/1147


MichaelHuth:
@t-b I thought about the effective stimset length increase for TTL on ITC and if it can happen on DA too.
I came to the conclusion that it can not happen because:
On TTL what happens is that if you have 2+ TTL channels where at least one has a shorter stimset than the others, all TTL channels are put in a 2D wave that has the size of `max(TTLChannel_Stimsets)`. This happens in DC before the TTL channels are decimated into the data wave. Thus, considering the single TTL stimset in DC the length gets increased and then decimated with the increased length.

For DA on the other hand, the stimset from WB is never increased in size in DC and never decimated to the data wave with the increased size. So `singleSetLength` stays valid here for the decimated stimset size.
(The delta mechanism for another sweep extends the stimset within the WB and not "externally" later in e.g. DC)

t-b:
@MichaelHuth Thanks for the massive effort. I'll have a look tomorrow.

t-b (changes requested):
See comment.

MichaelHuth:
@t-b 
> The attached experiment extracts some epochs from freshly measured data, and the border are not 100% correct.

The DA data in the experiment seems to be correct. There are three issues though:
- SFH_GetSweepsForFormula was not adapted to take our new convention for epoch times into account and retrieved one point too much. I fixed that with another commit.
- You looked with the data operation on AD channels, but Epochs only refer to DA or TTL.
- In case that was measured in a loopback configuration on ITC then AD channels show indeed an offset compared to DA. This is a serious bug, but not related to epochs but to acquisition and needs to be investigated.

MichaelHuth:
> The change here throws away OD1 from the setup. Is that intended?

The oodDAQ epochs are from the output of the oodDAQ algorithm. Thus, they are not sample exact and I excluded them from the test.

t-b:
LGTM, thanks for the fixes.

t-b (approved):

MichaelHuth:
@t-b I fixed the epoch EndTime correction after acquisition.
- If the acquisition was not aborted early, no correction should happen
- If a correction happens the time should be sample exact regarding DA/TTL

This fix revealed a bug in the epoch creation where the size of the first epoch in flipped stimsets was too short by one data wave point. The effect was that the end of the Stimset and the end of that epoch did not match. This is also fixed.

As well as a small fix for a uncaught RTE in EP_SortEpochs.

t-b:
Old revision: 3ad5e788 (Doc: Update documentation for epochs, 2024-02-19)
New revision: 4145e6a3 (Doc: Update documentation for epochs, 2024-02-19)

New commits:
```
 -:  --------- >  7:  559e33f0c Util: Add utility function FindFirstNaNIndex
 -:  --------- >  8:  2214eeaf6 EP: global EndTime reference for epochs is now time of DA wave size + 1
 -:  --------- >  9:  81818f617 EP: Fix uncaught RTE in EP_SortEpochs
```

The new commits are good. The RTE in EP_SortEpochs is also a nice find. Additional changes are also very nice.

t-b (approved):

MichaelHuth:
rebased

timjarsky:
@MichaelHuth @t-b, what is the definition of "sample point exact"? Is it a functional definition, or should the difference between the epoch time and the sample point = 0?

I ask because I do see a tiny difference between the epoch time and sample point; 
image

timjarsky:
[pulse_train_sample_point_exact.zip](https://github.com/AllenInstitute/MIES/files/14704528/pulse_train_sample_point_exact.zip)

MichaelHuth:
I have to look into that, where that ns deviation could originate from.

MichaelHuth:
@timjarsky The issue was that the waves for the traces for displaying the epochs were in single precision. I pushed a commit that changes the trace waves to be double precision and this seems to fix the difference here. The small discrepancy should be gone now.

timjarsky (approved):

t-b (approved):

PR #1998, opened 2024-02-06
MIES_TestPulse_Single.ipf: Zero DAC after stopping acquisition

### LINKED ISSUE #1965: Add test to verify that zeroDAC works
Currently we have no test that zeroDAC actually zeroes the output. As the CI machine uses a loopback configuration this would be possible by reading the output through the looped ADC channel.

Same logic applies for a potential test of HW_WriteDAC.

### LINKED ISSUE #1996: TPS_TestPulseFunc is missing zeroDAC=1 and abortOnError
Check also other places.

t-b:
We always pass zeroDAC = 1 for TP MD so there is no reason to deviate for
single device from that approach.

We also pass the abort on error flag as done in a lot of other call sites
of HW_StopAcq.

Close #1996
Close #1965

### DIFF
--- a/Packages/tests/HardwareBasic/UTF_TestPulseAndTPDuringDAQ.ipf
+++ b/Packages/tests/HardwareBasic/UTF_TestPulseAndTPDuringDAQ.ipf
@@ -1271,3 +1271,40 @@ static Function TPDuringDAQwithPS_REENTRY([str])
 
 	CHECK_EQUAL_WAVES(stimScale, {tpAmp, daGain, NaN, NaN, NaN, NaN, NaN, NaN, NaN}, mode = WAVE_DATA)
 End
+
+// UTF_TD_GENERATOR v0:SingleMultiDeviceDAQ
+// UTF_TD_GENERATOR s0:DeviceNameGenerator
+static Function TPZerosDAC([STRUCT IUTF_MDATA &md])
+	STRUCT DAQSettings s
+	InitDAQSettingsFromString(s, "MD" + num2str(md.v0) + "_RA1_I0_L0_BKG1_TP1" + \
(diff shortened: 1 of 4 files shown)


Packages/tests/HardwareBasic/UTF_TestPulseAndTPDuringDAQ.ipf
```diff
@@ -1271,3 +1271,41 @@ static Function TPDuringDAQwithPS_REENTRY([str])
...
+	CHECK_WAVE(indizes, NUMERIC_WAVE)
+	CHECK_EQUAL_VAR(DimSize(indizes, ROWS), 1)
+
+	DAC = DACs[indizes[0]]
+	CHECK_GE_VAR(DAC, 0)
+
+	deviceID = ROVar(GetDAQDeviceID(str))
+	hardwareType = GetHardwareType(str)
+
+	CHECK_LE_VAR(HW_ReadADC(hardwareType, deviceID, DAC), 0.01)
```
MichaelHuth, line 1310: I think that is not correct, because you have to get the ADC channel that is associated to the DAC channel and call ReadADC with the ADC channel.

MichaelHuth (changes requested):
It would also be good to mention in the commit/test description message that the test requires a loopback configuration.

Packages/tests/HardwareBasic/UTF_TestPulseAndTPDuringDAQ.ipf
t-b: Will fix. Is also wrong for the other new test.

t-b: Done. I've also made the code shorter and are now testing for SD and MD.

MichaelHuth (approved):

PR #2000, opened 2024-02-06
Tests/AdditionalExperimentCleanup: Preserve the DebugPanel waves

### LINKED ISSUE #1974: DebugPanel MouseWheel does not scroll list
- MouseWheel does not scroll list.
- Also list position is reset to top when a test is run

t-b:
Although 82a4bc59 (AdditionalExperimentCleanup: Handle reopening the debug panel correctly, 2021-12-22) claimed to fix the debug panel closing and reopening properly, it failed to preserve the listbox wave including the selections.

We now do that.

Close #1974


MichaelHuth (approved):
ok

IIf we do something like that more often, we should consider a generic logic to preserve the window position on closing and set it on reopening as well.

PR #2015, opened 2024-02-21
Recreate DataConfigurationresult Structure from Labnotebook

MichaelHuth:
- Recreate DataConfigurationresult structure from Labnotebook
- Use the recreated structure to recreate epochs sample exact

### DIFF
--- a/.github/workflows/build-pr.yml
+++ b/.github/workflows/build-pr.yml
@@ -121,7 +121,7 @@ jobs:
       experiment: Packages/tests/Compilation/CompilationTester.pxp
       installer_flags: "-s git"
       artifact_name: Compilation-Each-Commit-assets
-      timeout_minutes: 60
+      timeout_minutes: 120
 
   Documentation:
     name: 👷 Documentation
--- a/Packages/MIES/MIES_DataConfigurator.ipf
+++ b/Packages/MIES/MIES_DataConfigurator.ipf
@@ -284,12 +284,21 @@ End
 /// @param device  panel title
 /// @param dataAcqOrTP acquisition mode, one of #DATA_ACQUISITION_MODE or #TEST_PULSE_MODE
 static Function DC_CalculateDAQDataWaveLength(string device, variable dataAcqOrTP)
+
 	variable hardwareType = GetHardwareType(device)
 	NVAR stopCollectionPoint = $GetStopCollectionPoint(device)
 
+	return DC_CalculateDAQDataWaveLengthImpl(stopCollectionPoint, hardwareType, dataAcqOrTP)
+end
+
+/// @brief device independent implkementation for @ref DC_CalculateDAQDataWaveLength
+Function DC_CalculateDAQDataWaveLengthImpl(variable dataLength, variable hardwareType, variable dataAcqOrTP)
+
+	variable exponent
+
 	switch(hardwareType)
 		case HARDWARE_ITC_DAC:
-			variable exponent = FindNextPower(stopCollectionPoint, 2)
+			exponent = FindNextPower(dataLength, 2)
 
 			if(dataAcqOrTP == DATA_ACQUISITION_MODE)
 				exponent += 1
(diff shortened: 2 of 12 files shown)


MichaelHuth:
rebased

t-b:
Review: 

- Can we use only main/release-XXX as base branches? It is heavily confusing to merge A PR into another PR.

(Command: git nicelog --reverse  feature/1990-make_epochs_samplepoint_exact.. main..)

16778bf7 (DC: Split device independent part of decimationFactor calc to own function, 2024-02-21)

Makes sense.

c3f08e99 (Getters: Add function to get TP wave as free wave, 2024-02-28)

Jeep.

7ce8916a (Getters: Add function to get TPSettingsCalculated as free wave, 2024-02-28)

GetTPSettingsCalculated now has dead code at the end. I think the `return` in the else branch is not right.

f8678991 (Getters: Add getter for a free sampling intervals wave, 2024-02-28)

Jeep.

572f85f3 (TP: Split core calculation of TP properties in device dep and indep functions, 2024-02-28)

Yes.

8de72302 (DC: Move data wave size and insertStart calculation in device independent functions, 2024-02-28)
8202e669 (EP: Add more debug output to EP_AddEpoch, 2024-02-28)
cd00eeb6 (EP: Add a device independent function for EP_AdaptEpochInfoChannel, EP_SortEpochs, 2024-02-29)

Makes sense.

cd2814e5 (EP: Fix EP_EpochStrToWave, did not set DimLabels, 2024-02-29)
33259176 (OOD: Moved stimset offset modification to own device independent function, 2024-02-29)
d68b0505 (SWS: Moved determination of acquired time and NaN filling of AD channels to own functions, 2024-02-29)
99191da3 (DC: Move data length calculation to device independent function, 2024-02-29)

All good.

deff3d40 (DC: Add function to recreate DataConfigurationResult structure from LNB, 2024-02-21)

- [ ] I don't think the check `if(DimSize(s.DACList, ROWS))` makes sense here. We always have one DAC so this should be an assertion or?
- [ ] DC_RecreateDataConfigurationResultFromLNB_ADC/DC_RecreateDataConfigurationResultFromLNB_DAC make no ADC/DAC channels an assertion

- [ ] This is not needed:
```
-static Function EP_CollectEpochInfoDA(WAVE/T epochWave, STRUCT DataConfigurationResult &s)
+Function EP_CollectEpochInfoDA(WAVE/T epochWave, STRUCT DataConfigurationResult &s)
```

6767cbde (EP: Add utility functions for epoch recreation from LNB, 2024-03-01)

- [ ] EP_FetchEpochsFromRecreated should use EP_EpochWaveToStr instead of duplicating the code

eeada3f3 (Tests: Add ignoreErr argument to ListToNumericWave, 2024-03-01)

- [ ] Missing tests

f2eee74a (Tests: Add test to TestEpochGeneric that verifies that epoch recreation roundtrips, 2024-02-29)

- [ ] Commit message: `eery` -> `very`
- [ ] I'm not understanding the comment `// also TP channels can be active but have no epochs`. We don't test TTL channels here.

baa92348 (CI: Increase timeout for testing each commit for compilation to 2h, 2024-03-05)

Makes sense.

MichaelHuth:
> GetTPSettingsCalculated now has dead code at the end. I think the return in the else branch is not right.

That return is intentional because the SetTPSettingsCalculatedProperties should not be called twice. I have a single point of creation in GetTPSettingsCalculatedAsFree, where the properties are set. In GetTPSettingsCalculated there are then three paths:

1. The wave exists with the correct version
2. The wave exists but needs to be upgraded
3. The wave does not exist and should be created as permanent

For 2 I have to set the latest properties at the end, for 3 this is already done in the getter for the free version of the wave, so I can return early.

I have slightly rearranged the code, but I think there is no dead code.

t-b:
> I have slightly rearranged the code, but I think there is no dead code.

You are right. I've misread the code.

LGTM now.

t-b (approved):

PR #2037, opened 2024-03-07
bugfix: FIFO2WAVE does not work with single point ranges

t-b:
Backport of #2034.

Will merge once CI passes.


t-b:
@MichaelHuth The newly added test TestNIAcquisitionReliability fails here. This is the latest release branch. Can you have a look?

MichaelHuth:
Does not look like a FIFO2WAVE fail.
It seems the IUTF_BackgroundMonitor ran into a timeout, this test seems to run longer on the CI.

MichaelHuth:
Locally the TestNIAcquisitionReliability test runs 5 min.

PR #2058, opened 2024-03-13
EP: fix epoch sorting to be numerically correct

### LINKED ISSUE #2046: EP_SortEpochs sorts alphabetically, but must sort numerically.
Epoch sorting is broken.

### LINKED ISSUE #2059: EP_SortEpochs deletes last epoch of channel if channel is filling all rows of epoch wave
If a single channel of the epochs has exactly as many epochs as the size of the 4d epochs wave then EP_SortEpochs deletes the last epoch of that channel and creates a uncaught RTe due to a missing range check.

MichaelHuth:
The epoch sorting had two issues:

1. The sorting was alphabetically such that 2.xxx was between 19.xxx and 20.xxx which is incorrect.

2. The function assumed that there is always some remaining rows to cutoff at the end. If that was not the case then the epoch in the last row is deleted.

1. since a2172f0
2. since 4ed816a

close #2046
close #2059

### DIFF
--- a/Packages/MIES/MIES_Constants.ipf
+++ b/Packages/MIES/MIES_Constants.ipf
@@ -27,7 +27,7 @@ Constant ANALYSISBROWSER_PANEL_VERSION    =  3
 Constant STIMSET_NOTE_VERSION = 9
 
 /// Version of the epoch information for DA+TTL data
-Constant SWEEP_EPOCH_VERSION = 8
+Constant SWEEP_EPOCH_VERSION = 9
 
 /// Version of the labnotebooks and results (numerical and textual) waves
 ///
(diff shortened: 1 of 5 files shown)


MichaelHuth:
@t-b The CI machine has somehow run low on disk space, that leads to CI failing on acquisition at a random point.

t-b:
Review:
- Same comment about `since $commithash` as in the other MR
- NOISE_GEN_XOSHIRO should be preferred for `3` in call to enoise.
- We do have a constants for HASH_SHA2_256 for WaveHash. If you want to use something else please also create a constant.
- Needs to raise SWEEP_EPOCH_VERSION
- You are not using a fixed seed for enoise. Which I think makes sense for these kind of tests. But maybe you want to call `INFO("mixRandom = %n", n = mixRandom)` before the CHECK_EQUAL_STR assertion so when it fails we get the resorted order and this allows us to reproduce the bug.

t-b:
I don' see the points I raised adressed.

MichaelHuth:
sorry I somehow missed the review.

t-b (approved):
Np. LGTM.

PR #2063, opened 2024-03-13
WB: fix recreation of dependent Stimsets from Stimset parameter waves

MichaelHuth:
The WB has an issue, where non-existing stimsets are not treated as needing an update. Thus, combine type stimsets that depend on another stimset being created first can not be created because the other stimset is not being created from the parameter waves.

since 792d0d2

close https://github.com/AllenInstitute/MIES/issues/2060

- [x] add test

PR #2064, opened 2024-03-14
Add tests for nwb stimset roundtrip

### LINKED ISSUE #2054: Menu "Export all stimsets into NWB" does not export dependent formula stimsets
If a stimset is created with the combine epoch that depends on a source stimset for the formula, then the source stimset is not saved.
This breaks the stimset with the combine epoch because it can not be recreated without the missing stimset.

This problem applies also to "Export all data into NWB"

### LINKED ISSUE #2057: Add roundtrip tests for combine/custom stimsets
- PXP/NWB
- Store sweep with used stimsets only
- Check after loading the used stimsets that the dependencies (referenced stimsets and custom waves are included)

### LINKED ISSUE #2067: Sometimes AB converts stimsets with custom waves to third party stimsets on load from PXP or fails to load these stimsets entirely
Error:
When stimsets with custom waves were loaded through the AB then only the already saved stimset wave and not the parameter waves + custom wave was loaded.

- This converted this stimset to third party
- If originally only the parameter waves and custom wave existed in the PXP then it failed entirely

Diagnosis:
  AB_LoadWave expected exactly one wave to be loaded after a call to
  AB_LoadDataWrapper that loads waves recursivley in all sub data folders
  If in a sub DF wave(s) with the same name template existed then
  these were loaded too.
  While the loaded waves contained the intended wave, the result was
  rejected.

### LINKED ISSUE #2068: ST_GetStimsetList incorrectly matching against saved stimsets
ST_GetStimsetList uses incorrectly a case sensitive matching in GetListDifference to determine third party stimsets.
The effect is that ST_GetStimsetList can return duplicates like

`formula_DA_0;formula_da_0;`

MichaelHuth:
- [x] add tests for stimset roundtrip through stimset save/load NWB
- [x] add tests for stimset roundtrip through stimset save/load PXP
- [x] add tests for stimset roundtrip through sweep save/load NWB
- [x] add tests for stimset roundtrip through sweep save/load PXP

requires #2063

close #2057
close #2067
close #2068
close #2054

### DIFF
--- a/Packages/tests/Basic/UTF_AnalysisBrowserTest.ipf
+++ b/Packages/tests/Basic/UTF_AnalysisBrowserTest.ipf
@@ -164,3 +164,199 @@ static Function LoadDependentStimsetsFromPXP()
 
 	KillWindow $abWin
 End
+
+// UTF_TD_GENERATOR DataGenerators#RoundTripStimsetFileType
+static Function RoundTripDepStimsets([string str])
+
+	string fName, abWin, sweepBrowsers, stimsets, refList, setNameF, setNameB, formula, set
+	string setNameFormula = "formula"
+	string setNameBase = "baseSet"
+	string baseFileName = "RoundTripDepStimsets." + str
+	string chanTypeSuffix = "_DA_0"
+	variable nwbVersion = GetNWBVersion()
+
+	DFREF dfr = GetWaveBuilderPath()
+	KillDataFolder dfr
+
+	setNameF = ST_CreateStimset(setNameFormula, CHANNEL_TYPE_DAC)
+	ST_SetStimsetParameter(setNameF, "Total number of epochs", var = 1)
+	ST_SetStimsetParameter(setNameF, "Total number of steps", var = 1)
+	ST_SetStimsetParameter(setNameF, "Type of Epoch 0", var = EPOCH_TYPE_SQUARE_PULSE)
+	ST_SetStimsetParameter(setNameF, "Duration", epochIndex = 0, var = 10)
+	ST_SetStimsetParameter(setNameF, "Amplitude", epochIndex = 0, var = 1)
+
+	setNameB = ST_CreateStimset(setNameBase, CHANNEL_TYPE_DAC)
+	ST_SetStimsetParameter(setNameB, "Total number of epochs", var = 1)
+	ST_SetStimsetParameter(setNameB, "Total number of steps", var = 1)
+	ST_SetStimsetParameter(setNameB, "Type of Epoch 0", var = EPOCH_TYPE_COMBINE)
+	formula = "2*" + LowerStr(setNameF) + "?"
+	ST_SetStimsetParameter(setNameB, "Combine epoch formula", epochIndex = 0, str = formula)
+	ST_SetStimsetParameter(setNameB, "Combine epoch formula version", epochIndex = 0, str = WAVEBUILDER_COMBINE_FORMULA_VER)
+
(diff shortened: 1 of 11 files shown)


Packages/tests/Basic/UTF_AnalysisBrowserTest.ipf
```diff
@@ -164,3 +164,82 @@ static Function LoadDependentStimsetsFromPXP()
...
+	wbPanel = WBP_CreateWaveBuilderPanel()
+	CHECK_PROPER_STR(wbPanel)
+	PGC_SetAndActivateControl(wbPanel, "SetVar_WaveBuilder_P0", val=10)
+	PGC_SetAndActivateControl(wbPanel, "SetVar_WaveBuilder_P2", val=1)
+	PGC_SetAndActivateControl(wbPanel, "setvar_WaveBuilder_baseName", str=setNameFormula)
+	PGC_SetAndActivateControl(wbPanel, "button_WaveBuilder_SaveSet")
+	PGC_SetAndActivateControl(wbPanel, "setvar_combine_formula_T6", str="A*2", switchTab=1)
+	PGC_SetAndActivateControl(wbPanel, "setvar_WaveBuilder_baseName", str=setNameBase)
+	PGC_SetAndActivateControl(wbPanel, "button_WaveBuilder_SaveSet")
+	KillWindow/Z wbPanel
```
t-b, line 187: Please use the stimset creation API instead of the wavebuilder, examples are at https://github.com/AllenInstitute/MIES/blob/main/Packages/doc/ipf/example-stimulus-set-api.ipf.

t-b:
Review:

d229744c (WB: fix uncaught RTE on initial WB creation, 2024-03-14)

Good find.

a4595827 (AB: Add optional recursive argument to AB_LoadDataWrapper, 2024-03-15)

Nice fix.

735bee09 (AB: fix AB_LoadWave rejecting data if more than one wave is loaded, 2024-03-15)

Yes.

c30d79e1 (STAPI: fix ST_GetStimsetList incorrectly matching against saved stimsets, 2024-03-15)

Very nice history digging!

- [ ] Can we have the changes for GetListDifference in its own commit with tests?
- [ ] Add a test to UTF_StimsetAPI.ipf so that this stays fixed

e7abe693 (NWB: fix Retrieve custom waves from all dependent stimsets, 2024-03-15)

- [ ] I haven't check but one of tests added later would fail without this fix right?

fb43f33e (Tests: Add dep Stimset roundtrip tests through sweep save/load, 2024-03-15)

- [ ] Can you create the global variables/strings in the CDF instead? As the tests will start in a temporary folder this allows you to skip the cleanup. CreateDependentStimset can also create the global waves in the CDF or?

```
+    return [setNameB, stimsetList, "root:" + wName, amplitude]
```

- [ ] Prefer `GetWavesDataFolder(wName, 2)` over `"root:" + wName`

4f2b17f0 (Tests: Add NWB roundtrip tests for combine and custom stimsets, 2024-03-14)

Looks good.

MichaelHuth:
> I haven't check but one of tests added later would fail without this fix right?

Yes the test in Tests: Add dep Stimset roundtrip tests through sweep save/load fails without this fix.

MichaelHuth:
@t-b I also changed the description of `GetListDifference` because I think it was incorrect.

t-b:
> @t-b I also changed the description of GetListDifference because I think it was incorrect.

Jeep. Thanks.

LGTM.

t-b (approved):

PR #2071, opened 2024-03-21
Increase timeout for acquisition in tests to 20 mins

### LINKED ISSUE #2069: TestNIAcquisitionReliability: Test failing with instrumentation in main
Failure: https://github.com/AllenInstitute/MIES/actions/runs/8302798808/job/22726821442

Introduced in 380c9b88 (bugfix: FIFO2WAVE does not work with single point ranges, 2024-03-06).

Can you reproduce the failure with instrumentation and expensive?

MichaelHuth:
The NI reliability test is running rather slow on the CI machine and exceeds 10 mins timeout for the acquisition. This results in a fail.

From the esitmation of 0.96 acquisition/s of the CI machine, a raise to 1200 s should be enough.

An alternative would be to upgrade the CI machine, as a factor of 2.5 times slower than my 4 year old development machine is questionable. (Also with respect to the time a full CI run takes)

close #2069


t-b:
On main we run the tests with instrumentation and that makes it slower.

t-b:
Let's try that.

t-b (approved):

Issue #2074, opened 2024-03-25
Recreate user epoch info from IVSCC analysis functions


t-b:
All user epochs from https://alleninstitute.github.io/MIES/file/_m_i_e_s___analysis_functions___patch_seq_8ipf.html should be recreated if possible. The PSQ_Ramp ones could be impossible.

t-b:
Start with `U_CR_CE`/`U_CR_SE`. The test cases `XXX2` are the ones which have everything passing, so `PS_CR2` is a good test case to start.

t-b:
Tags | Short Name | Analysis function | Description | Level
-- | -- | -- | -- | --
Name=Baseline Chunk;Index=x | U_BLCx | DA, RB, RA, CR, SE, VM, AR | Baseline QC evaluation chunks | -1
Name=Baseline Chunk QC Selection;Index=x | U_BLSx | SE, VM, AR | Selects the chunk for Baseline QC evaluation | -1
Name=DA Suppression | U_RA_DA | RA | DA was suppressed in this time interval | -1
Name=Unacquired DA data | U_RA_UD | RA | Interval of unacquired data | -1
Type=Testpulse Like;Index=x | U_TPx | PB, SE, AR | Testpulse like region in stimset | -1
Type=Testpulse Like;SubType=Baseline;Index=x | U_TPx_B0 | PB, SE, AR | Pre pulse baseline of testpulse | -1
Type=Testpulse Like;SubType=Pulse;Amplitude=y;Index=x | U_TPx_P | PB, SE, AR | Pulse of testpulse | -1
Type=Testpulse Like;SubType=Baseline;Index=x | U_TPx_B1 | PB, SE, AR | Post pulse baseline of testpulse | -1
Type=Chirp Cycle Evaluation | U_CR_CE | CR | Evaluation chunk for bounds state | -1
Type=Chirp Spike Evaluation | U_CR_SE | CR

PR #2083, opened 2024-04-11
Recreate user epochs for PSQ_Chirp

MichaelHuth:


t-b:
@MichaelHuth Can you check with the data we have on the FTP in folder `issue-881` that the chirp epochs are recreated correctly?

MichaelHuth:
There seems to be no data file with a stimset where PSQ_Chirp as used.

t-b:
How have you looked? I found some in C57BL6J-628261.02.01.02.nwb.

grafik

MichaelHuth:
I patched in a fallback for old data and reordered some LNB dependencies.
Though, I can not test it for correctness ad the old data has no USER epochs to compare and used an old version of the analysisfunction code. With this the epoch recreation creates more epochs as there were originally present.
This raises the issue that the old file has regular epochs, such that by default no epoch recreation kicks in. However, if I enforce epoch recreation manually, I get more epochs out that originally saved.

t-b:
> Though, I can not test it for correctness ad the old data has no USER epochs to compare and used an old version of the analysisfunction code.

I think you need to visually check that the reconstructed user epochs are correct.

> This raises the issue that the old file has regular epochs, such that by default no epoch recreation kicks in. However, if I enforce epoch recreation manually, I get more epochs out that originally saved.

This should not be an issue as we want to regenerated all epochs including user epochs.

t-b:
Review:

- [ ] This PR does not close #2074 as it only implements parts of it

a501aa2d3 (EP: Make EP_AddUserEpoch device independent, 2024-04-11)

Jeep.

8ec12641d (Tests: Move helper function for epoch recreation testing to UTF_HelperFunctions, 2024-04-11)

Good.

7769990cb (Ana: refactor - move parts of Chirp user epoch creation to own device independent functions, 2024-04-11)

- [ ] Please add a comment at PSQ_CR_AddSpikeEvaluationEpoch/PSQ_CR_AddCycleEvaluationEpoch to say that the returned units are seconds.

bdcb65074 (Ana: Rename totalOnsetDelay in PSQ_CR_AddSpikeEvaluationEpoch to totalOnsetDelayMS, 2024-04-11)

Nice.

e5b3b19e6 (EP: Add EpochRecreation support for User epochs from PSQ_Chirp, 2024-04-11)

- [ ] I think the part

added epoch recreation test to PS_CR2, PS_CR12 and PS_CR13

in the commit message does not reflect the code anymore.

- [ ] Can we switch to using constants for the analysis function types? MapAnaFuncToConstant should help.
- [ ] No need to cite the function name in DEBUGPRINT, it is also wrong.

EP_AddRecreatedUserEpochs_PSQ_Chirp:

- [ ] Not being able to query the analysis function parameters should assert out

CommonAnalysisFunctionChecks:
- [ ] There should be always a "Generic Function" entry available
- [ ] The GetRowIndex call makes me shiver. You already have the analysis function type in `type` so no need to query the LBN again.

- [ ] TestEpochReceationRemoveUserEpochs: Missing `r`

adcbf417e (EP: Add recreation support for PSQ_Chirp baseline epochs, 2024-04-17)

- [ ] "number of evaluated chunk" -> "index of evaluated chunk"
- [ ] Don't passt the name of analysis functions around but their numeric type, in TestEpochReceationRemoveUnsupportedUserEpochs
- [ ] I know we talked about only removing BLC0 and BLC1, but now that i think
      about it again I think we should keep all "^BLC\d+$" epochs. Please also make
      that a strconstant close to PSQ_BASELINE_CHUNK_SHORT_NAME_PREFIX so that we
      see that both belong together. I'd also appreciate if we could move away
      from the loop handling here and use FindIndizes and friends.

3288f5dd0 (Tests: Add test for checking user epochs in historic data, 2024-04-20)

- [ ] Commit message: " does not has" > "does not have"
- [ ] prefer createDFWithAllParents over NewDataFolder
- [ ] Why the C&P of code in ExportEpochsFromFileToDFOpenAnalysisBrowserHelper
      from OpenAnalysisBrowser? I also don't see what
      ExportEpochsFromFileToDFOpenAnalysisBrowserHelper has to do with epochs.
      If you want to use OpenAnalysisBrowser outside of test code. Fine let's
      remove test assertions from OpenAnalysisBrowser.
- [ ] WAVE/Z refEpochs = $"" -> "Wave/ZZ refEpochs"
- [ ] Please add some documentation for CompareEpochsOfSweep and ExtendRefEpochsWithUserEpochs
- [ ] ExtendRefEpochsWithUserEpochs: I don't understand why you check

```
+        if(strsearch(shortName, EPOCH_SHORTNAME_USER_PREFIX, 0) != 0)
+            continue
+        endif
```

from the naming epochChannelRecUser only holds user epochs. Maybe this should be an assert?

General comment:
If you have test code which does

```
Function MyTEstFunction()

for(...)

  ...
  CHECK_XXX()
endfor

End
```

you **always** need an assertion before the loop that the number of loop
iterations is larger than zero. Otherwise you have passing test without
checking anything.

- [ ] The comment "For user epochs also the values are checked" for CompareEpochsHistoricChannel does not seem to match the code.

cf63db435 (Tests: Split generators in HistoricDataTests for PXP and NWB files, 2024-04-22)

Looks good!

t-b (changes requested):
See comment.

MichaelHuth:
> There should be always a "Generic Function" entry available

There is not always a "Generic Function" entry available, e.g. see the early MIES experiments from historic data.

MichaelHuth:
> I'd also appreciate if we could move away from the loop handling here and use FindIndizes and friends

I rewrote that part, though I think its more complex and less efficient now.

t-b:
Review:

- [ ] PSQ_BASELINE_CHUNK_SHORT_NAME_RE_MATCHER should be anchored to front and end, i.e. `^BLC[[:digit:]]+$` as we don't want to match "aBLC0b".

I've created https://github.com/AllenInstitute/MIES/issues/2095 for the obvious FindIndizes enhancement and https://github.com/AllenInstitute/MIES/issues/2096 for DeleteWavePoint.

Rest looks good.

t-b (changes requested):
See comment.

MichaelHuth:
Ok I removed the generic assembly of the regex in favor of including it directly in the strings.

t-b (approved):

PR #2087, opened 2024-04-17
UTF_AsynFrameworkTest.ipf: Correct out-of-order check

t-b:
Since the introduction of the tests in 14cd1ff8d6 (Add Async Framework, 2018-09-17) we tried to verify that the execution really happened out of order. The fix in b8ae7c65ce (Asyn: Fix WARN(1) in tests for orderless execution, 2022-11-03) did not fix it though as we actually need to check that the returnOrder array is not ascending.

As the workload in RunGenericWorker already does a randomized workload we feel lucky that it is never sorted and check that.

### DIFF
--- a/Packages/tests/Basic/UTF_AsynFrameworkTest.ipf
+++ b/Packages/tests/Basic/UTF_AsynFrameworkTest.ipf
@@ -732,15 +732,8 @@ static Function TASYNC_RunOrderless()
 	endfor
 	CHECK(!timeout)
 
-	if(!timeout)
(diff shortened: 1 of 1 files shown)


MichaelHuth:
LGTM, though we still need to see how reliable that "lucky" approach works. Otherwise we need a specialized worker to increase the chance of unordered thread finishing.

t-b:
@MichaelHuth So you approve?

MichaelHuth (approved):

Issue #1737, opened 2023-05-23
Logfile asserts: "Could not find an edge"


t-b:
Example:
```
{
  "action": "assert",
  "exp": "2023_03_28_155848",
  "id": "4539fc5",
  "message": "Could not find an edge",
  "source": "ASSERT",
  "stacktrace": "Stacktrace:
   DQM_FIFOMonitor(...)#L139 [MIES_DataAcquisition_Multi.ipf]
   AS_HandlePossibleTransition(...)#L105 [MIES_AcquisitionStateHandling.ipf]
   AFM_CallAnalysisFunctions(...)#L135 [MIES_AnalysisFunctionManagement.ipf]
   PSQ_Chirp(...)#L4501 [MIES_AnalysisFunctions_PatchSeq.ipf]
   PSQ_CR_GetSpikeEvaluationRange(...)#L4629 [MIES_AnalysisFunctions_PatchSeq.ipf]
   PSQ_GetPulseDurations(...)#L211 [MIES_AnalysisFunctions_PatchSeq.ipf]
   PSQ_DeterminePulseDuration(...)#L255 [MIES_AnalysisFunctions_PatchSeq.ipf]
   ",
  "ts": "2023-03-28T16:09:35.937-07:00"
}
```

Experiments:
```
"2022_05_26_200549"
"2023_02_07_140042"
"2023_03_06_131925"
"2023_03_09_120624"
"2023_03_28_155848"
"2023_04_04_124444"
"2023_04_10_130206"
"2023_04_11_125118"
"2023_04_13_141646"
"2023_04_14_114222"
"2023_04_26_090004"
"2023_04_26_104528"
"2023_04_28_092237"
"2023_04_28_124657"
```

Files:
```
.\W10DT1120232-08-audrey.mcc-2023-04-18T16_29_55Z\Log.jsonl.1SESuX
.\W10DT1120232-07-krista.blake-2023-04-14T18_26_11Z\Log.jsonl.5dqKMF
.\W10DT1120232-03-svc-ivscc-2023-04-11T16_10_14Z\Log.jsonl.iZkN43
.\W10DT1120232-03-svc-ivscc-2023-04-12T17_39_55Z\Log.jsonl.k2vUhh
.\W10DT1120232-03-svc-ivscc-2023-03-29T19_08_40Z\Log.jsonl.Pc08oq
.\W10DT1120232-03-svc-ivscc-2023-03-07T17_05_50Z\Log.jsonl.FCkXSQ
[...]

PR #2093, opened 2024-04-22
Recreate BLC, BLS user epochs

MichaelHuth:
requires #2083
requires upcoming LNB readout fix for non-consecutive sweep numbers


t-b:
If possible I'd like to review this first round this week.

t-b:
Issue: #2074 

Review:
fbce383fa (AnaPSQ: Factor out adding epochs for baseline chunk selection, 2024-04-22)

Make sense.

230302798 (EPR: Add BLC epochs for other PSQ functions and BLS epochs, 2024-04-22)

- [ ] I don't think anaFuncName and setName are used here.

- [ ] Can we define EP_AddRecreatedUserEpochs_GetAnaFuncParams in
      MIES_AnalysisFunctionHelpers.ipf or another place for general consumption? The
      generic function should also not throw.
      Just as a reminder we do have `WaveText` to spare the wave statement.

- [ ] The lines with chunkLengthTime, chunkStartTimeMax are C&P

```
+            numRequiredEpochs = PSQ_VM_REQUIRED_EPOCHS
+        endif
+        if((type == PSQ_ACC_RES_SMOKE) || (type == PSQ_TRUE_REST_VM))
+            if(!IsNaN(numEpochs) && numEpochs != PSQ_VM_REQUIRED_EPOCHS)
+                return NaN
+            endif
```

I thought we already discussed about this and you wanted to change it?

- [ ] All CreateAnaFuncLBNKey invocations should get a waMode
- [ ] Prefer the constant for "U_BLS[0-9]+"

b73e87a47 (AnaPSQ: Refactor SE UserEpoch creation for BLS, TP to dev indep function, 2024-04-23)

Good.

ac0f10dd5 (Ana: Refactor duration calculation to dev independent function, 2024-04-23)

Good.

2d6f7d7a2 (EP: fix EP_GetEpochs causing uncaught RTE when optional epochWave is null, 2024-04-24)

Nice find.

f0304da72 (EP: EP_SortEpochs sort as lowest priority to CRC of tag string, 2024-04-24)

Change is good. But it left the reader puzzled why you used a CRC. Maybe
because you can not mix textual and numeric keys for sorting?

2ba5e0870 (EP: Add BLC, BLS epochs and TP epochs for Seal Evaluation as recreated epochs, 2024-04-23)

- [ ] The waMode change is not mentioned and should happen in the commit introducing the code.
- [ ] `DEBUGPRINT("Could not reconstruct durations")` should mention why this
      can happen and refer to an issue as we think we need to fix that as well.
      I think this is https://github.com/AllenInstitute/MIES/issues/1737.

- [ ] This commit also includes other changes to Baseline which should happen separately and/or earlier.
- [ ] Don't use FindDuplicates but prefer GetUniqueEntries as the latter also handles one point waves properly.
- [ ] TestEpochRecreationRemoveUnsupportedUserEpochs: 
      - [ ] TP_.._B1 is missing a trailing $
      - [ ] tpEpochs should only be define in the if block before Concatenate

c95067395 (Tests: Adapt user epoch test for anaFunc, 2024-04-24)

Good!

070091cc3 (Ana: Add note in EP_EvaluateBaselineProperties for epochsWave trick, 2024-04-24)

Nice one.

f4924ab28 (Tests: Adapt HistoricData tests for new supported recreated epochs, 2024-04-30)

Good.

Can you edit https://github.com/AllenInstitute/MIES/issues/2074#issuecomment-2064545684 so that we see which user epochs are still missing?

PR #2108, opened 2024-05-13
DFREFClear: Add it

t-b:


t-b:
Will merge once CI passes.

MichaelHuth:
One thought regarding Async speed up is:

If dfr is global, one could try this:
- get the full path to dfr
- clear dfr
- MoveDataFolder $fullpath, :

If dfr is free I don't see a similar way as the existence depends on a existing reference or if it is the current DF. But we use the current DF as free DF (with no DFREF to it) already as the target. For that case only DuplicateDataFolder to : (or some other directly specified target path) keeps the target unreferenced.

If we would go away from the solution to use as target a free DF, instead we could use a global DF with a known name, where we never create a DFREF for. Then we could move an incoming free DF to a global DF and then try the first sketched approach (with two moves in total).

t-b:
@MichaelHuth Wrong thread. If you want to give that a try, feel free to reopen #1568.

PR #2023, opened 2024-02-23
SweepFormula/psxKernel: Various fixes

t-b:
@timjarsky This has our bugfixes for psxKernel and too large decayTau. I've removed the adaptations for the sweep range extraction as #1990 will fix that in a better way (also for old data).

- [x] Add test
- [x] Drop empty results in PSX_Operation
- [x] intersecting ranges bug IVSCCMiniAnalysis on the FTP (same name, new content)
- [x] StatsRangeTesting: Check that the correct range was used for multi dataset examples
- [x] Update psxStats documentation
- [x] Throw away events which have an amplitude which has a different sign compared to the psxKernel amplitude
- [x] Replace PSX_DEFAULT_X_START_OFFSET with a factor and the decay tau from psxKernel
- [x] Fix incorrect usage of StatsQuantiles as `/Z` with `err = ...` does not make sense
- [x] psx should return sweepData waves if there are no events, as psxPrep relies on that for the histogram. Otherwise the data psx uses and what psxPrep plots is different
- [x] Fix kernelAmp sign check again
- [x] Reverse order of offsetting and filtering, rename **all** names
- [x] Hack session with Tim to check that it is now what we want
- [x] Implement first version with outcome from hack session
- [x] psx rise time calculation seems to off now (see Sample NWB files from Jessica - playing around.pxp, sweep 48 event 0)
- [x] Fix bug, Tim uploaded data for assertion (Sample NWB files from Jessica.pxp)

```
  !!! Assertion FAILED !!!
  Message: "Can not work with NaN as mapIndex"
  Please provide the following information if you contact the MIES developers:
  ################################
  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  Stacktrace:
  SF_button_sweepFormula_display(...)#L2995 [MIES_SweepFormula.ipf]
SF_FormulaPlotter(...)#L1920 [MIES_SweepFormula.ipf]
SF_GetTraceColor(...)#L1518 [MIES_SweepFormula.ipf]
SFH_GetLabNoteBookForSweep(...)#L1842 [MIES_SweepFormula_Helpers.ipf]
  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  Time: 2024-11-26T14:05:34-08:00
  Locked device: [- none -]
  Current sweep: [- none -]
  DAQ: [- none -]
  Testpulse: [- none -]
  Acquisition state: [- none -]
  Experiment: Sample NWB files from Jessica (Packed)
[...]

Issue #1863, opened 2023-08-29
SweepFormula: Parser issues


t-b:
- [x] `1e-3` gives "Operand for - must be numeric". -> https://github.com/AllenInstitute/MIES/pull/2742

- [ ] `1 - NaN` gives "Operand for - must be numeric". Same for `inf`.

- [x] `[]` Empty array is not supported -> https://github.com/AllenInstitute/MIES/pull/2753

t-b:
Estimated time to fix: 2d

t-b:
New issues:
- `[1,]` is not valid but `[1, ]` is. We should always bail out

t-b:
@copilot Can you add support for `NaN`, `Inf`, `-inf` to the sweepformula parser and executor? The values should be treated as numeric values inside sweep formula. Create a new PR.

t-b:
All are fixed.

PR #2109, opened 2024-05-13
Revise sweepformula select operation

### LINKED ISSUE #2012: Extend the `select` operation to allow more fine grained sweep selection
The select operation [1] has the following signature:

```
select([array channels, array sweeps[, string mode[, string clampMode]]])
```

now we would like to add more control over sweep selection. Not only should it be able to select via displayed and clampmode but also via stimulus set and if it passed or not.

Sweep passing state is a LBN entry, see `PSQ_FMT_LBN_SWEEP_PASS` [2] as input for `CreateAnaFuncLBNKey`.

Not all analysis functions have that entry, only DA, SP, RA, CR, PB, SE, VM, AR so from the complete list

    PSQ_PipetteInBath PB
    PSQ_SealEvaluation SE
    PSQ_AccessResistanceSmoke AR
    PSQ_TrueRestingMembranePotential VM
    PSQ_DaScale (sub threshold) DA (sub)
    PSQ_SquarePulse (long pulse) SP
    PSQ_Rheobase (long pulse) RB
    PSQ_DaScale (supra threshold) DA (supra)
    PSQ_SquarePulse (short pulse) SP
    PSQ_Rheobase (short pulse) RB
    PSQ_Ramp RA
    PSQ_Chirp CR

and that means PSQ_Rheobase needs to be covered differently (see AD_FillWaves line 344, but note #2017).
Idea:

```
select([array channels, array sweeps, selectXXX(), ...])
```

where all `selectXXX` operations would be `and`-ed.

```
selectOptVis([all | displayed]) # default: displayed
selectOptCM([all | ic | vc | izero]) # default: all
selectOptStimset([name1, name2]) # name supports wildcards
selectOptIVSCCSweepQC(passed | failed)
selectOptIVSCCSetQC(passed | failed)
```

As users would also like to select multiple options OR'ed (e.g. passing sweeps from stimset A and failing sweeps from stimset B) we propose to allow an array of `select()` for all operations which accept a select statement (`tp`, `data`, `psxKernel`, `psxStats`, `epochs`, `labnotebook`).

[1]: https://alleninstitute.github.io/MIES/SweepFormula.html#select
[2]: https://alleninstitute.github.io/MIES/file/_m_i_e_s___analysis_functions___patch_seq_8ipf.html

MichaelHuth:
- [x] implementation
    - [x] add new tag to data to skip plotting in formula plotter, skip plotting filter.ranges component
    - [x] set data types on sub waves of combined data (multiple data sets of different type)
    - [x] in upmost wave add custom legend text representation of filter.ranges data
    - [x] workaround PSQ_RHEOBASE for sweepQC, ...Dashboard.ipf L342
    - [x] proper SCI handling (setQC only)
    - [x] add to data
    - [x] review play around
    - [x] add to other select dependent operations
    - [x] add second revision sel* operations
- [x] docu
- [ ] ~~Modifying ifn SF documentation needs to wait for https://github.com/AllenInstitute/MIES/pull/2023~~
- [x] After documentation review from TB, the SF notebook can be updated
- [x] tests

close #2012


MichaelHuth:
@t-b I think the SF changes could be technically tested. I have not implemented the Tagging/Plotter/Legend changes yet though.

Here are some examples:
```
sel1 = select(channels(AD), sweeps(1,2), selectVis(all), selectIVSCCsweepQC(passed))

sel2 = select(channels(), sweeps(), selectVis(all), selectIVSCCsweepQC(passed))

rng = selectRange(epochs("E1", $sel1))

ep = epochs("E1", $sel1)
lnb = labnotebook(ADC, [$sel1,$sel1])
sel = select($sel1, $sel1)

#tp = tp(tpbase(), [$sel1, $sel1])

sel3 = select($sel1, selectRange($ep + [0, 0]))

#data([$sel1, $sel2])
#data($sel1)
#$lnb

#data([select(), 10]) # fails

#data([select(), abc]) # fails

data([select(), $sel1])

```

Arrays for the select argument are allowed for tp, labnotebook, data.
epochs takes only a single select argument.
Since the argument order of select is now free, it was not possible to replace sweeps() with e.g. [1,2]. Thus, sweeps() takes now sweep numbers as arguments like sweeps(1, 2).

Rules:
- all new select*() operation can appear multiple times in select(...) except selectRange(...). selectRange(...) can appear only once.
- all select*() operations in select are logical ANDed (e.g. selectVis(all), selectVis(displayed) -> displayed)
- if select(...) has another select(...) as argument, then all unset filter properties are treated as 'allow all'
- if select(...) has no other select(...) as argument it uses the default settings as before, e.g. `select()`
- specifying selects as arrays for data means `data(select())` is the same as `data([select()])` and can be furhter extended with `data([$sel0, $sel1, $sel2, ...])` where each select is treated independently with its attributed ranges.
- epochs takes only a single select argument (we can discuss if there is a use case for select arrays here)

The two expressions with the fails comment relate to the adapted formula executor.

MichaelHuth:
I am currently looking into an issue where:
```
ep = epochs(E1, select())
selectRange($ep + [0, 0]) # works
selectRange(epochs(E1, select()) + [0, 0]) # fails parsing
```

t-b:
@timjarsky Ready for playing around.

t-b:
Tim and I played around today. And it is very nice!

I have compiled a list of items to look at:

- [ ]
  BSP_CheckProc_ChangedSetting(...)#L1293 [MIES_BrowserSettingsPanel.ipf]
  UpdateSweepPlot(...)#L6407 [MIES_MiesUtilities.ipf]
  SB_UpdateSweepPlot(...)#L357 [MIES_AnalysisBrowser_SweepBrowser.ipf]
  PostPlotTransformations(...)#L3970 [MIES_MiesUtilities.ipf]
  SF_Update(...)#L2764 [MIES_SweepFormula.ipf]
  PGC_SetAndActivateControl(...)#L231 [MIES_ProgrammaticGUIControl.ipf]
  SF_button_sweepFormula_display(...)#L2812 [MIES_SweepFormula.ipf]
  SF_ExecuteVariableAssignments(...)#L5990 [MIES_SweepFormula.ipf]
  SF_FormulaExecutor(...)#L1160 [MIES_SweepFormula.ipf]
  SF_OperationSelect(...)#L4925 [MIES_SweepFormula.ipf]
  JWN_SetStringInWaveNote(...)#L239 [MIES_JSONWaveNotes.ipf]
    Time: 2024-06-07T15:20:03+02:00
    Experiment: Basic (Packed)
    Igor Pro version: 9.0.6.1 (56657)

  The line

	JWN_SetStringInWaveNote(selectResult, SF_META_DATATYPE, SF_DATATYPE_SELECT)

  in SF_OperationSelect needs to check if selectResult is null.

- [ ] selectVis should have a default argument value of "displayed"
- [ ] selectCM should have a default argument value of "all"
- [ ] selectCM(all) does not select anything
- [ ] Change prefix from sub-operations from select to sel to make it shorter
- [ ] Prefix channels and sweeps with sel
- [ ] Typo: sonly
- [ ] SFH_IsValidSingleSelection: Please add a generic function which returns a wave with matches given a
      wildcard expresssion. There is nothing builtin into IP IIRC only stringmatch
      and ListMatch.

```
	for(matchSet : filter.stimsets)
		if(stringmatch(setName, matchSet))
			found = 1
			break
		endif
	endfor
	if(!found)
		return 0
	endif
```

- [ ] SFH_IsValidSingleSelection: Can we only fetch the stimset name if
      required? For the standard case of match pattern `*` we could avoid that.
- [ ] Allow `selectRange()` meaning the full range
- [ ] select(channels(), channels(AD), ...) does not work as expected
- [ ] Does it make sense to move all operations accepting data operation as input to select arrays?
- [ ] Rename data operation to plot ?! Or better names?

t-b:
@MichaelHuth CI will not run when you have merge conflicts.

t-b:
@MichaelHuth How is the `cursor` operation handled by the changes here? Would you do `selectRange(cursors(A, B))`?

t-b:
@timjarsky Does it make sense to add another selector `selExp` to select the experiment name? This would come in handy if you have sweeps from different experiments in one sweepbrowser. And what about `selDev` for selecting the device?

timjarsky:
Yes, those all make sense. Thanks @t-b

MichaelHuth:
> @MichaelHuth How is the `cursor` operation handled by the changes here? Would you do `selectRange(cursors(A, B))`?

cursors is not changed and returns an untyped array.

selRange returns datasets of ranges and has a datatype that allows it to be used with select as the datatype enables to have unordered select arguments.
So `selrange(cursors(A,B))` gets the cursor range. You can not use `cursors` as replacement for `selrange`. And you can only use two cursors in the argument of cursors inside selrange because selrange accepts only arrays with two elements (start / end).

Options would be:
- give cursors the same data type as selrange and change it to return a dataset. This would allow cursors to be used as replacement for selrange (with exactly two cursors specified).

t-b:
> So selrange(cursors(A,B)) gets the cursor range.

That's good enough.

MichaelHuth:
Add wildcard support for `selexp`, `seldev`. Accept only a single match.

selci:
- first get N x 3 wave with selections
- then use sweepNr, channelType, ChannelNumber with SCI to fill up missing SCI sweeps numbers (they have the same channelType/channelNumber)

t-b:
I've tried playing around, but I get an assertion when loading data from the analysisbrowser:

```
PGC_SetAndActivateControl(...)#L231 [MIES_ProgrammaticGUIControl.ipf]
AB_ButtonProc_LoadSweeps(...)#L2749 [MIES_AnalysisBrowser.ipf]
PGC_SetAndActivateControl(...)#L231 [MIES_ProgrammaticGUIControl.ipf]
BSP_ButtonProc_ChangeSweep(...)#L1485 [MIES_BrowserSettingsPanel.ipf]
UpdateSweepPlot(...)#L6410 [MIES_MiesUtilities.ipf]
SB_UpdateSweepPlot(...)#L374 [MIES_AnalysisBrowser_SweepBrowser.ipf]
PostPlotTransformations(...)#L3978 [MIES_MiesUtilities.ipf]
LayoutGraph(...)#L2561 [MIES_MiesUtilities.ipf]
TUD_GetUserDataAsWave(...)#L177 [MIES_TraceUserData.ipf]
GetSetIntersection(...)#L3876 [MIES_Utilities.ipf]
  Time: 2024-06-26T13:26:17+02:00
  Experiment: Experiment (Packed)
  Igor Pro version: 9.0.6.1 (56657)
```

The bug can also be reproduced with `runwithOpts(testsuite = "UTF_TraceUserData.ipf")` in Basic.pxp.

MichaelHuth:
> I've tried playing around, but I get an assertion when loading data from the analysisbrowser:

That was some fallout from the FindIndizes change to double precision as TUD also uses the combination with GetSetIntersection. I have fixed that.

I have added now selsci and selrac.

Currently the performance bottleneck is FindDimLabel is GetLastSetting.
It is called very often due to SF_GetSelectData calling GetActiveChannels that iterates over all channels of the type.

MichaelHuth:
Ready for another try.

I checked also the GetLastSetting DimLabel speed  up and for the select case I tried it shows a 20% - 25% performance increase.

t-b:
>  Does that make sense or should it also add the corresponding other channeltype / channelnumber (i.e. always ADC + DAC) ?

Yes that makes sense.

t-b:
> I checked also the GetLastSetting DimLabel speed up and for the select case I tried it shows a 20% - 25% performance increase.

Is that percent or percentage points?

MichaelHuth:
> > I checked also the GetLastSetting DimLabel speed up and for the select case I tried it shows a 20% - 25% performance increase.
> 
> Is that percent or percentage points?

Results from my "test" in seconds:
|     | FindDimlabel | FindValue |
| --- | ------------ | --------- |
| RAC | 5.7          | 4.25      |
| SCI | 0.3          | 0.24      |

t-b:
Topics for next chat:

```
# - Add selSetCycleCount, selSetSweepCount
# - Add a way to select the n'th SCI/RAC, do it with selSCIIndex(n) and selRACIndex(n)? Output range of valid sci/rac indizes on error message
# - Rename selSCI/selRAC to selExpandSCI/selExpandRAC?!
```

t-b:
I've hacked my way around the assertion.
- Please regenerate the databrowser macro to avoid the spurious listbox in the main window. The regeneration code is now fixed in main.
- Using Gad2-IRES-Cre;Ai14-709273.06.02.02.nwb and loading all sweeps in a sweepbrowser and doing

```
sel = select(selchannels(AD), selSCIIndex(0), selvis(all), selSweeps())
dat = data($sel)

$dat
```

I get a bunch of sweeps where I only have expected sweep 0. The result for `selRACIndex` does also not look right.

selSetSweepCount/setlSetCycleCount are working nicely.

MichaelHuth:
@t-b Good for another try.

I fixed the algorithm for the SCI/RAC index (I hope).
I could not reproduce the ASSERTion on loading data into the sweepbrowser.

t-b:
Some more testing:

Gad2 nwb with all sweeps loaded:

-

  ```
  selb = select(selchannels(AD), selvis(all), selSweeps())
  selc = select($selb, selstimset("X1PS_SubThres*"), selSCIIndex(0), selRACIndex(1))

  data($selc)
  ```

  gives an RTE in SF_GetSelectDataWithRACorSCIIndex

  as cycleIdsZapped in 

  WAVE cycleIdsZapped = ZapNaNs(cycleIds)

  is null.

-

  ```
  sela = select(selchannels(AD), selSCIIndex(5), selvis(all), selSweeps())
  selb = select(selchannels(AD), selRACIndex(5), selSCIIndex(0), selvis(all), selSweeps())
  ```

  works as expected.

- 

Databrowser:

- SF_GetSelectDataWithRACorSCIIndex


	expNames[]             = SelectString(isSweepBrowser, DB_EXPNAME_DUMMY, sweepMap[selectData[p][%SWEEPMAPINDEX]][%FileName])
  

  gives an RTE if used in the databrowser. Reason is that SelectString always evaluates all arguments.


Pvalb nwb with all sweeps loaded:

- 

  ```
  sela = select(selchannels(AD), selvis(all), selSweeps(), selstimset("X2LP_Search*"))
  selb = select($sela, selSCIIndex(0))

  data($sela)

  and

  data($selb)
  ```

  $selb is empty. If I move the selstimset from sela to selb it works.


- 

  ```
  selc = select(selchannels(AD), selvis(all), selSweeps(), selstimset("X2LP_Search*"), selSCIIndex(2), selExpandRAC())

  seld = select(selchannels(AD), selvis(all), selSweeps(129), selExpandRAC())

  sele = select(selchannels(AD), selvis(all), selSweeps(129), selstimset("X2LP_Search*"), selExpandRAC())
  ```

  IMHO these three selections should all give sweeps 129 - 142.

MichaelHuth:
rebased

t-b:
Reassigning so that Tim can test it.

timjarsky:
@MichaelHuth, what is the syntax for wildcard use with `selStimset`? Thank you.

Packages/doc/SweepFormula.rst
```diff
@@ -653,64 +927,71 @@ data
...
 
 The `data` operation is the core of the `SweepFormula` library. It returns sweep data from *MIES*.
-It can be called in two variants:
 
 .. code-block:: bash
 
-   data(array range[, array selectData])
+   data(selectData)
+
+   data([selectData, selectData, ...])
```
t-b, line 936: Marker for chat.

MichaelHuth:
> @MichaelHuth, what is the syntax for wildcard use with `selStimset`? Thank you.

Internally StringMatch is used and the same wildcard pattern matching applies:
`DisplayHelpTopic "StringMatch"`

If no argument is given to `selstimset` then `"*"` is used internally.

timjarsky:
[C57BL6J-742780.14.03.03.zip](https://github.com/user-attachments/files/16641773/C57BL6J-742780.14.03.03.zip)

With the attached NWB file I can't get the following SF to work:
```
selCommon=select(selchannels(AD0), selvis(all), selCM(IC))
X3prelimSelect=select($selCommon, selStimset("X3*"))
X3epoch=[E0,E2] 
X3epochs=epochs($X3epoch, $X3prelimSelect)# + [[0,10],[0,0]]
selX3=select(selRange($X3epochs), $X3prelimSelect, selivsccsweepqc(passed))

data($selX3)
```
error: `selectData != ranges row number` or `plus: wave size mismatch` if `+ [[0,10],[0,0]]` is commented in

Inserting the short names directly into selX3 works. e.g., selX3=select(selRange([E0,E2]), $X3prelimSelect, selivsccsweepqc(passed))

Nearly identical code applied to a different stimulus set also works:
```
selCommon=select(selchannels(AD0), selvis(all), selCM(IC))
X1prelimSelect=select($selCommon, selStimset("X1*"))
X1epoch=[E0,E2]
X1epochs=epochs($X1epoch, $X1prelimSelect) + [[0,150],[0,0]]
selX1=select(selRange($X1epochs), $X1prelimSelect, selivsccsweepqc(passed))
```

t-b:
@timjarsky @MichaelHuth Michael is looking into the error you are seeing. I've created https://github.com/AllenInstitute/MIES/pull/2238 to enhance the error message.

MichaelHuth:
Analysis for the case:
```
selCommon=select(selchannels(AD0), selvis(all), selCM(IC))
X3prelimSelect=select($selCommon, selStimset("X3*"))
X3epoch=[E0,E2] 
X3epochs=epochs($X3epoch, $X3prelimSelect)# + [[0,10],[0,0]]
selX3=select(selRange($X3epochs), $X3prelimSelect, selivsccsweepqc(passed))

data($selX3)
```

X3prelimSelect results in sweeps selected 18, 19, 20.
Now epochs is applied trying to retrieve ranges for epoch E0 and E2 from all three sweeps. The issue here is that sweep 18 has only epoch E0 and not E2, whereas 19 and 20 have both epochs.
Thus, the X3epochs result is a dataset with 3 entries (for sweep 18, 19, 20)

1. a size (2) array with E0 range from sweep 18
2. a size (2,2) array with E0, E2 range from sweep 19
3. a size (2,2) array with E0, E2 range from sweep 20

Now for addition you write a (2,2) array that can be added to dataset 2 and 3 but not for dataset 1.

What could be written instead is something like:
`X3epochs=epochs($X3epoch, $X3prelimSelect) + dataset([0,10],[[0,10],[0,0]],[[0,10],[0,0]])`
which basically created for the addition three datasets with each of the correct data layout that allows an addition.

The error `selectData != ranges row number`:

In the line `selX3=select(selRange($X3epochs), $X3prelimSelect, selivsccsweepqc(passed))` the select results from X3prelimSelect are intersected with the select results from all sweeps where IVSCC sweep QC passed. IVSCC sweep QC did not pass for sweep 18, such that the intersection is 19, 20.

Now with the selRange in the same select statement the three datasets with epoch ranges from sweep 18, 19, 20 are included in the complete select information.
When this is evaluated by `data($selX3)` the error is raised because it has 3 datasets for ranges as input and 2 sweeps.

You might wonder what the difference is when you write `selRange([E0,E2])` directly in the select statement instead of resolving it previously through an epochs statement. When resolving through an epochs statement then the resulting data is a numeric array where e.g. E0 is resolved to e.g. [200, 300]. In the upper case the resolved data from epochs created one dataset per selected sweep and the rows in the numeric array in a dataset count the ranges. So if the sweep has E0 and E2 one gets two rows. If this is used e.g. with `selRange($X3epochs)` in select then it is only included in the complete select information. The ranged extraction of sweep data happens in the data operation.

If instead of the numeric ranges textual ranges (aka epoch names) are included in select then the actual numeric range is determined when the data operation is executed. For the upper case that means if data from sweep 18 is extracted and as range E0, E2 is specified then for E2 there is no range as E2 does not exist. Thus, Sweep 18 / E2 is skipped. So the mapping for selected sweep with given ranges is done one at a time.

When already resolving previously through epochs() to numeric ranges then the mapping must be clear before starting to iterate over the sweep selections. Because if I get 3 datasets with range specifications but have only 2 sweeps selected there is no way to know how to properly map these.

MichaelHuth:
@timjarsky 

Possible change: If no wildcard patterns are used in the textual epoch specification for epochs() then we can keep non-existing epochs as empty ranges. e.g. for sweep 18 we would return a (2,2) array where the range for E2 is [NaN,NaN]. Then in data() we skip sweep data extraction if the current range evaluated is [NaN,NaN].

If wildcards are used for epoch names in epochs we can not do that because we do not know how many matches we get for each sweep. We also do not know how the matches order up, such that a "auto indexed" addition may hit on unrealted ranges.

timjarsky:
@MichaelHuth Thanks for your detailed reply. I have not had time to digest it but will respond asap.

In the meantime, I tried using the selexp operation and found a behavior I don't understand. I'm attaching two nwb files (one here, the second file `Pvalb-IRES-Cre;Ai14-749133.08.11.01` is on the FTP because it was too big to post here), I loaded the x3* and x4* stimulus sets from each experiment into the same data browser. 

`selexp("*sst*")` returns the expected data
`selexp("*pvalb*")` returns an empty plot

[Sst-IRES-Cre;Ai14-745427.08.11.02.zip](https://github.com/user-attachments/files/16686224/Sst-IRES-Cre.Ai14-745427.08.11.02.zip)

timjarsky:
@MichaelHuth please ignore this comment https://github.com/AllenInstitute/MIES/pull/2109#issuecomment-2301144322

applolgies if you have already invested some time

t-b:
😎

timjarsky:
@t-b @MichaelHuth I've finished testing. The motivation for this PR was to facilitate the selection of baseline epochs from multiple stimulus sets, which I achieved on recent NWB2 files from the pipeline using the code below. I also tested selecting data from different NWB2 files, which worked nicely.

Three additional features that would be helpful are:

1. skipping over missing epochs - using the full stimset name is okay (rather than wildcards), especially if there were a way to pass the acquired stimulus set names over zeroMQ
2. Generate a warning when data is missing
3. Create a way to pass (or store in the lab notebook?) SF warnings over zeroMQ

```
#X0
selCommon=select(selchannels(AD0), selvis(all), selCM(IC), selivsccsweepqc(passed))
X0epoch=[E0,E1,E2]
selX0=select( selStimset("X0*"), $selCommon, selRange($X0epoch))

#X1
X1prelimSelect=select($selCommon, selStimset("X1*"))
X1epoch=[E0,E2]
X1epochs=epochs($X1epoch, $X1prelimSelect) + [[0,150],[0,0]]
selX1=select(selRange($X1epochs), $X1prelimSelect)

#X2, passing and failing sweeps
X2prelimSelect=select($selCommon, selStimset("X2*"))
selX2=select(selRange(E0), $X2prelimSelect)

#X3
X3prelimSelect=select($selCommon, selStimset("X3*"))
X3epoch=[E0,E2] 
X3epochs=epochs($X3epoch, $X3prelimSelect) + [[0,125],[0,0]]
selX3=select(selRange($X3epochs), $X3prelimSelect)

#X4 FI data
X4prelimSelect=select($selCommon, selStimset("X4*"))
X4epoch=[E0,E2] 
X4epochs=epochs($X4epoch, $X4prelimSelect) + [[0,100],[0,0]]
selX4=select(selRange($X4epochs), $X4prelimSelect)

#X5 short pulse search
X5prelimSelect=select($selCommon, selStimset("X5*"))
selX5=select(selRange(E0), $X5prelimSelect)

#X6 short pulse rheo
X6prelimSelect=select($selCommon, selStimset("X6*"))
X6epoch=[E0,E2] 
X6epochs=epochs($X6epoch, $X6prelimSelect) + [[0,100],[0,0]]
selX6=select(selRange($X6epochs), $X6prelimSelect)

#X7 Ramp
X7prelimSelect=select($selCommon, selStimset("X7*"))
selX7=select(selRange(E0), $X7prelimSelect)

data([$selX0, $selX1, $selX2,$selX3, $selX4, $selx5, $selx6, $selx7])
```

and

```
selPVAD=select(selexp("*pvalb*"), selvis(all),  selstimset("X3*", "X4*"), selchannels(AD0), selivsccsweepqc(passed), selrange(E1))
datPVAD = data($selPVAD)

selPVDA=select(selexp("*pvalb*"), selvis(all),  selstimset("X3*", "X4*"), selchannels(DA0), selivsccsweepqc(passed), selrange(E1))
datPVDA = data($selPVDA)

selsstAD=select(selexp("*sst*"), selvis(all),  selstimset("X3LP_*", "X4PS_*"), selchannels(AD0), selivsccsweepqc(passed), selrange(E1))
datsstAD = data($selsstAD)

selsstDA=select(selexp("*sst*"), selvis(all),  selstimset("X3LP_*", "X4PS_*"), selchannels(DA0), selivsccsweepqc(passed), selrange(E1))
datsstDA = data($selsstDA)

selPVRep= select(selsweeps([16]), selchannels (AD0), selvis(all), selexp("*pvalb*")) 

selsstRep= select(selsweeps([24]), selchannels (AD0), selvis(all), selexp("*sst*")) 

RepTracesEpochpv=epochs(E1, $selPVRep) +[-50,150]
RepTracesEpochsst=epochs(E1, $selsstRep) +[-50,150]

selrepPV=select(selrange($RepTracesEpochpv), $selPVRep)
selrepSST=select(selrange($RepTracesEpochsst), $selsstRep)

integrate(derivative(data($selreppv)))

with

integrate(derivative(data($selrepsst)))+100

and

merge(apfrequency($datPVAD))

vs

merge(max($datPVDA))

with

merge(apfrequency($datsstAD))

vs

merge(max($datsstDA)
```

timjarsky (dismissed):

t-b:
1+2 is moved to https://github.com/AllenInstitute/MIES/issues/2249

3 is moved to https://github.com/AllenInstitute/MIES/issues/2248

Labnotebook querying, and getting all used stimsets as well is available in https://github.com/AllenInstitute/MIES/pull/2247.

@timjarsky Can solve both issues separately and the rebase this MR on top of that? We can also chat about this today.

t-b:
> @timjarsky Can we solve both issues separately and the rebase this MR on top of that? We can also chat about this today.

Yes we take that approach.

MichaelHuth:
I looked at the coverage for the basic tests and noted a few cases that I will add to the tests (nothing special, mostly some statements to trigger SFH_ASSERTs or formulas to return no selections).

t-b:
I don't think it makes sense anymore to wait for #2023 for the ifn changes. So please make the ifn changes you want now and I adapt later in #2023.

t-b:
FYI: I'm fixing the psx tests right now.

t-b:
@MichaelHuth I've added two commits, one of them fixes psx. THis needs integration with your commits and docu fixes. Feel free to include it in your commit and drop my authorship.

The SFH_ASSERT commit was a debug help, which I think is nice to have.

t-b:
Please assing back to me once all basic tests (except TestHelpNotebook) are fixed and the psx changes are integrated.

t-b:
Documentation review 69a5f5cb1 (SFH_ASSERT: Output error message when debugging tests, 2024-09-17):

- [ ] Some calls to sweeps/channels are still present, found with `git grep -P '\b(sweeps|channels)\('`:

```
SweepFormula.rst:   time(data([0, 1000], channels(AD), sweeps())) == [0, 0.004, 0.008, 0.012, ...]
SweepFormula.rst:   data(select(selrange(E1), channels(AD)))
SweepFormula.rst:   data(select(selrange($ep), channels(AD)))
SweepFormula.rst:   data(select(selrange($ep), channels(AD)))
SweepFormula.rst:`sweeps()` there is no previous operation, thus the parameter would be `opStack=""`.
```

- [ ] (The only code comment)

Can we prefer SFH_GetArgumentAsWave instead of SFH_ResolveDatasetElementFromJSON?

$ git diff main... | grep -i "^+.*SFH_Resolve" | grep -v Function
+               WAVE chanSpec = SFH_ResolveDatasetElementFromJSON(jsonId, jsonPath, graph, SF_OP_SELECTCHANNELS, i, checkExist = 1)
+                       WAVE data = SFH_ResolveDatasetElementFromJSON(jsonId, jsonPath, graph, SF_OP_SELECTSWEEPS, i, checkExist = 1)

SFH_GetArgumentAsWave can also do the type checking.

- [ ] I'm not understanding this comment here. Shouln't it be `current clamp or voltage clamp mode`?

```
   # when used in select refers to sweep data acquired in current clamp and voltage clamp mode
   selcm(ic, vc)
```

- [ ] The operations selrac/selsci only appear in the documentation.

- [x] I would also drop the `when used in selet` in the examples.

- [x] You can't use `inf` as number, see https://github.com/AllenInstitute/MIES/issues/1863

- [x] The example does not compute for me.

```
   # extract the first pulse from TTL1 as epoch and extract the AD data in that range
   sel1 = select(selrange("E0_PT_P0"), selchannels(TTL1))
   data(select(selrange($ep), channels(AD)))
```

- [x] Very nice examples for `data`!

- [x] I've pushed some minor fixes.

t-b:
- [ ] Some low hanging test fixes:

```
$ git grep -P '\b(sweeps|channels)\(' MIES tests
MIES/MIES_SweepFormula.ipf:                  sprintf msg, "Unhandled channel type %g in channels() at position %d", channelType, j
tests/Basic/UTF_SweepFormula.ipf:    str = "c=cursors(A,B)\rC=select(channels(AD),[0,1],all)\rd=data($c,$C)\r\r$d"
tests/Basic/UTF_SweepFormula_PSX.ipf:    code = "psx(psxKernel([50, 150], select(channels(AD6), [0, 2], all)), 100, 100, 0)"
tests/HistoricData/UTF_HistoricEpochClipping.ipf:    jsonId = MIES_SF#SF_FormulaParser("data(\"Stimset;\", select(channels(AD),sweeps()))")
tests/UTF_HelperFunctions.ipf:       // create sweeps with dummy data for sweeps() operation thats called when omitting select
```

MichaelHuth:
>  I'm not understanding this comment here. Shouln't it be `current clamp or voltage clamp mode`?

My thoughts here were that internally it is logically ORed, but the equivalent in common language use is `and` to express the same.

MichaelHuth:
> The example does not compute for me.
> 
> ```python
>    # extract the first pulse from TTL1 as epoch and extract the AD data in that range
>    sel1 = select(selrange("E0_PT_P0"), selchannels(TTL1))
>    data(select(selrange($ep), channels(AD)))
> ```

(channels has to be selchannels here)

Though, the idea was to have a case as example where the range is extracted from a different channel type (TTL1) and with that range the data is extracted from channel AD.

t-b:
> > The example does not compute for me.
> > ```python
> >    # extract the first pulse from TTL1 as epoch and extract the AD data in that range
> >    sel1 = select(selrange("E0_PT_P0"), selchannels(TTL1))
> >    data(select(selrange($ep), channels(AD)))
> > ```
> 
> (channels has to be selchannels here)
> 
> Though, the idea was to have a case as example where the range is extracted from a different channel type (TTL1) and with that range the data is extracted from channel AD.

`$ep` is not defined. If you are going that route this is the same as the example below without offset right?

MichaelHuth:
> > > The example does not compute for me.
> > > ```python
> > >    # extract the first pulse from TTL1 as epoch and extract the AD data in that range
> > >    sel1 = select(selrange("E0_PT_P0"), selchannels(TTL1))
> > >    data(select(selrange($ep), channels(AD)))
> > > ```
> > 
> > 
> > (channels has to be selchannels here)
> > Though, the idea was to have a case as example where the range is extracted from a different channel type (TTL1) and with that range the data is extracted from channel AD.
> 
> `$ep` is not defined. If you are going that route this is the same as the example below without offset right?

Yes, your fix is correct.

Packages/MIES/MIES_SweepFormula_Helpers.ipf
```diff
@@ -238,7 +238,7 @@ Function/WAVE SFH_GetArgumentAsWave(variable jsonId, string jsonPath, string gra
 	if(ParamIsDefault(expectedWaveType))
 		checkWaveType = 0
 	else
-		checkWaveType = !!checkWaveType
```
t-b, line 241: Nice find!

t-b:
I'll have a look! ✨

MichaelHuth:
> * [x]  I think
> 
> ```
> sweepNo, CLAMPMODE_ENTRY_KEY, l, channelType, DATA_ACQUISITION_MODE)
> ```
> 
> in SF_GetSelectData is not doing the right thing as `l` is the index and `activeChannels[l]` the channel number.

`l == activeChannels[l]`

because from the `GetActiveChannels` documentation:
> It returns fixed size waves with the active entries having the same value as their index.

MichaelHuth:
> This does not set `clampMode` so the old value is reused in the next loop
> iteration.

Yes, but clampMode is only filtered when clampMode is assigned at this location. I adapted the code to make that more clear.

MichaelHuth:
> * SF_OperationTPIterate: You can nowadays also write `for(WAVE/WAVE/Z selectDataComp : selectDataArray)`

Ok, but I think Igor uses /Z there implicitly.

MichaelHuth:
> * [ ]  SF_OperationSelect: I don't understand why
> 
> ```
> if(selectArgPresent)
>     SF_SetSelectionFilterIncludeAll(graph, filter)
> else
>     SF_SetSelectionFilterDefaults(graph, filter)
> endif
> ```
> 
> is done after the loop. I would understand before or in the loop.

The idea is that the filter structure is first initialized with special values that indicate an uninitialized state for each entry with `SF_InitSelectFilterUninitalized`. Then the loop runs and gathers argument information that can optionally appear.
Afterwards structure entries can remain, where no argument was encountered that sets it. `SF_SetSelectionFilterDefaults` sets these entries then. The approach was chosen because depending on the arguments encountered (e.g. select()) the entries can get a different default value when not encountered as argument.

MichaelHuth:
> Maybe related
> 
> ```
> sweeps = 25...29
> 
> sela = select(selchannels(AD0), selsweeps($sweeps), selrange([E1, E2]),
> selvis(all)) selb = select(select(selchannels(AD0), selsweeps($sweeps),
> selrange([E1, E2]), selvis(all)))
> 
> data($sela)
> 
> and
> 
> data($selb)
> ```
> 
> gives a different selection for sela and selb. Data is on the FTP: Adaptive12Sept.zip

This is correct, for the `sela` case:

You get sweep / channel / range
* 25, AD0, E1
* 25, AD0, E2
* 26, AD0, E1
* 26, AD0, E2
* 27, AD0, E1
* 27, AD0, E2
* 28, AD0, E1
* 28, AD0, E2

For the `selb` case:

* 25, AD0, Inf
* 26, AD0, Inf
* 27, AD0, Inf
* 28, AD0, Inf

because from the range from the outermost select is not specified and thus `selrange()`

Equivalent would be: `selb = select(select(selchannels(AD0), selsweeps($sweeps), selvis(all)), selrange([E1, E2]))`

MichaelHuth:
> * Can we do
> 
> ```diff
> diff --git a/Packages/MIES/MIES_SweepFormula.ipf
> b/Packages/MIES/MIES_SweepFormula.ipf index defe98779..2f7dcb862 100644 ---
> a/Packages/MIES/MIES_SweepFormula.ipf +++ b/Packages/MIES/MIES_SweepFormula.ipf
> @@ -2889,8 +2889,6 @@ static Function SF_GetTPFitQuality(WAVE residuals, WAVE
> sweepData, variable begi
> 
>  static Function/WAVE SF_OperationTPIterate(string graph, WAVE/WAVE mode,
>  WAVE/WAVE/Z selectDataArray, WAVE/Z ignoreTPs, string opShort)
> 
> -    WAVE/WAVE/Z result = $""
> -
>      for(WAVE/WAVE selectDataComp : selectDataArray)
> 
>          if(!WaveExists(selectDataComp)) @@ -2903,11 +2901,6 @@ static
>          Function/WAVE SF_OperationTPIterate(string graph, WAVE/WAVE mode,
>          WAVE/WA continue endif
> 
> -        if(!WaveExists(result))
> -            WAVE/WAVE result = sweepData
> -            continue
> -        endif
> -
>          Concatenate/FREE/WAVE/NP {sweepData}, result endfor
> ```
> 
> and have fewer code?

I can remove the line `WAVE/WAVE/Z result = $""` and have the wave reference of result implicitly be handled by the Igor compiler, even if the loop in not executed.
I can not remove the if part. The reason is intricate. The behavior of Concatenate for wave reference waves depends on the current state of result. By Igorhelp it should apply the "waveWave" syntax. Where a single wave reference source wave is treated as an array of source waves. This applies if result is no wave reference wave. 
To properly use the waveWave syntax I would need to gather all results in a 1D wave reference wave where each contains wave references. This would require me to create for each entry a new single element wave reference wave.

However if result is already a wave reference wave then Concatenate works as the waves would be numeric and that is what I want here. To get this case I just need that first initialisation in the if and then for every other sweepData I can just call concatenate.

See also discussion in https://github.com/wavemetrics/igor/issues/5393

MichaelHuth:
> * [ ]  SFH_GetAnalysisFunctionType needs some documentation. It is very nice
>   though!! If you pass in the LBNs you can also make it available for all maybe
>   in MIES_MiesUtilities_Logbook.ipf. From there you can also use
>   GetHeadstageForChannel.

I move this change to the end as it makes sense to first have the generalized LNB getters utility function in place for SB/DB

t-b:
>>        SF_OperationTPIterate: You can nowadays also write for(WAVE/WAVE/Z selectDataComp : selectDataArray)
>
> Ok, but I think Igor uses /Z there implicitly.

Yes, IP does use it implicitly but for me it's better readable explicitly.

> > * [ ]   SF_OperationSelect: I don't understand why
> > 
> > ```
> > if(selectArgPresent)
> >     SF_SetSelectionFilterIncludeAll(graph, filter)
> > else
> >     SF_SetSelectionFilterDefaults(graph, filter)
> > endif
> > ```
> >   
> > is done after the loop. I would understand before or in the loop.
> 
> The idea is that the filter structure is first initialized with special values that indicate an uninitialized state for each entry with `SF_InitSelectFilterUninitalized`. Then the loop runs and gathers argument information that can optionally appear. Afterwards structure entries can remain, where no argument was encountered that sets it. `SF_SetSelectionFilterDefaults` sets these entries then. The approach was chosen because depending on the arguments encountered (e.g. select()) the entries can get a different default value when not encountered as argument.

Got it.

> > * [x]   I think
> > 
> > ```
> > sweepNo, CLAMPMODE_ENTRY_KEY, l, channelType, DATA_ACQUISITION_MODE)
> > ```
> >   
> > in SF_GetSelectData is not doing the right thing as `l` is the index and `activeChannels[l]` the channel number.
> 
> `l == activeChannels[l]`
> 
> because from the `GetActiveChannels` documentation:
> 
> > It returns fixed size waves with the active entries having the same value as their index.

Aha okay, I knew this was correct ;)

> > Maybe related
> > ```
> > sweeps = 25...29
> > 
> > sela = select(selchannels(AD0), selsweeps($sweeps), selrange([E1, E2]),
> > selvis(all)) selb = select(select(selchannels(AD0), selsweeps($sweeps),
> > selrange([E1, E2]), selvis(all)))
> > 
> > data($sela)
> > 
> > and
> > 
> > data($selb)
> > ```
> >   
> > gives a different selection for sela and selb. Data is on the FTP: Adaptive12Sept.zip
> 
> This is correct, for the `sela` case:
> 
> You get sweep / channel / range
> 
>     * 25, AD0, E1
> 
>     * 25, AD0, E2
> 
>     * 26, AD0, E1
> 
>     * 26, AD0, E2
> 
>     * 27, AD0, E1
> 
>     * 27, AD0, E2
> 
>     * 28, AD0, E1
> 
>     * 28, AD0, E2
> 
> 
> For the `selb` case:
> 
>     * 25, AD0, Inf
> 
>     * 26, AD0, Inf
> 
>     * 27, AD0, Inf
> 
>     * 28, AD0, Inf
> 
> 
> because from the range from the outermost select is not specified and thus `selrange()`
> 
> Equivalent would be: `selb = select(select(selchannels(AD0), selsweeps($sweeps), selvis(all)), selrange([E1, E2]))`

Thanks for spelling it out for me. Yes this is how we agreed it should work.

> > * Can we do
> > 
> > ```diff
> > diff --git a/Packages/MIES/MIES_SweepFormula.ipf
> > b/Packages/MIES/MIES_SweepFormula.ipf index defe98779..2f7dcb862 100644 ---
> > a/Packages/MIES/MIES_SweepFormula.ipf +++ b/Packages/MIES/MIES_SweepFormula.ipf
> > @@ -2889,8 +2889,6 @@ static Function SF_GetTPFitQuality(WAVE residuals, WAVE
> > sweepData, variable begi
> > 
> >  static Function/WAVE SF_OperationTPIterate(string graph, WAVE/WAVE mode,
> >  WAVE/WAVE/Z selectDataArray, WAVE/Z ignoreTPs, string opShort)
> > 
> > -    WAVE/WAVE/Z result = $""
> > -
> >      for(WAVE/WAVE selectDataComp : selectDataArray)
> > 
> >          if(!WaveExists(selectDataComp)) @@ -2903,11 +2901,6 @@ static
> >          Function/WAVE SF_OperationTPIterate(string graph, WAVE/WAVE mode,
> >          WAVE/WA continue endif
> > 
> > -        if(!WaveExists(result))
> > -            WAVE/WAVE result = sweepData
> > -            continue
> > -        endif
> > -
> >          Concatenate/FREE/WAVE/NP {sweepData}, result endfor
> > ```   
> >   
> > and have fewer code?
> 
> I can not remove the if part. The reason is intricate. The behavior of Concatenate for wave reference waves depends on the current state of result. By Igorhelp it should apply the "waveWave" syntax. Where a single wave reference source wave is treated as an array of source waves. This applies if result is no wave reference wave. To properly use the waveWave syntax I would need to gather all results in a 1D wave reference wave where each contains wave references. This would require me to create for each entry a new single element wave reference wave.
> 
> However if result is already a wave reference wave then Concatenate works as the waves would be numeric and that is what I want here. To get this case I just need that first initialisation in the if and then for every other sweepData I can just call concatenate.
> 
> See also discussion on WB bug #5393.
> 
> To satisfy the Igor compiler I then also need the first result initialisation line.

Sigh. You are right.

MichaelHuth:
> [ ] If you initialize filter.experimentName and filter.device to "" you don't
> need to handle them in
> SF_SetSelectionFilterIncludeAll/SF_SetSelectionFilterDefaults

Yes, but this would violate the concept applied for the arguments, that there is an uninitialized value and a default value that are different. Or we drop this concept for some arguments, though I think that would also be somewhat unexpected.

MichaelHuth:
> [ ] SF_MatchSweepMapColumn should use SF_GetSweepMapIndices

I added a second path for the wildcard matching in SF_GetSweepMapIndices. Though its not as nice, because I need the entries and not the indices in the first place.

MichaelHuth:
> * Not your code but we should prefer SFH_GetArgumentSelect to `SF_ExecuteFormula("select()", ...`

Do you mean that we should create a "fake" JSON to imitate zero arguments?

MichaelHuth:
Should be already good for another round.

Specifically: 
- SF when multi experiments / devices are loaded
- Performance

MichaelHuth:
I added some fix for PSX in the last commit to make the tests work. 

It most likely requires a proper fix there that uses a mapIndex if graph is a SB. I think also the combo key logic needs adaptation including the LNB getter.

t-b:
> I added some fix for PSX in the last commit to make the tests work.
>
> It most likely requires a proper fix there that uses a mapIndex if graph is a SB. I think also the combo key logic needs adaptation including the LNB getter.

Thanks for the hotfix. I'll do a proper one tomorrow.

t-b:
@MichaelHuth Thanks, I'll take over.

t-b:
I've adapted PSX to now correctly use the LBNs with mapIndex. For that I also pulled out the earlier adapations to now have them in only one commit.

Range-diff: [range-diff.txt](https://github.com/user-attachments/files/17368979/range-diff.txt)

Will merge once CI passes.

t-b (approved):

PR #2121, opened 2024-05-21
Extend GetSetIntersection to return indices of intersecting elements

MichaelHuth:
This is useful in an upcoming PR #2109 and probably on other locations.


t-b (approved):
LGTM.

PR #2123, opened 2024-05-22
Adaptive Supra enhancements

### LINKED ISSUE #2119: Adaptive supra: Reuse f-I data from previous adaptive supra SCI
Only the immediately preceeding adaptive supra SCI should be used but only:
- if set QC failed
- the same targetV of autobias was used
- and all analysis parameters are the same

### LINKED ISSUE #2129: Adaptive supra bugs out
Link: https://aiephys.slack.com/archives/C06P3MYEV6H/p1716920346361379

- [x] Tim will upload the file

### LINKED ISSUE #2130: Add SF operation to control formula plotter window
Use case: Sometimes you want analysis function automation browsers to have move the window to the front and sometimes not.

Name: XXX

### LINKED ISSUE #2131: Investigate adaptive supra threshold edge case
```What would adaptive supra do for cells that only have one spike at all current intensities?```

Write a test.

### LINKED ISSUE #2141: Adaptive Supra: Have SF plot with all f-I data which we are evaluating in one SF-plot
Just for the user to visualize

### PR DESCRIPTION
**Pull request by t-b**
- [x] #1894 
- [x] Reuse f-I data from previous failing adaptive SCIs
- [x] Flowchart
- [x] Raise LBN version and supra version
- [x] Needs https://github.com/AllenInstitute/MIES/pull/2100 so that the automation browser mode works properly again
- [x] Restrict survey plot to passing sweeps
- [x] Fix bug where APFrequency/DAScale values were duplicated, this could be a generic issue which will be fixed in #2153
- [x] Constant f-I data in RhSuAd currently asserts out, add optional parameter, with 1.5 as default, as a factor to the input DAScale range of the data to calculate the minimum DAScale width.
- [x] Remove NumPointsForLineFit and always take the last two points which have a different AP frequency.
- [x] Make adaptive tests pass for every commit
- [x] We want NumSweepsWithSaturation to look at RhSuAd and SCI slopes, currently we only look at SCI slopes
- [x] Test for 21767c54 (PSQ_DS_AdaptiveIsFinished: Fix requirement for multiple f-I slope QC matches, 2024-07-16)
- [x] Resolve TODOs/FIXME
- [x] Commit cleanup
- [x] Missing tests

```
PSQ_DS_AdaptiveIsFinished:

46 2925	 if(!WaveExists(sweepPassed) || DimSize(sweepPassed, ROWS) < numSweepsWithSaturation)
0  2926	   return 0
46 2927	endif

PSQ_DAScale

16 3864 if(ret)
0  3865    printf "Could not calculate DAScale step widths min/max.\r"
0  3866    ControlWindowToFront()
0  3867    return 1
16 3868 endif

14  3904 if(!WaveExists(futureDAScales))
11  3905   if(PSQ_DS_AdaptiveIsFinished(device, s.sweepNo, s.headstage, numSweepsWithSaturation, fromRhSuAd = 1))
0  3906     PSQ_ForceSetEvent(device, s.headstage)
0  3907     RA_SkipSweeps(device, Inf, SWEEP_SKIP_AUTO, limitToSetBorder = 1)
0  3908  
0  3909     WAVE/ZZ DAScales
0  3910     break
11  3911   endif
14  3912 endif
```

- [x] Q: Discuss with Tim: We currently determine passing f-I slope for each sweep after the sweep. But it could be the case that the maximum slope changes after deciding that we passed. Do we need to revise our earlier decision?
A: No revision, but our method to determine the sweeps with saturation should change. We only want to consider sweeps with passing f-I slope and passing sweep QC which have a larger DAScale value than the maxSlope sleep.
- [x] It should be the case that we always have both sweeps passing for the f-I slope. Check that and assert out otherwise
- [x] Q: Check what happens when we have the same DAScale values for a f-I slope fit. A: Don't start the measurement if we only have constant DAScale data in RhSuAd (testcase : `PS_DS_AD8`). If this happens in the adaptive suprathreshold run this is equivalent to a NaN f-I slope/offset and we just try again. This is also covered in a test.
- [x] Check that we display/test the dashboard messages in CI
- [x] Fix the fail messages to mention how many f-I slope QC sweeps we were expecting vs being present. ~~Show them in one line as they need to be consecutively (with ignoring failed sweeps).~~
- [x] Survey plot: Use DAscale as x-axis for f-I slope data (lower plot)

Requires:
- [x] #2174
- [x] #2175
- [x] #2176
- [x] #2187
- [x] #2203

Close #2119 
Close #2129
Close #2130
Close #2141
Close #2131


t-b:
@timjarsky Ready for a first test.

t-b:
https://byte-physics.de/public-downloads/aistorage/transfer/pr/feature_2123-dascale-adaptive-supra-threshold-reuse-fi-data/80b899ed7814bc3b49e8d3b3c471be16c1808cb5/2/BuildInstaller-user-assets/

t-b:
@timjarsky Ready!

t-b:
@timjarsky I've rebased against main so that #2213 is included here. Would be good to update the rig computer with the version from the latest CI.

timjarsky (approved):

PR #2153, opened 2024-06-26
Tests: Ensure that the CDF is empty after the test case

### LINKED ISSUE #2144: Global variables V_ITCError/V_XOPError are present in databrowser
Slack: https://aiephys.slack.com/archives/C01CS4T8MGA/p1717794866962749?thread_ts=1717710611.362109&cid=C01CS4T8MGA

Either the ITC XOP creates those globally or we in MIES.

t-b:
- [x] Fix all tests
- [x] Nicify commits
- [x] Introduce common function for cleaning up OperationQueue's V_flag
- [x] Check all occurences of operation queue invocations and use new cleanup function
- [x] Add tests for GetListOfObjects with typeFlag in bugfix commit
- [x] WM bugreports: 
    - [x] ~~SetOperationVar issue without registered output variable~~: Does not reproduce in MWE
    - [x] VariableList does not complain about invalid values
    - [x] Crash when trying to morph a permanent datafolder to a free datafolder

Close #2144


Packages/tests/HardwareBasic/UTF_BasicHardwareTests.ipf
```diff
@@ -1864,6 +1864,9 @@ static Function StopDAQDueToUncompiled_REENTRY([str])
 	CheckThatTestpulseRan_IGNORE(str)
 
 	CheckDAQStopReason(str, DQ_STOP_REASON_UNCOMPILED)
+
+	// remove left over from ???
```
MichaelHuth, line 1868: The reason is the following:
You fill the execution queue with a command and then another command "KillVariables/Z V_flag".
Then after the first command is executed the background task is executed before "KillVariables/Z V_flag".
Thus, this solution is rather flaky for all the reentry testcases if there is an Execute/P/Q/Z in the first part.

Packages/tests/Basic/UTF_Utils.ipf
```diff
@@ -2485,6 +2485,88 @@ Function GetListOfObjectsWorks2()
...
+	DFREF dfrDeep = $":test:test2"
+	CHECK(DataFolderExistsDFR(dfrDeep))
+
+	Make dfrDeep:wv3
+	Make dfrDeep:wv4
+	variable/G   dfrDeep:var3
+	variable/G/C dfrDeep:var4
+	string/G     dfrDeep:str2
+
+	result   = GetListOfObjects(dfr, ".*", recursive = 0, typeFlag = COUNTOBJECTS_DATAFOLDER)
```
MichaelHuth, line 2520: Wouldn't that be something for a MD test case?

MichaelHuth (changes requested):
The reentry TCs need another solution for V_flag cleanup from Execute/P/Q/Z

Packages/tests/Basic/UTF_Utils.ipf
t-b: Yeep will do.

Packages/tests/HardwareBasic/UTF_BasicHardwareTests.ipf
t-b: Change ForceCompile to not use /Z and remove above fixup.

t-b: Fixed.

Packages/tests/Basic/UTF_Utils.ipf
t-b: Fixed.

t-b:
Fixed, see the attached range diff: [range-diff.txt](https://github.com/user-attachments/files/16101869/range-diff.txt)

MichaelHuth (approved):

PR #2165, opened 2024-07-08
DAP_CheckPressureSettings: Enhance error message

t-b:
For some device types like the USB-6001 the AD channel mode is hardcoded. So users need to use a different AD channel so that we can use differential mode.

In a recent debug session this was forgotten by the main developers, so the error message does need enhancement.


t-b:
- [ ] Output the working channels for USB-6001

t-b:
Will merge once CI passes.

PR #2171, opened 2024-07-08
Add table operation in SweepFormula

t-b:


MichaelHuth:
@timjarsky 

> Also, could you make the dimension labels on the tables more informative? Thank you.

I changed the table that it shows dimension labels. The dimension labels are from the data waves that are from the result in the expression in `table(expr)`. This means that depending on that operation the resulting wave can have labels or not.
e.g. for the result from `select()` we add dimension labels, but most operations do not add dimension labels.
So tables are rather limited in that respect.

In Igor 10 for tables also tooltips work. I can retrieve the indices and the wave where the cursor is currently positioned. So it would be possible to generate some tooltip text including information present in the JSON wave note.
Would such a tooltip a viable option?

MichaelHuth:
> @MichaelHuth Chatted with Tim about this today. And we need to add a way for the user to C&P the sweepformula code which produced the table. So maybe add another column with just the SF code which produced the other columns?

We can collect the formulas that lead to a certain table window in the plotter and put this in the user data of the window. I am not so convinced that adding another column is a good solution as it contains then data that is not part of the SF output.
If the user does not need to see the formulas (or the tooltip is enough for this) then I would prefer a button (like here on github) that just puts the formulas in the clipboard.

Should this output then be just a list of formulas or should it be adapted with "with" inserted?

Variant 1:
```
formula(a)
formula(b)
```

Variant 2:
```
formula(a)
with
formula(b)
```

The latter would have the advantage that it can be pasted as is to a SF notebook again.

@timjarsky Would a button with the "put in clipboard" functionality work as well?

MichaelHuth:
@timjarsky I reported a bug to WM regarding the focussing of the subwindows when right clicking. They suggested a workaround for it and I added that here. This means that with this change it is not required now to give focus to a table subwindow before right clicking to copy the formulas to the clipboard. It will also work if the table subwindow below the mouse is not the active subwindow.

timjarsky (approved):

t-b:
@MichaelHuth One of the tests fails

```
•runwithOpts(testcase = "TestPlotting")
  Start of test "MIES with Basic"
  Entering test suite "UTF_SweepFormula.ipf"
  Entering test case "UTF_SweepFormula#TestPlotting"
  MIES BUG_TS: Encountered pending RTE: 725, ChildWindowList;Expected window name.
  BUG_TS: Should never be called during automated testing.
  MIES BUG_TS: Encountered pending RTE: 725, ChildWindowList;Expected window name.
  BUG_TS: Should never be called during automated testing.
  MIES BUG_TS: Encountered pending RTE: 725, ChildWindowList;Expected window name.
  BUG_TS: Should never be called during automated testing.
  MIES BUG_TS: Encountered pending RTE: 725, ChildWindowList;Expected window name.
  BUG_TS: Should never be called during automated testing.
  MIES BUG_TS: Encountered pending RTE: 725, ChildWindowList;Expected window name.
  BUG_TS: Should never be called during automated testing.
  MIES BUG_TS: Encountered pending RTE: 725, ChildWindowList;Expected window name.
  BUG_TS: Should never be called during automated testing.
  6 == 0: is false. Assertion "CHECK_EQUAL_VAR(bugCount_ts, 0)" failed in TEST_CASE_END_OVERRIDE➔TestHelperFunctions#CheckForBugMessages (UTF_HelperFunctions.ipf, line 24➔1069)
  Leaving test case "UTF_SweepFormula#TestPlotting"
  Failed with 1 errors
  Leaving test suite "UTF_SweepFormula.ipf"
  Test finished with 1 errors
    ▶ Assertion "CHECK_EQUAL_VAR(bugCount_ts, 0)" failed in TEST_CASE_END_OVERRIDE➔TestHelperFunctions#CheckForBugMessages (UTF_HelperFunctions.ipf, line 24➔1069)
  End of test "MIES with Basic"
```

PR #2176, opened 2024-07-10
Unify and enhance analysis parameter checking

### LINKED ISSUE #1522: Improve analysis parameter checking
Superseedes #1321 and #1390.

We currently do parts of the checks on stimset save and before DAQ (AFH_CheckAnalysisParameter) and other parts only before DAQ (DAP_CheckAnalysisFunctionAndParameter).

In addition we are inconsequent as DAP_CheckAnalysisFunctionAndParameter only checks the required parameters.

Rework the logic to allow the following:
- Check on stimset saving as well before DAQ:
  - That the info from _GetParams is valid. This should be split out into a separate function
    for better understandability.
  - Required parameters are present
  - Check that all parameters (required, optional, present) have:
    - valid names
    - valid types
    - valid content if wave
  - The types of required/optional parameters must match the type of the present parameter

t-b:
- [x] Needs #2175

Close #1522

### DIFF
--- a/Packages/MIES/MIES_AnalysisFunctionHelpers.ipf
+++ b/Packages/MIES/MIES_AnalysisFunctionHelpers.ipf
@@ -843,6 +843,7 @@ End
 /// - Valid and matching types
 /// - Non-empty waves supplied
 /// - All required parameters are present
+/// - No additional parameters are present if some are required/optional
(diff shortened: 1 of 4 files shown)


t-b:
Will merge once CI passes.

PR #2187, opened 2024-07-15
Various preparations for adaptive suprathreshold enhancements

t-b:


Packages/MIES/MIES_Utilities_WaveHandling.ipf
```diff
@@ -1249,6 +1249,35 @@ Function SearchForDuplicates(wv)
...
+	variable numPoints, i, numDuplicates, idx
+
+	numPoints = DimSize(wv, ROWS)
+	ASSERT_TS(numPoints == numpnts(wv), "Wave must be 1D")
+
+	if(numPoints < 2)
+		return $""
+	endif
+
+	Make/FREE/D/N=(numPoints) indizes = NaN
```
MichaelHuth, line 1264: For performance reasons, fill the wave in a separate line via FastOp

MichaelHuth (changes requested):
In commit message 7d7b31564682d64f31f358cac9b5f7bfabe086b7
I usually find it helpful for these fixes if there is a brief description what the old/previous behavior was for the described case.

t-b:
Thanks for the review:

> In commit message https://github.com/AllenInstitute/MIES/commit/7d7b31564682d64f31f358cac9b5f7bfabe086b7
I usually find it helpful for these fixes if there is a brief description what the old/previous behavior was for the described case

Done.

> For performance reasons, fill the wave in a separate line via FastOp

Fixed.

> This does not find neighbouring NaNs, is that intended?
> If yes then this should be mentioned in the function description (or it should be specified what "equal neighbouring points" exactly means)
>
> And add test for that.

Dooh. Thanks for reading carefully. This is now fixed.

Range diff: 
[range-diff.txt](https://github.com/user-attachments/files/16250275/range-diff.txt)

MichaelHuth (approved):
LGTM

PR #2188, opened 2024-07-16
Utils: fix function popcount for argument < 0

MichaelHuth:
- popcount was not working correctly for negative numbers
- added tests

since 6340df62 (Add function to calculate the population count, 2015-06-28)


t-b (approved):

t-b:
Needs conflict resolution.

MichaelHuth:
The failing PS_CR6 test seem to be unrelated to the popcount change as popcount is not called when this test runs. I also can not reproduce the fail locally.

t-b (approved):

Issue #2156, opened 2024-07-02
Fix electrode naming for NWB files


t-b:
GetCellElectrodeNames seems not be used for NWB writing. Add a test as well.

MichaelHuth:
It looks like electrode names are saved already.
Add test to TestNwbExportV2 in TestTimeSeries that checks attribute vs. GetCellElectrodeNames() per headstage

PR #2197, opened 2024-07-24
Adapt nwbv2 export tests to check against custom electrode names

MichaelHuth:
- previously the test checked only against the standard electrode names
- Add test with a custom electrode name set

close https://github.com/AllenInstitute/MIES/issues/2156


t-b:
- The reentry test should check that the expected number of sweeps were acquired
- And also use two headstages with differing electrode names

t-b (changes requested):
See comment

t-b (approved):

t-b:
> @t-b The only thing I noticed is that the electrode names need to be valid HDF5 identifiers.

Good find. I've created https://github.com/AllenInstitute/MIES/issues/2200 so that users can conveniently set the cell electrode name with validity checking.

Issue #2198, opened 2024-07-24
Tests failing with Sutter


MichaelHuth:
Some tests that are failing with sutter:

- CheckPulseInfoGathering: pulseInfo comparison uses magic numbers that probably depend on sample interval and these are slightly off for SUTTER
- RoundTripDepStimsetsRecursionThroughSweeps: Something with the NWB saving is wrong, as the created NWB file is more or less empty. Test fails if it tries to load this file.
- In UTF_BasicHardwareTests, UTF_TestPulseAndTPDuringDAQ some TP? related tests fail the "Empty DF" check at the end due to W_sigma still left.
- RestoreDAEphysPanel, RestoreAndSaveConfiguration, CheckIfConfigurationRestoresDAEphysWithUnassocDA
- EP_EpochTestSamplingFrequency: Sampling frequency list contains soem frequencies that are invalid for sutter (like 100 kHz)
- TPDuringDAQTPAndUnAssoc: SamplingInterval check issue
- PS_DS_AD1: Some issue with attempt to changing controls in the DAEPhys panel that are disabled when SUTTER is used, probably in test setup

PR #2216, opened 2024-08-06
Add support for saving each device to its own NWB file

### LINKED ISSUE #978: NWB Export with multiple devices issue
PXP and NWBv2:
[mies-igor.zip](https://github.com/AllenInstitute/MIES/files/6638994/mies-igor.zip)

The PXP has data from two devices each 1 sweep from HS0.

Looking at the data with the HDF5 Browser 

grafik

and the dump output

```
   SOFTLINK "electrode" {
      LINKTARGET "/general/intracellular_ephys/electrode_0"
   }

...

   SOFTLINK "electrode" {
      LINKTARGET "/general/intracellular_ephys/electrode_0"
   }
```

shows that both AD timeseries are linked to the same electrode and therefore to the same device. This is because as they use the same headstage number (electrode in NWB lingua) and we did not thought about that we need one NWB electrode for each device.

This bug is present in NWBv1 although less severe there because we also store the device name in the source attribute.

### LINKED ISSUE #2127: Add DA_Ephys checkbox to write data into separate NWB file
- Add setvariable string with pattern name, `%exp-%hw-%electrode`, (%exp: experiment name, %hw: hardware serial, %electrode: electrode name) current default being `%exp`
- Add NWB function to regenerate NWB file name
- Add GUI group for NWB settings

MichaelHuth:
close #978
close #2127


t-b:
Review:
afcd865d7 (NWB: Create separate NWB file for each device, 2024-08-06)

- [ ] The change in NWB_GenerateDeviceDescription should be its own commit. Does the
test RoundTripDepStimsetsRecursionThroughSweep from
shttps://github.com/AllenInstitute/MIES/issues/2198 then passes? If yes the
commit message should say so.

- [ ] Using chunked compression in NWB_PrepareExport needs also to be its own commit.

- [ ] CloseAllNWBFiles/CloseNWBFile need to have a NWB_ prefix.

- [ ] LoadMiesFolderFromPXP also own commit.

- [ ] ExportEpochsFromFileToDF:
`last += 1` ???

Same in TestEpochRecreationShortNames.

Is that a generic error when using BSP_FirstAndLastSweepAcquired? If yes we should do

```
for(i = first; i <= last; i += 1)
```

instead of wildly adding +1.

6dfc14e27 (Fix: RecreateMissingSweepAndConfigWaves handle non-missing waves properly, 2024-08-08)

Nice find!

b159debd7 (AB: Move AnalysisBrowser file name generation for ReExport to own function, 2024-08-08)

Juup.

f531e385d (HistoricDataTest: Add test to export NWBv1 data through AnalysisBrowser, 2024-08-08)

Good!

While playing around I found out that the "already open check" in
AB_AddExperimentEntries is not used when you D&D a file into IP as BeforeFileOpenHook
uses AB_AddFile directly. Happy to table that after this PR if that is not new.

MichaelHuth:
Regarding `BSP_FirstAndLastSweepAcquired`:

The function returns the last index inclusive.

Also the function has a bigger general issue that it is not multiple device aware. I think it actually should have a device argument and return the first / last sweep of that device. Returning that for multiple devices seems for me that it gives results that are not senseful usable anywhere.

t-b:
> Also the function has a bigger general issue that it is not multiple device aware. I think it actually should have a device argument and return the first / last sweep of that device. Returning that for multiple devices seems for me that it gives results that are not senseful usable anywhere.

Databrowsers can only be locked to one device so it works for them. And for Sweepbrowsers it can have duplicated sweep numbers for multiple devices yes.

MichaelHuth:
`TestEpochRecreationShortNames` and the generator: I rewrote the the whole test and the generator to work properly with multiple devices loaded in the SB.

I also added two utility functions to the SB for that.

Most of the SB functions are single experiment / single device and will return wrong results/data for the multi case. Also e.g. SweepFormula that uses these functions does not work correctly then.

t-b (approved):

t-b (approved):

t-b (approved):

timjarsky (approved):

PR #2219, opened 2024-08-07
Packages/doc/Doxyfile: Adapt rst aliases

t-b:
Newer doxygen versions like 1.12.0 have a different behaviour for embedding rst.

After some playing around the following changes fixed it.


t-b:
The fix for the new doxygen version does not work with old doxygen, see https://github.com/AllenInstitute/MIES/actions/runs/10290846397/job/28481771444?pr=2219.

PR #2229, opened 2024-08-15
DC: Add BorderVals check for sutter hardware

### LINKED ISSUE #2214: DC_CheckIfDataWaveHasBorderVals needs to support Sutter hardware
And also prefer HW_GetVoltageRanges.

MichaelHuth:
- cleaned up function DC_CheckIfDataWaveHasBorderVals
- use also HW_GetVoltageRange for ITC
- use the same approach for NI and Sutter
- return the proper channel index for NI (was always 0 before)
- adapt user message

close #2214

### DIFF
--- a/Packages/MIES/MIES_DataConfigurator.ipf
+++ b/Packages/MIES/MIES_DataConfigurator.ipf
@@ -1950,9 +1950,9 @@ static Function [variable result, variable row, variable column] DC_CheckIfDataW
 					return [1, V_row, V_col]
 				endif
 
-				return [0, NaN, NaN]
 			endfor
-			break
+
+			return [0, NaN, NaN]
 		case HARDWARE_SUTTER_DAC:
 			// @todo Determine what to check here
 			break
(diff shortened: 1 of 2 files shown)


t-b (dismissed):

MichaelHuth:
@t-b The failing test(s) actually have a deeper issue. One bug in main was that for NI only the first channel was checked for BorderVals because the return was inside the for loop and not outside.
The fix revealed now that actually some tests have BorderVals after the setup and thus, don't acquire.

t-b:
@MichaelHuth Thanks for digging. The commit history currently hides this bugfix. So please a separate commit with the fix and the test fixes.

MichaelHuth:
@t-b For channels with no amplifier we use a default gain factor of 1 mV/V or V/pA. With an default TP amplitude of 10 mV / -50 pA we get target DA values of 10 V / - 50 V that will always trigger the check for border values.

Changing that is also a bit incovenient then, because I need to go to the Hardware tab, set the gain there for the HS that has no amp and then again in the DA tab for the DA channel. I think in that moment the amp setting in the hardware tab should be disabled and not checked to be the same as for the DA channel. Then the user only needs to handle the DA channel gain.

Also I limited for Sutter/NI the border values check to DA channel only. Because AD is not yet measured and TTL can only have 0 or 1. That raises the question if there should be a check on the AD values after acquisition for border values. I think there should be at least a warning if the AD gain was set such way that the acquired values were capped. If there is later some data analysis run in the capped range the results are probably bogus.

t-b (approved):

PR #2230, opened 2024-08-15
Add three new ZMQ publishers for TP results / Retrieve TP results

### LINKED ISSUE #2157: Extend zeromq PUB filters for TP updates
Multiple options:
- as fast as possible
- every X seconds

Send out "Testpulse Cycle Id" with timestamps (increase to sub second accuracy) and TP index in that cycle.

This avoids client having to poll updates. Only publish updates when the testpulse is running.

Tim:
- [x] Clarify with @campagnola if we need to send out the timeseries as well in addition to the features (resistances etc.) we extracted
- [ ] Think about a way for external users to query a single stored TP

MichaelHuth:
Publishers added:
testpulse results live
testpulse results 1s update
testpulse results 5s update
testpulse results 10s update

The TP results from TP analysis are packed in a JSON in TP_TSAnalysis and published to ZMQ.
The time information when the next publishing is due is stored in the TUFXOP.

PUB_Publish was changed to be threadsafe and added an optional argument to control if the JSON should be released.

- [x] Rebase, fix conflicts and apply new linter
- [x] Merge https://github.com/AllenInstitute/ZeroMQ-XOP/pull/64
- [x] Add a new message type which sends out the raw TP AD data as well. See https://github.com/AllenInstitute/MIES/pull/2230#issuecomment-2583860680.
- [x] Fix CI (mostly done, Tim needs to approve vcredist on one of the runners)
- [x] Update ZeromMQ XOPs, see https://github.com/AllenInstitute/ZeroMQ-XOP/pull/66
- [x] Implement #2355
- [x] #2371
- [x] https://github.com/AllenInstitute/MIES/pull/2363
- [x] Include all amplifier settings in JSON data, but only the ones for the selected clamp mode, add units
- [x] Don't send out messages when using TP_TSAnalysi from SF or any other function

Close #2157


MichaelHuth:
@t-b Feedback request for general design improvements.

t-b:
General approach looks good.

Two comments:
- PUB_AddTPResultEntry: We only want a `value` object if it has a unit. See the other publishing functions for an example. The publishing docu can be generated with FetchAndParseMessage and OUTPUT_DOCUMENTATION_JSON_DUMP.

- PUB_CheckPublishingTime this should use TSD_ReadVar/TSD_WriteVar.

t-b:
Git review:

ef3f1d2fd (Util: Add two conversion function for DA/AD unit string return, 2024-08-22)

Good idea!
- [ ] s/in clampmode/on clampmode/

- [ ] Can we tweak them to also accept a channel typ parameter (XopChannelConstants) so that they also return the DA units and use these functions in GetChanAmpAssignUnit? In that way we only have a single place where we define the units depending on the clamp mode.

27dac754d (TP: Add more information that is transferred to the TP analysis thread, 2024-08-22)

Nice!

6368bbb98 (PUB: Add four zeromq publishers for TP data, 2024-08-22)

- [ ] s/jst/just/ in the commit message.

- [ ] Can you move the ZMQ_FILTER_* constants close to AUTO_TP_FILTER? And their names need to be something like "testpulse:", this is done so that people can easily subscribe to all messages from a certain area. See ZMQ_SUBSCRIBE in https://libzmq.readthedocs.io/en/latest/zmq_setsockopt.html. Maybe that needs a comment though.

- [ ] PUB_Publish changes can be their own commit.

- [ ] The test TestTPPublishing fails here with:

```
•runwithOpts(testcase = "TestTPPublishing")
  Start of test "MIES with HardwareBasic"
  Entering test suite "UTF_TestPulseAndTPDuringDAQ.ipf"
  Entering test case "TestPulseAndTPDuringDAQ#TestTPPublishing:ITC"
  Entering reentry "TestPulseAndTPDuringDAQ#TestTPPublishing_REENTRY"
  String mismatch (case sensitive):
str1: 0:0:0> h e a r t b e a t
str2: 0:0:0> t e s t p u l s e   r
: is false. Assertion "CHECK_EQUAL_STR(filter, expectedFilter)" failed in TestPulseAndTPDuringDAQ#TestTPPublishing_REENTRY➔FetchPublishedMessage (UTF_HelperFunctions.ipf, line 1426➔228)
  Encountered "AbortOnValue" Code 10009 in test case "TestPulseAndTPDuringDAQ#TestTPPublishing" (UTF_TestPulseAndTPDuringDAQ.ipf)
  Leaving test case "TestPulseAndTPDuringDAQ#TestTPPublishing:ITC"
  Failed with 2 errors
  Leaving test suite "UTF_TestPulseAndTPDuringDAQ.ipf"
  Test finished with 2 errors
    ▶ Assertion "CHECK_EQUAL_STR(filter, expectedFilter)" failed in TestPulseAndTPDuringDAQ#TestTPPublishing_REENTRY➔FetchPublishedMessage (UTF_HelperFunctions.ipf, line 1426➔228)
    ▶ Encountered "AbortOnValue" Code 10009 in test case "TestPulseAndTPDuringDAQ#TestTPPublishing" (UTF_TestPulseAndTPDuringDAQ.ipf)
  End of test "MIES with HardwareBasic"
```

The problem is likely that FetchPublishedMessage only works with a single filter. So we might need to teach FetchPublishedMessage to check fo multiple filters and return then multiple messages.

- [ ] Can we check the exact numeric values of the published data (with some tolerance) instead of using IsInteger or != NaN? The data is in the labnotebook or TPStorage (and really only these two).

- [ ] Test in UTF_ZeroMQPublishing.ipf missing.
- [ ] This test is also required to fill in the documentation for PUB_TPResult like for the other PUB_XXX functions.

- [ ] Can we name the parameters in TP_PrepareAnalysisDF, so that the code in TP_TSAnalysis is easier to grasp? See NWB_ASYNC_SerializeStruct/NWB_ASYNC_DeserializeStruct.

- [ ] GetTPResultAsyncBuffer: numCols should only be calculated if required. The normal case it that the wave exists with the correct size.

- [ ] Can we NaN the wave in the upgrade case? Just to be sure that we don't have bogus data lying around?

- [ ] The wave upgrade cleanup in GetTPStorage is welcome, but should be its own commit.

- [ ] GetTPStorage: The number of layers is different in upgrade compared to creation.
- [ ] GetTPStorage: Are the label names in layer 31 to 39 from TP_ANALYSIS_DATA_LABELS? If yes we need a comment to say so.

9c33299d7 (TP: Added two functions that allow to retrieve info about TPs, 2024-08-22)

commit message:
- [ ] s/both function/both functions/
- [ ] s/meta information/metadata/

- [ ] Should be its own commit:

> the former function TP_GetStoredTPs was renamed to TP_GetConsecutiveTPsUptoMarker
> to prevent misleading naming.

- [ ] The code to recreate the DAC data should be its own function and *not* duplicated.

- [ ] TP_GetStoredTPsFromCycle/TP_GetStoredTP: I would expect the returned metadata (tpresult, or the waves in tpsResult) to be 1D.
- [ ] Why do both functions not require a headstage parameter? If I understand the code in SCOPE_UpdateOscilloscopeData correctly a tpmarker is constant for all headstages of a given TP. Maybe make the test use two headstages with different clamp modes to verify that.

MichaelHuth:
Change TP return functions to have an headstage argument and slice also HS.
Then return tpADC, tpDAC, tpStorage slice as 1d wave.

* Add test that DimLabels of layers are correctly transferred to ROWs (if Redimension does that itself)

t-b:
@MichaelHuth Please rebase

t-b:
CI is failing.

Review:
92ad402b3 (Util: Add two conversion function for DA/AD unit string return, 2024-08-22)
557177403 (TP: Add more information that is transferred to the TP analysis thread, 2024-08-22)

Looks all good.

31d177113 (Getters: Cleanup of GetTPStorage wave upgrade, 2024-08-25)

Now looking at this change in isolation, I realize that this is not doing the right thing.

Consider a wave upgrade for version 13. The old code set layers 24 to 30. But
the new code sets layers 17 and 20 to 30. Overwriting valuable data!

28023219d (PUB: Preparation to add four zeromq publishers for TP data, 2024-08-25)
00941de29 (PUB: Add four publishers to publish TP data, 2024-08-23)

TestTPPublishing_REENTRY: 

Instead of

Make/FREE/T filters = {ZMQ_FILTER_TPRESULT_NOW, ZMQ_FILTER_TPRESULT_1S, ZMQ_FILTER_TPRESULT_5S, ZMQ_FILTER_TPRESULT_10S}

you can just write 

WAVE filters = DataGenerators#PUB_TPFilters()

00578959c (TP: Rename TP_GetStoredTPs to TP_GetConsecutiveTPsUptoMarker, 2024-08-25)

We nowadays have IsValidHeadstage.

e86dda871 (TP: Added two functions that allow to retrieve info about TPs, 2024-08-22)

Looks good!

t-b:
> @t-b I diagnosed an issue with the JSON construction for publishing in this PR that requires an update of the JSON XOP to the current version where we fixed the handling of adding values that are large integers.

Let's update the JSON XOP then, 7a3992638 (Update JSON-XOP, 2023-09-25) shows how this is done. We probably would like to update the MacOSX XOP as well.

t-b (approved):

t-b:
We return a wave reference wave from the get tp function and I was curious if we are handling recursive wave references correctly, yes we do.

```igorpro

Function Dostuff()

	make/o/WAVE/N=1 data
	make/o/WAVE/N=1 elem1
	make/o/WAVE/N=1 elem2

	data[0] = elem1
	elem1[0] = elem2
	
	print data, elem1, elem2

	print zeromq_test_serializewave(data)
End

```

gives

```
•dostuff()
  data[0]= {1.11244e+09}
elem1[0]= {1.11244e+09}
elem2[0]= {0}
  {
    "data": {
        "raw": [
            {
                "data": {
                    "raw": [
                        {
                            "data": {
                                "raw": [
                                    null
                                ]
                            },
                            "date": {
                                "modification": 1725915394
                            },
                            "dimension": {
                                "size": [
                                    1
                                ]
                            },
                            "type": "WAVE_TYPE"
                        }
                    ]
                },
                "date": {
                    "modification": 1725915394
                },
                "dimension": {
                    "size": [
                        1
                    ]
                },
                "type": "WAVE_TYPE"
            }
        ]
    },
    "date": {
        "modification": 1725915394
    },
    "dimension": {
        "size": [
            1
        ]
    },
    "type": "WAVE_TYPE"
}

```

t-b:
Fixed conflicts.

timjarsky:
@campagnola, can you confirm that this PR has the necessary functionality to pass TP data from MIES to ACQ4?

campagnola:
I will work with Ben to get this tested. Neither of us is very familiar with Igor, though; is the protocol documented somewhere?

t-b:
To test this I would do:
- Install the user installer on a rig
- Start Igor Pro
- Lock a device
- Start the TP

In a python console:
- Call the function `FFI_GetAvailableMessageFilters` via ZeroMQ, the JSON message should look like

```
{
  "version" : 1,
   "CallFunction" : {
     "name" : "FFI_GetAvailableMessageFilters"
   }
}
```

python example code is at https://github.com/AllenInstitute/ZeroMQ-XOP/tree/main/examples#python-client.
- Create a zeromq sub socket and subscribe to one or more of the topics returned from `FFI_GetAvailableMessageFilters`. The new ones are:

```
StrConstant ZMQ_FILTER_TPRESULT_NOW       = "testpulse:results live"
StrConstant ZMQ_FILTER_TPRESULT_1S        = "testpulse:results 1s update"
StrConstant ZMQ_FILTER_TPRESULT_5S        = "testpulse:results 5s update"
StrConstant ZMQ_FILTER_TPRESULT_10S       = "testpulse:results 10s update"
```

- Look at the sent messages, their layout is defined at the function comment for `PUB_TPResult`.

We already do have tests that things work like we think it should, so this is more a request to check that it works like you think it should. The idea of the PR is that programs interfacing with MIES (aka ACQ4) don't need to poll the TP values with `FFI_ReturnTPValues` but get a message with all the data in the requested interval.

Another thing we added is the possibility to query the stored TPs via `TP_GetStoredTP`.

t-b:
rebased against latest main.

t-b:
Fixed conflicts.

MoxenWolf:
Hey Thomas!  Here's an update.  We (Luke and I) tested the new messages being published for Test Pulses.  We would like for the raw data (binary rather than JSON if possible), including command and recording, to be included in the published test pulse message.  Ideally, this would be sent as a additional part of message like this...
```[id, json_dump, raw_data_new]``` 

Does this sound possible?

**Comment by t-b** [comment]
> We (Luke and I) tested the new messages being published for Test Pulses.

So the new publishing methods worked and TP fetching methods worked?

> We would like for the raw data (binary rather than JSON if possible), including command and recording, to be included in the published test pulse message. Ideally, this would be sent as a additional part of message like this...
[id, json_dump, raw_data_new]

Doable, for sure.

I need to look into how we could transfer binary data. Would something like BSON work or anything else from the list at [1] be good on your end? But even if we send the data in binary form this would still amount to some data e.g. with ITC hardware and default TP settings you have 2 arrays with 6666 points single precision floats which makes it ~50kB (6666 * 4 * 2) per 33ms aka 1.5MB/s. Just to mention that, as you also have to then manage that data ;)

The calculation assumes that you use the method to get an update for each TP.

[1]: https://github.com/nlohmann/json?tab=readme-ov-file#binary-formats-bson-cbor-messagepack-ubjson-and-bjdata

**Comment by ben-at-allen** [comment]
Hey @t-b .  Is there anything else you need from @campagnola or myself in order to complete this request?

**Comment by t-b** [comment]
> Hey @t-b . Is there anything else you need from @campagnola or myself in order to complete this request?

I think I'm good from you guys. I'll be chatting with Tim to see how/if we implement that.

**Comment by t-b** [comment]
Just to mention: There is already a way to query the TP data from MIES TP_GetStoredTPs.

**Comment by campagnola** [comment]
> Just to mention: There is already a way to query the TP data from MIES `TP_GetStoredTPs`.

As I mentioned above, when we tried to access TP data in this way, it appeared we would receive far fewer samples of data than expected given the duration of the test pulse * sample rate. 

That said, i order to keep TP data communication efficient, I think it would be best to just send the complete TP data with the original notification message coming from MIES. Happy to discuss in a call at some point if you'd like.

**Comment by t-b** [comment]
> Happy to discuss in a call at some point if you'd like.

Yes let's chat about that.

>> Just to mention: There is already a way to query the TP data from MIES TP_GetStoredTPs.

> As I mentioned above, when we tried to access TP data in this way, it appeared we would
> receive far fewer samples of data than expected given the duration of the test pulse * 
> sample rate.

Looks like I overlooked that. Sorry. I'll look into that, because we will send the same data out anyway.

**Comment by t-b** [comment]
Conclusion from today's meeting:
- Add new ZeroMQ pub message with the raw AD data attached as another multipart message block
- Only every TP version
- Don't require to store every TP in MIES with that

**Comment by t-b** [comment]
@MoxenWolf Ready to test. The new filter name is `testpulse:results live with data` and has three frames. Filter, json message and then the raw AD data (single precision float). As you are only using one HS the raw AD data wave is 1D.

**Comment by t-b** [comment]
@MoxenWolf I've chatted with Tim about the zeromq log files getting too large. And we decided that it is the easiest fix for now to disable logging. This is now done in this branch if you get c9fdc24c2 (GetZeroMQXOPFlags: Turn off logging, 2025-02-25) which is the latest version of this branch.

**Comment by t-b** [comment]
Include all amplifier settings in JSON data.

**Comment by t-b** [comment]
@campagnola @ben-at-allen Ready for next round of testing. The JSON now holds all amplifier settings the MIES GUI holds.

**Comment by ben-at-allen** [comment]
Hey @t-b .  I have a question about reconstructing the Command based on the metadata being sent along with the test pulse.  I'm getting a value of 20000 for the key 'tp length DAC' indicating that the test pulse is 20k in length.  However, I'm only getting back 10000 points for the test pulse itself.  When I go to reconstruct the command, I see the following metadata values that I can use...

- 'pulse start point DAC': which currently yields 9000
- 'pulse duration DAC': which currently yields 2000
- 'clamp amplitude': which currently yields 50
- 'sample interval DAC': which currently yields 0.005 (in ms)

My current goal is to reconstruct the command.  But, if I were to do this using the 10k length test pulse that I'm getting sent to us, the pulse length would start at 9k and it would last for 2k points, overshooting the 10k.  Are you sending back a scaled version of the test pulse that includes 10k points instead of 20k?  I'm sorry if this question was answered elsewhere and thank you.

**Comment by t-b** [comment]
Does the TP AD data you get look sensible? I've checked again and can't reproduce that issue. What floating point type are you expecting for the TP AD data? It is FP32.

I've used the MIES testcase TestTPPublishing which uses 2HS, sampling interval of 10 micro s, duration 10ms, baseline 35%.

With that I get a AD TP wave with 3333 rows and the following JSON

```json
  "properties": {
    "baseline fraction": {
      "unit": "%",
      "value": 35
    },
    "clamp amplitude": {
      "unit": "pA",
      "value": -50
    },
    "clamp mode": 1,
    "device": "ITC18USB_Dev_0",
    "headstage": 0,
    "pulse duration ADC": {
      "unit": "points",
      "value": 1000
    },
    "pulse duration DAC": {
      "unit": "points",
      "value": 1000
    },
    "pulse start point ADC": {
      "unit": "point",
      "value": 1166
    },
    "pulse start point DAC": {
      "unit": "point",
      "value": 1166
    },
    "sample interval ADC": {
      "unit": "ms",
      "value": 0.01
    },
    "sample interval DAC": {
      "unit": "ms",
      "value": 0.01
    },
    "time of tp acquisition": {
      "unit": "s",
      "value": 646.8000000000013
    },
    "timestamp": {
      "unit": "s",
      "value": 3826642958.501
    },
    "timestampUTC": {
      "unit": "s",
      "value": 3826635758.501
    },
    "tp cycle id": 2871429,
    "tp length ADC": {
      "unit": "points",
      "value": 3333
    },
    "tp length DAC": {
      "unit": "points",
      "value": 3333
    },
    "tp marker": 1230115666
  }
```

and when I reconstruct the TP DA with

```igorpro
Function Reconstruct()

	variable tpMarker, headstage, tpLength, pulseLength, pulseStart, samplingInterval
	string device

	// BEGIN CHANGE ME
	device = "ITC18USB_Dev_0"
	tpMarker =  1230115666
	tpLength = 3333
	pulseLength = 1000
	pulseStart = 1166
	headstage = 0
	samplingInterval = 0.01
	// END CHANGE ME

	WAVE/WAVE tpAll = TP_GetStoredTP(device, tpMarker, headstage, 0)
	Duplicate/O tpAll[0], root:tpADC/WAVE=tpADC

	Make/o/N=(tpLength) tpDAC = 0
	tpDAC[pulseStart, pulseStart + pulseLength] = 1
	setscale/P x, 0, samplingInterval, "ms", tpDAC

	Display
	appendtoGraph/R tpDAC
	appendtoGraph tpADC[*][0]
	ModifyGraph rgb(tpADC)=(0,0,65535)
End
```

and get the following plot

grafik

So for me this looks okay.

Can you post a MWE in python which shows what is not working?

ben-at-allen:
@t-b ok yeah it’s probably pulling it in right now as float 64.  I can’t access the system right now but I will verify as soon as I can.  It’s a one-liner numpy.frombuffer command so I’m sure it’s just that.

t-b:
Fixed conflicts and linted with new ipt version.

t-b:
@ben-at-allen I've fixed the unexpected TP message from the databrowser. And also added a test to catch errors like that in the future. Thanks for the report.

t-b:
@ben-at-allen You want version 5b0bbddb9 (tests/Basic: Ensure that we don't get unexpected ZeroMQ messages, 2025-04-11) as I just had to fix something.

t-b:
From https://aiephys.slack.com/archives/C07LLQEDKU0/p1744410706950439:

> Ok Mies has been updated and tested on Rig 5

So I'm merging this.

PR #2253, opened 2024-09-04
Update JSON XOP

MichaelHuth:
-one important fix here is the better handling of JSON_AddValue
 for numbers that can be added as integer.


t-b (approved):

t-b:
The failure looks real.

t-b:
Fixed the CI failure. It was related to the JSON XOP, but in very edge-case way.

t-b (approved):

PR #2290, opened 2024-10-25
Fix ReachTargetVoltage handling for too small resistance

t-b:
We used to always use 100pA if the resistance is tool small, but the code obviously only wanted to do that only on the first sweep. Fix that.


timjarsky:
Acquisition continues after required sweeps have been acquired. It's been a while, but I recall it should stop automatically.

[2025_01_09_122405_2.zip](https://github.com/user-attachments/files/18367052/2025_01_09_122405_2.zip)

t-b:
> Acquisition continues after required sweeps have been acquired. It's been a while, but I recall it should stop automatically.

That's the "new" approach we do in the modern analysis functions. ReachTargetVoltage was one of the first ones and does not follow that approach. We can fix that but this is IMHO tangent to the fix in this PR.

In your pxp I see in the history some bug messages

```
 MIES BUG_TS: Encountered pending RTE: 1321, a wave read;Index out of range for wave "_free_".
  MIES BUG_TS: Encountered pending RTE: 1321, a wave read;Index out of range for wave "_free_".
  (ITC1600_Dev_0): Skipping analysis function "ReachTargetVoltage".
  The stimset has too many sweeps, increase the size of DAScales.
  (ITC1600_Dev_0): Skipping analysis function "ReachTargetVoltage".
  The stimset has too many sweeps, increase the size of DAScales.
```

which need fixing. I tried to reproduce that here, but that did not work. So I would need a screen share for that.

timjarsky:
@t-b I also get the error in main, so I guess we designed it such that the number of sweeps needs to match the size of DAScales (not an analysis param).

```
  (ITC1600_Dev_0): Skipping analysis function "ReachTargetVoltage".
  The stimset has too many sweeps, increase the size of DAScales.
  ```

timjarsky (approved):

PR #2360, opened 2025-02-27
DAEphys: Sync MIES to MCC on checkbox click

t-b:
We are currently only syncing MIES to MCC settings when changing the headstage or clamp mode.

But we forgot to sync initially which is required if we don't change headstage or clamp mode.

Bug present since forever.

@timjarsky Found while working on the amplifier state change PUB message.


t-b:
Not yet ready. Need to force writing (`checkBeforeWrite = 0`) when enabling the checkbox.

timjarsky (approved):

Issue #2206, opened 2024-07-29
Various installer bugfixes


t-b:
- [ ] Remove suggestion for 32-bit MIES release, this is way too old
- [ ] We had a a report that the installer is hidden after calling the uninstaller. This does not reproduce here
- [ ] Remember "All user/Current user" selection from previous installation. Use registry if we already use it, otherwise a ini/json file.

t-b:
@timjarsky 

The last issue

> When installing as normal user which does not have admin rights, then elevating to an admin user and then selecting "Current user" this installs for the admin user and not the normal uesr. Might be a NSIS issue.

is the one reported on slack, https://aiephys.slack.com/canvas/C06P3MYEV6H.

When looking at the installer code I can see how this goes wrong. For a normal end-user installer we request admin execution level.

And after that there is no way to get back to the original user for an "user" installation.

This is kind of on purpose from Windows because writing per-user data from an UAC elevated installer running as adin is considered bad practice, see https://stackoverflow.com/a/55537022.

Now if we want to fix it there are several ways:
- (The officially advised way) Don't elevate the installer by default to admin. When run as non-administrator this would only allow a user installation. Vice-versa an admin run (rught click run as) installation would only allow an all-user installation. This basically makes the point of the letting the user decide moot.
- Alternatively we could gather a list of users via https://nsis.sourceforge.io/User_Management_using_API_calls#Enumerate_all_users and then let in the all-user installation case decide under which user to install the files.

t-b:
We go for option 1 (officially advised way).

MichaelHuth:
- Remove elevation request.

As user:
Show "all users" greyed out and with "must be admin" annotations
Have "current user" available (default)

As admin:
Have "all users" availabe (default)
Have "current user" available

t-b:
> The uninstaller is not signed, according to [stackoverflow.com/a/72504367](https://stackoverflow.com/a/72504367) this can with NSIS 3.08 or later be done using !uninstfinalize

Please pay attention to the existing approach in signing. We should not enter the passphrase wrong as that will brick the certificate stick quickly.

t-b:
> The uninstaller is not signed, according to [stackoverflow.com/a/72504367](https://stackoverflow.com/a/72504367) this can with NSIS 3.08 or later be done using !uninstfinalize

Let's not do that. As that would complicate building the installer for local use without signing.

MichaelHuth:
Regarding elevation request and the current user / all user options:

[...]

PR #2364, opened 2025-03-05
Installer updates

MichaelHuth:
close https://github.com/AllenInstitute/MIES/issues/2206

If calling one of the powershell scripts fail the installer will now catch that and report through Messagebox (non-silent case).
In any case it will also return errorlevel code 1 then.


MichaelHuth:
@t-b I noticed that we use the option to exit with an non-zero error level code only on one other location in the installer. I was pondering if we shouldn't actually add that to all cases where we quit early due to e.g. non-fulfilled requirements.

Currently this is only caught by the fact that an early abort leaves an incomplete installation. In this change the powershell scripts are called at the end, where it is not immediatly obvious that the installer quit early in silent mode.

t-b:
> @t-b I noticed that we use the option to exit with an non-zero error level code only on one other location in the installer. I was pondering if we shouldn't actually add that to all cases where we quit early due to e.g. non-fulfilled requirements.
> 
> Currently this is only caught by the fact that an early abort leaves an incomplete installation. In this change the powershell scripts are called at the end, where it is not immediatly obvious that the installer quit early in silent mode.

Yes let's do that.

MichaelHuth:
@t-b 
> This change breaks the installer naming, as now also the user installer has the -cis suffix.

I can not reproduce this. The `-cis` naming is done in `create-installer.sh` and there was no change for this.
The unelevated installer still has the -cis suffix, whereas the elevated does not.

t-b:
> @t-b
> 
> > This change breaks the installer naming, as now also the user installer has the -cis suffix.
> 
> I can not reproduce this. The `-cis` naming is done in `create-installer.sh` and there was no change for this. The unelevated installer still has the -cis suffix, whereas the elevated does not.

create-installer.sh has this code:

```
if [ "$#" -eq 0 ]
then
  echo "RequestExecutionLevel admin" > installer/$NSISREQUEST
else
  echo "RequestExecutionLevel user" > installer/$NSISREQUEST
  sed -i 's/\.exe/-cis.exe/' installer/$NSISOUTFILE
fi
```

so if I pass an argument I get an user installer but the name has the cis suffix.

MichaelHuth:
@t-b I added a workaround for the case Get-ProcessMitigation returns multiple blocks. The correct way would be to use the process id of the currently running Igor64 process. Though, I don't know an easy solution to get this id.

Calling "tasklist.exe" does not help much, because if there are two Igor64.exe processes I get both as result and do not know which one is the one of the Igor64.exe that I called from.

MichaelHuth:
@t-b: For the pass of "Build dev installer" and "Build user installer" the Project Settings need to be adapted. I do not have the rights to do that. The job is not just called "Build installer".

t-b:
> @t-b: For the pass of "Build dev installer" and "Build user installer" the Project Settings need to be adapted. I do not have the rights to do that. The job is now just called "Build installer".

Fixed.

t-b:
Review:
f0843d7a0 (Update NSIS to version 3.10, 2025-03-05)
25c5c1816 (Fixes in installer script for NSIS 3.10 update, 2025-03-05)
f2960b33a (Installer: Remove hint for MIES installation for 32-bit Igor Pro, 2025-03-06)
37789e012 (Installer: Bring installer to front after uninstaller ran, 2025-03-06)
2895e3134 (Installer: Remove "Permanently remove MIES" MsgBox on uninstallation, 2025-03-06)
1775958ce (Installer: Fix "uninstaller already running" error, 2025-03-06)
229a1c606 (Installer: Attempt admin uninstall only when actually admin, 2025-03-06)

All good.

4ee7dce7a (Installer: Write out configuration to json file, 2025-03-06)

The commit message still mentions that installation_configuration.json is added to the repo.

c1635057b (CI: Build release installer as unelevated version, 2025-03-07)

.github/workflows/build-release.yml still references the old asset name.

3973c3d5a (Installer: Execute the ASRL disable script again, 2025-03-07)
34b213a16 (Installer: Add handling of failed FixOffice365.ps1 call, 2025-03-10)
7cabadf71 (Installer: Return well defined error code for install fail conditions, 2025-03-11)
cf4fa4adf (Fix Installer: Disabling ASLR for the Igor process was never applied, 2025-03-12)
565b92420 (Installer: Remove FixOffice365 script, 2025-03-12)
0de1c01aa (Util: Adapt powershell call in GetASLREnabledState, 2025-03-12)

Rest looks good.

t-b (dismissed):

t-b:
Fixed conflicts and linted with new ipt version.

timjarsky:
@t-b @MichaelHuth new installer worked in the pipeline.

timjarsky (approved):

PR #1783, opened 2023-06-12
Add epoch information for unassociated DA channels + display in DB

MichaelHuth:
Checked: 
- [x] EP_CollectEpochInfo: works on basis of DA channels and already collects unassoc DA
- [x] EP_FetchEpochs: works on all DA channels
- [x] NWB_AppendSweepLowLevel: Saves already unassoc. DA channel epoch info L1232 and L1243
- [x] TestNwbExportv2: Tests already unassoc. DA channels as well L528 ff.

close https://github.com/AllenInstitute/MIES/issues/1765

PR #2385, opened 2025-04-02
Insert recreated epoch info on massexport

### LINKED ISSUE #881: Think about adding/fixing up labnotebook entries on NWB export

### LINKED ISSUE #2009: Allow recreating TTL epoch info from labnotebook entries
@MichaelHuth We don't tackle this here now.

### LINKED ISSUE #2076: Add epoch recreation export in NWB2 for mass processing
from https://github.com/AllenInstitute/MIES/pull/2041#issuecomment-2017914543

> I've chatted with Tim about this and here is our conclusion:

> Overwrite the existing epoch info in NWBv2 files, existing IPNWB functions: WriteSingleChannel, WriteEpochs, AppendEpochTable
> Should be done only on request (i.e. just implement a function)
> We will then do the epoch overwriting with an approach like MIES_MassExperimentProcessing.ipf

> We can do that in another PR given that this PR is self-contained with tests.

Conceptionally it has to be solved that NWB2 export is threadsafe and epoch recreation is not.
An idea would be to iterate through all sweeps before and cache/buffer the recreated epochs such that NWB2 export can access the data threadsafe.

MichaelHuth:
- [x] Needs https://github.com/AllenInstitute/MIES/pull/2432
- [x] Needs https://github.com/AllenInstitute/MIES/pull/2569
- [ ] Wait for epoch support in ipfx
- [ ] Test a full QC run with recreated epoch info on a pxp without epoch info

close #2076

close #881

Close #2009

### DIFF
--- a/Packages/MIES/MIES_Constants.ipf
+++ b/Packages/MIES/MIES_Constants.ipf
@@ -39,8 +39,8 @@ Constant SWEEP_EPOCH_VERSION = 9
 /// - New/Changed layers of entries
 ///
 ///@{
-Constant LABNOTEBOOK_VERSION = 86
-Constant RESULTS_VERSION     = 4
+Constant LABNOTEBOOK_VERSION = 87
+Constant RESULTS_VERSION     = 5
 ///@}
 
 /// @name Analysis function versions
@@ -2724,6 +2724,12 @@ Constant TRACECOLORMAX = 21
 ///@}
 
 StrConstant BSP_USER_DATA_SF_CONTENT_HASH = "SweepFormulaContentHash"
+
+/// Labnotebook capabilities are stored in the key wave note
+/// @anchor LabnotebookCapabilityKeys
+///@{
+StrConstant LBN_CAP_SUPPORTS_ENTRYSOURCETYPE = "SupportsEntrySourceType"
+///@}
 
 Constant HM_HASHMAP_WAVE_VERSION = 1
 
--- a/Packages/MIES/MIES_Constants.ipf
+++ b/Packages/MIES/MIES_Constants.ipf
@@ -39,7 +39,7 @@ Constant SWEEP_EPOCH_VERSION = 9
 /// - New/Changed layers of entries
 ///
 ///@{
-Constant LABNOTEBOOK_VERSION = 85
+Constant LABNOTEBOOK_VERSION = 86
 Constant RESULTS_VERSION     = 4
 ///@}
 
@@ -1043,6 +1043,7 @@ StrConstant STIMSET_SIZE_KEY              = "Stimset Size"
 StrConstant STIMSET_ERROR_KEY             = "Wavebuilder Error"
 StrConstant AUTOBIAS_PERC_KEY             = "Autobias %"
 StrConstant SWEEP_EPOCH_VERSION_ENTRY_KEY = "Epochs Version"
+StrConstant POSTPROCESSED_ENTRY_KEY       = "PostProcessed"
 
 Constant WAVEBUILDER_STATUS_ERROR = 1
 
--- a/Packages/MIES/MIES_Epochs.ipf
+++ b/Packages/MIES/MIES_Epochs.ipf
@@ -1795,7 +1795,7 @@ End
 /// @param sweepDFR        single sweep folder, e.g. for measurement with a device this wold be DFREF sweepDFR = GetSingleSweepFolder(deviceDFR, sweepNo)
 /// @param sweepNo         sweep number
 /// @returns recreated 4D epoch wave
-static Function/WAVE EP_RecreateEpochsFromLoadedData(WAVE numericalValues, WAVE/T textualValues, DFREF sweepDFR, variable sweepNo)
+Function/WAVE EP_RecreateEpochsFromLoadedData(WAVE numericalValues, WAVE/T textualValues, DFREF sweepDFR, variable sweepNo)
 
 	STRUCT DataConfigurationResult s
 	variable channelNr, plannedTime, acquiredTime, adSize, firstUnacquiredIndex
--- a/Packages/MIES/MIES_Epochs.ipf
+++ b/Packages/MIES/MIES_Epochs.ipf
@@ -1811,6 +1811,7 @@ Function/WAVE EP_RecreateEpochsFromLoadedData(WAVE numericalValues, WAVE/T textu
 
 	WAVE/T recEpochWave = GetEpochsWaveAsFree()
 	EP_CollectEpochInfoDA(recEpochWave, s)
+	EP_CollectEpochInfoTTL(recEpochWave, s)
 	EP_AddRecreatedUserEpochs(numericalValues, textualValues, sweepDFR, sweepNo, s, recEpochWave)
 
 	WAVE/Z channelDA = GetDAQDataSingleColumnWaveNG(numericalValues, textualValues, sweepNo, sweepDFR, XOP_CHANNEL_TYPE_DAC, s.DACList[0])
--- a/Packages/MIES/MIES_Epochs.ipf
+++ b/Packages/MIES/MIES_Epochs.ipf
@@ -1199,6 +1199,7 @@ static Function EP_AddEpoch(WAVE/T epochWave, variable channel, variable channel
 	ASSERT(!isNull(epTags), "Epoch name is null")
 	ASSERT(!isEmpty(epTags), "Epoch name is empty")
 	ASSERT(!isEmpty(epShortName), "Epoch short name is empty")
+	ASSERT(epBegin >= 0, "Epoch begin must be >= 0")
 	ASSERT(epBegin <= epEnd, "Epoch end is <= epoch begin")
 	ASSERT(epBegin < upperlimit, "Epoch begin is greater than upper limit")
 	ASSERT(epEnd > lowerlimit, "Epoch end lesser than lower limit")
--- a/Packages/MIES/MIES_ExperimentDocumentation.ipf
+++ b/Packages/MIES/MIES_ExperimentDocumentation.ipf
@@ -143,7 +143,7 @@ static Function ED_SetLabnotebookRowToPostProcessed(WAVE values, variable row)
 	endif
 End
 
-static Function ED_InitNewRow(WAVE values, variable rowIndex, variable sweepNo, variable entrySourceType, variable acqState)
+static Function ED_InitNewRow(WAVE values, variable rowIndex, variable sweepNo, variable entrySourceType, variable acqState, variable insertAsPostProc)
 
 	variable timestamp
 	string   timestampStr
@@ -151,7 +151,9 @@ static Function ED_InitNewRow(WAVE values, variable rowIndex, variable sweepNo, 
 	if(IsTextWave(values))
 		WAVE/T valuesT = values
 		valuesT[rowIndex][0][] = num2istr(sweepNo)
-		valuesT[rowIndex][3][] = num2istr(entrySourceType)
+		if(GetLBNCapability(values, LBN_CAP_SUPPORTS_ENTRYSOURCETYPE))
+			valuesT[rowIndex][3][] = num2istr(entrySourceType)
+		endif
 
 		valuesT[rowIndex][4][] = num2istr(acqState)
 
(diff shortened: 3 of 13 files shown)


MichaelHuth:
@t-b This PR now always recreates the epochs, whereas the issue states epochs should only be recreated if we do not have any epoch data.

However, we recently changed the epoch (re)creation to be sample exact. With the upper approach we would get a mix of old non-sample exact epochs and "new" sample exact recreated epochs possibly on exported older files.

If that mixing is ok, I add the check to only recreate if there is no epoch info already.

t-b:
> However, we recently changed the epoch (re)creation to be sample exact. With the upper approach we would get a mix of old non-sample exact epochs and "new" sample exact recreated epochs possibly on exported older files.

> If that mixing is ok, I add the check to only recreate if there is no epoch info already.

I think we want to regenerate if the epoch info is not-present or out-of-date (SWEEP_EPOCH_VERSION).

t-b:
Not a review just some drive-by-comments:
- ED_InsertPostProcessRow should work more like ED_AddEntriesToLabnotebook. This would give the same speed benefits (writing all entries in one go) but alleviate the caller of the burden of to manage the internals. Or maybe ED_AddEntriesToLabnotebook should gain a flag for adding a post processing entry?
- Your code does not handle unassociated channels, CreateLBNUnassocKey can derive an LBN key for unassociated channels. We support that since 406440a53 (Merge pull request #1783 from AllenInstitute/feature/1783-add_epoch_info_for_unassoc_DA, 2023-06-30).
- The post processed lbn entry should have the unit LABNOTEBOOK_BINARY_UNIT
- You probably want an assertion after FindRange in case this did return NaN for first and last row

t-b:
I've reapplied the ipt formatting and linting changes. But don't know howto fix the merge conflict.

MichaelHuth:
@t-b Look slike the double key check for LBN entry adds actually found a case.
In https://github.com/AllenInstitute/MIES/commit/061808a7dfabe27e2e3d816612778cf53cc49613#diff-3c3244d0695973447e26d7c941860a630c1fd265833753040b49c4b3d7c20520 I introduced "Sampling interval multiplier" at index 26, where an entry of the same key was already existing at 36 (after the change at 38).

Probable fallout:
The writing to the value source wave in DC_DocumentChannelProperty uses FindDimLabel to determine the column, so it writes to 38. 

The copying in ED_createWaveNotes runs the indexing upwards, such that the last value written to the LBN is read from source index 38.

Thus, there should be no side effects.

t-b:
Does that also resolve the #881?

Review:

59c47e742 (DAP: Change comment getter for user comment to retrieve string directly from control, 2025-04-02)

Jeep.

5c982a337 (LBN: Do labnotebook upgrade on most common MIES entry points, 2025-04-02)

- [x] The commit message is not correct:
    - We upgrade on device locking for DAEphys
    - upgrade on every sweep plot (maybe we can even optimize that?)
    - upgrade on NWB export

- [x] I don't see the upgrade on analysis browser data loading in the diff.

970d2ab90 (bugfix: Handle missing result waves gracefully in Logbookviewer, 2025-04-02)

- [x] textualValues and numericalValues do exist, but they only exist with the default entries which are removed
from LBV_CleanLogbookParamNames, right? The code change now results in the cache not being used. The problem is that
the cache is not able to store null wave refs. We could hack our way here with an empty wave and translate the
empty wave to a null wave ref in LBV_GetAllLogbookParamNames. Or we store a wave ref wave? Or?

6e1855b86 (ED: Factor out LNB row initialization to ED_InitNewRow, 2025-04-02)

Looks good.

4f015046b (ED: Remove unused variable in ED_FindIndizesAndRedimension, 2025-04-03)

Nice find. Hopefully a future IPT version can help us here as well.

6801110b2 (ED: Add check against duplicated keys in ED_CheckValuesAndKeys, 2025-04-03)

I'm trying to come up with a faster way, but fail to see one.

- [x] The only possible optimization is to not do anything if keys has only one column
- [x] Oh and missing newline after the ASSERTs before your code.

The reason I think the one column check is relevant is that all analysis function LBN values are added via
ED_AddEntryToLabnotebook which only adds one entry.

dace78715 (Bugfix: SweepSettings Key wave had duplicate key "Sampling Interval Multiplier", 2025-04-07)

Nice.

d5e9e9d6f (CA: Refactor cache key generation for LB row and index cahce to own functions, 2025-04-04)

- [x] Typo in subject line.

Otherwise nice refactoring.

403130ac5 (AFH: Add function to get headstage from channelType/channelNumber, 2025-04-02)

- [x] That should use an elseif instead of the second if or maybe even a switch?

d67d5b033 (ED: Add feature to ED_AddEntriesToLabnotebook for value insertion, 2025-04-02)

- [x] The commit message says "sweep block" but I think this is more like the entry source type block or?

- [x] ED_AddEntriesToLabnotebook I think finally deserves some proper documentation.
- [x] ED_SetLabnotebookRowToPostProcessed: We want newlines after each logical block inside functions.
- [x] ED_SetLabnotebookRowToPostProcessed: Missing assertion for out of range row.

- [x] The headstage contigency mode for PostProcessed is not DEPEND, but ALL, see ED_ParseHeadstageContigencyMode
  how that resolves. If it would be DEPEND you would not set someting in the INDEP_HEADSTAGE layer.

- [x] ED_InsertRowAfterSweepBlock: Two lines for querying and writing NOTE_INDEX please.

- [x] ED_FindIndizesAndRedimension: It does still resize the rows in values, but now you are not setting the updated
NOTE_INDEX in ED_createWaveNote/ED_createTextNotes anymore when inserting post processing. Maybe you should
pass insertAsPostProc into ED_FindIndizesAndRedimension and set the note index there if required? Moving
SetNumberInWaveNote can also be done in a prep commit. There is also a test missing which would have failed
with this bug, it should be enough if InsertRowForProstProcessingTextual/InsertRowForProstProcessingNumerical
checks NOTE_INDEX.

- [x] This should also raise LABNOTEBOOK_VERSION.

3e42a4f8c (NWB: Insert recreated epoch data into LNB on NWB_ExportAllData, 2025-04-02)

- [x] InvalidateLBIndexAndRowCache looks wrong in this commit.

- [x] InsertRecreatedEpochsIntoLBN: The two loops try to be future proof but the inner one is wrong. The
  number of channels depends on the channel type. So I would just drop that and only do it for DAC. Maybe just
  iterate over recEpochs?

- [x] If you use IsAssociatedChannel instead of IsValidHeadstage you can drop the commit as the code explains
  itself.

- [x] Using AFH_GetHeadstageFromChannelNumberAndType is 100% wrong. This can only be used during DAQ in
  analysis functions. Maybe iterate through recEpochs and then fetch the headstage? This seems to be a
  repeating error. Anything we can do to avoid that?

- [x]  I would prefer to resize keys and values for unassociated entries as gotassocValues looks odd and then
  have only one call to ED_AddEntriesToLabnotebook.

- [x] The move of NWB_WriteLabnotebooksAndComments is unmotivated, and if done should be separate.

- [x] Dedicated historic test adding post processed epoch

- [x] This misses to set SWEEP_EPOCH_VERSION in the LBN. We need that for future fixing up.

cb88330f1 (Tests: Always attempt to recreate Epochs in TestExportingDataToNWB, 2025-04-03)

- [x] Motivation?

52abcfc1b (LBN: Fix typos in description of labnotebook entries, 2025-04-04)

  Nice!

t-b:
@MichaelHuth Did you see my review in https://github.com/AllenInstitute/MIES/pull/2385#issuecomment-2787486825?

MichaelHuth:
I dropped
https://github.com/AllenInstitute/MIES/commit/403130ac5d1ed2838b6e5ec036fd61109a5005ad (AFH: Add function to get headstage from channelType/channelNumber, 2025-04-02)

in favor of a LBN based utility function based on the approach used in DataConfigurationRecreation.

MichaelHuth:
> ED_FindIndizesAndRedimension: It does still resize the rows in values, but now you are not setting the updated
NOTE_INDEX in ED_createWaveNote/ED_createTextNotes anymore when inserting post processing. Maybe you should
pass insertAsPostProc into ED_FindIndizesAndRedimension and set the note index there if required? Moving
SetNumberInWaveNote can also be done in a prep commit. There is also a test missing which would have failed
with this bug, it should be enough if InsertRowForProstProcessingTextual/InsertRowForProstProcessingNumerical
checks NOTE_INDEX.

I noticed that, but I did not evaluate this as bug. The only thing that the call to ED_FindIndizesAndRedimension does when called is to increase the row size of the wave by 1, but not the NOTE_INDEX. (The possible column increase is as regular).

The NOTE_INDEX is updated by the caller in any case.

I think ED_FindIndizesAndRedimension should be split in two parts, one that increases the COLS and another that increases the ROWS. For the insertion I only need the COLS increase as I use another function to add a row.

I am open to move the NOTE_INDEX adjustment before the row value update.

MichaelHuth:
- [x] Optimize upgrade LBN in DB. Move from DB_UpdateSweepPlot to device locking of DB.
- [x] https://github.com/AllenInstitute/MIES/commit/970d2ab90e4e9ef02fd396037649cba9a51bb1f0 (bugfix: Handle missing result waves gracefully in Logbookviewer, 2025-04-02) Put result in WaveRef wave that also null ref can be pushed to Cache and Retrieve accordingly.

MichaelHuth:
The commit e26bbaf15de5e02dd25117a28d796b94bb4aab32 that includes the TTL Epoch should also solve #881 .

MichaelHuth:
@t-b I added LBN capabilities. However, I am not sure if it makes sense to just consider it when inserting. Because if one adds a new entry it will have an EntrySourceType (behavior as before). And then the stored capabilities contradict with the actual LBN state as the LBN then starts supporting EntrySourceType after row X.

Also the epoch recreation asserts out on your PXP file because it is some cursed state that WB can not recreate the Stimset, then some Epoch lengths turn out to be zero. That causes an ASSERT in the core calculation.

t-b:
> @t-b I added LBN capabilities. However, I am not sure if it makes sense to just consider it when inserting. Because if one adds a new entry it will have an EntrySourceType (behavior as before). And then the stored capabilities contradict with the actual LBN state as the LBN then starts supporting EntrySourceType after row X.

Yes I think we have to consider it also when adding.

> Also the epoch recreation asserts out on your PXP file because it is some cursed state that WB can not recreate the Stimset, then some Epoch lengths turn out to be zero. That causes an ASSERT in the core calculation.

I'll have a look once this is ready for the next round.

t-b:
Note to self: Check that the function that creates empty recreated epochs returns `$""` for no epoch info.

t-b:
I've looked through the code and played around with it. It works here.

As you have to rebase anyway due to the merge conflict, please also revise the commit message of 1f568dd2e (LBN: Store LBN capabilities on LBN Upgrade, 2025-04-14) as that does not explain why we are adding the entry source type capability.

t-b (changes requested):
See comment

t-b (dismissed):
LGTM.

t-b:
Fixed conflicts.

t-b:
Fixed conflicts and rebased to main: 
[range-diff.patch](https://github.com/user-attachments/files/24744729/range-diff.patch)

t-b:
@MichaelHuth Please have a look at 12077260a3 (HasLBNEntrySourceTypeCapability: Fix it, 2026-01-20).

t-b:
https://github.com/AllenInstitute/MIES/pull/2385#issuecomment-3774589582

MichaelHuth:
> @MichaelHuth Please have a look at [1207726](https://github.com/AllenInstitute/MIES/commit/12077260a3947347bb3ce67dc049d7ffd09b9f82) (HasLBNEntrySourceTypeCapability: Fix it, 2026-01-20).

I think the change for treating fully empty LBNs as supporting entry source type is ok.

t-b:
> > @MichaelHuth Please have a look at [1207726](https://github.com/AllenInstitute/MIES/commit/12077260a3947347bb3ce67dc049d7ffd09b9f82) (HasLBNEntrySourceTypeCapability: Fix it, 2026-01-20).
> 
> I think the change for treating fully empty LBNs as supporting entry source type is ok.

Thx.

t-b:
ipfx issue: https://github.com/AllenInstitute/ipfx/issues/518
pynwb epochs docu: https://pynwb.readthedocs.io/en/stable/pynwb.epoch.html#pynwb.epoch.TimeIntervals.add_interval
pynwb example code in MIES: https://github.com/AllenInstitute/MIES/blob/812c062a5fa5fa6362d34d572c53117dc3e0c652/tools/nwb-read-tests/nwbv2-read-test.py#L69

PR #2407, opened 2025-04-24
AB: Add column in experiment information list box for sweep time

### LINKED ISSUE #2251: AnalysisBrowser: Add column with session start time
Use high precision sweep start of first sweep for pxp.

MichaelHuth:
A column was added at the right side of the listbox with the experiment information that shows the time when the sweep was acquired.

Another small empty column was added on the right to reserve space where the vertical scrollbar appears. Otherwise the scrollbar covers information.

The column widths were adjusted such that the horizontal size increase of the AB is minimal.

The version of the AB and listbox wave was increased.

close #2251

### DIFF
--- a/Packages/MIES/MIES_AnalysisBrowser.ipf
+++ b/Packages/MIES/MIES_AnalysisBrowser.ipf
@@ -462,6 +462,53 @@ static Function/S AB_GetSettingNumFiniteVals(WAVE wv, string device, variable sw
 	return num2str(V_npnts)
 End
 
+/// @brief Returns the Session Start Time as ISO8601 timestamp string
+///        The required data must have been loaded before.
+///
+/// @param device       device name
+/// @param firstSweepNo first sweep number of experiment
+/// @param dataFolder   dataFolder reference of the specific loaded experiment data in AB
+/// @param fileType     file type @ref AnalysisBrowserFileTypes
+///
+/// @returns session Start Time as ISO8601 timestamp string, an empty string if this information is not present
+static Function/S AB_GetSessionStartTime(string device, variable firstSweepNo, string dataFolder, string fileType)
+
+	string sessionStartTime
+	variable sweepNo, size
+
+	strswitch(fileType)
+		case ANALYSISBROWSER_FILE_TYPE_IGOR:
+			WAVE textualValues = GetAnalysLBTextualValues(dataFolder, device)
+			sessionStartTime = GetLastSettingTextIndep(textualValues, firstSweepNo, HIGH_PREC_SWEEP_START_KEY, DATA_ACQUISITION_MODE)
+			if(IsEmpty(sessionStartTime))
+				WAVE/I sweeps = GetAnalysisChannelSweepWave(dataFolder, device)
+				size = GetNumberFromWaveNote(sweeps, NOTE_INDEX)
+				if(size == 0)
+					return ""
+				endif
+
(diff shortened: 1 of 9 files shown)


t-b:
Review:
477a5df3b (NWB: Split NWB_ReadSessionStartTime into a non- and asserting function part, 2025-04-24)

Makes sense.

fb09f30d3 (AB: Add column in experiment information list box for session start time, 2025-04-24)

- [ ] Can we have the session time determination be in its own function?
- [ ] This should also use something like

```igorpro
	strswitch(fileType)
		case ANALYSISBROWSER_FILE_TYPE_IGOR:
 		    ...
			break
		case ANALYSISBROWSER_FILE_TYPE_NWBv1:
		case ANALYSISBROWSER_FILE_TYPE_NWBv2: // intended fallthrough
 		    ...
			break
		default:
			ASSERT(0, "invalid file type")
	endswitch
```

so that we catch missing spots when we ever introduce a third data source.

- [ ] Is there a cheap fallback for pxps when the labnotebook key does not exist? Creation date of the `textualValues` wave?

MichaelHuth:
> Is there a cheap fallback for pxps when the labnotebook key does not exist? Creation date of the textualValues wave?

There is AFAIK only a Wave Modification date available easily, which is not helpful as the wave is created by loading. Wave Creation date can be saved to HDF5 through the attribute "IGORWaveCreationDateLocal", though getting that from a PXP ?

t-b:
> > Is there a cheap fallback for pxps when the labnotebook key does not exist? Creation date of the textualValues wave?
> 
> There is AFAIK only a Wave Modification date available easily, which is not helpful as the wave is created by loading. Wave Creation date can be saved to HDF5 through the attribute "IGORWaveCreationDateLocal", though getting that from a PXP ?

I thought `WaveInfo` also had the creation date. But okay, so let's not show anything.

t-b (dismissed):

t-b:
> > > Is there a cheap fallback for pxps when the labnotebook key does not exist? Creation date of the textualValues wave?
> > 
> > 
> > There is AFAIK only a Wave Modification date available easily, which is not helpful as the wave is created by loading. Wave Creation date can be saved to HDF5 through the attribute "IGORWaveCreationDateLocal", though getting that from a PXP ?
> 
> I thought `WaveInfo` also had the creation date. But okay, so let's not show anything.

@MichaelHuth WM support told me that there is a `CreationDate` function. Can you try that?

t-b:
The function AB_GetSessionStartTime did not do the right thing for empty pxps. The wave for GetAnalysisChannelSweepWave is NOTE_INDEX managed, so it can have only invalid sweeps. I've added a check for that.

MichaelHuth:
> The function AB_GetSessionStartTime did not do the right thing for empty pxps. The wave for GetAnalysisChannelSweepWave is NOTE_INDEX managed, so it can have only invalid sweeps. I've added a check for that.

I think the fix is reasonable, but slightly wrong.
GetAnalysisChannelSweepWave always returns a valid wave, so the /Z and the following existence check makes no sense. I think the DimSize check needs to be replaced by a NOTE_INDEX check.

There should probably be a /Z on the resolve for firstSweepConfig.

t-b (approved):

timjarsky:
@t-b, a simple but significant improvement for users.

timjarsky (approved):

PR #2417, opened 2025-05-06
Fix dashboard with edge case data from multiple experiments

t-b:


timjarsky (approved):

timjarsky:
Thanks for the quick fix @t-b

PR #2424, opened 2025-05-09
UpdateInfoButtonHelp: Set machine-readable text for C&P

### LINKED ISSUE #2421: psx: Enhance copying data to the clipboard for import into Excel and friends
It is currently too involved to import the data put to the clipboard via PSX_CopyHelpToClipboard and UpdateInfoButtonHelp.

When we look what IP writes to the clipboard when we copy a table of the wave

```igorpro
make/o data = {{1, 2}, {3, 4}}
Edit data
```

this is

```
00000000   31 09 33 0D 0A 32 09 34 0D 0A 00 00 00 00 74 00    1.3..2.4......t.
```

which can be translated via https://www.asciitable.com to:

```
1<TAB>3<CR><LF>2<TAB>4<CR><LF>
```

So the question is if we do that manually via

```
PutScrapText "1\t3\r\n2\t4\r\n"
```

does that work with Excel? @timjarsky

t-b:
By accepting the text wave to be pretty printed we can let the visualized data and the data copied to the clipboard differ so that the latter can be more easily C&P'ed into tab calculations like Excel/Calc.

Close #2421

### DIFF
--- a/Packages/MIES/MIES_SweepFormula_PSX.ipf
+++ b/Packages/MIES/MIES_SweepFormula_PSX.ipf
@@ -2308,7 +2308,8 @@ static Function PSX_FitAcceptAverage(string win, DFREF averageDFR, WAVE eventSto
 
 	specialEventPanel = PSX_GetSpecialPanel(win)
 
-	UpdateInfoButtonHelp(specialEventPanel, "button_fit_results", PSX_AVERAGE_FIT_RESULT_DEFAULT_HELP)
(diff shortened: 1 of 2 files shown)


timjarsky:
@t-b i can't run psx because of NaNs in the deconvolution wave. So, I'm currently unable to test this PR.

t-b:
@timjarsky On the FTP in the folder `pr-2362` there are some pxps with "easy" data to playaround with.

PR #2362, opened 2025-02-28
Support multiple select statements

t-b:
So this should be ready for a first test. I've adapted `psxKernel` and `psxStats` to allow passing in multiple selections.

@timj If you encounter issues please also upload/mention the filename of the PXP with the exact SF code. So that I can reproduce that later this week.

Close #2326

PR #2427, opened 2025-05-14
PSX iterations

t-b:
- [ ] #2362
- [ ] BandPassWithRingingDetection: Generalize for all four filter types and rename
- [ ] Make `psxKernel` a pure data container
- [ ] `psx` Add `PsxIterations([iterations = 0])` operation
  - Start from user given values for `iterations > 0`
  - Median values of all events with valid fit result for amplitude and weighted Tau
  - Calculate riseTau from rise times, use avg for now; TJ comes up with something better
  - Once finished, add resulting values to wave note
  - Let psx store that in the results wave using the id


t-b:
Closing for now.

PR #2432, opened 2025-05-28
Enhance SF labnotebook reporting

### LINKED ISSUE #2392: SF labnotebook: Warn user about wrong headstage contingency mode
When a user requests an entry from a HS, but it is indep, we currently don't output anything. But we could check in the labnotebook description tables and tell her if the contingency mode is wrong.

### LINKED ISSUE #2394: labnotebook: Warn about wrong entry source type
The following SF code

```
labnotebook("User comment", select(selvis(all), selsweeps(38)))
```

does not output anything. The reason is that "User comment" has entry source type UNKNOWN_MODE. We could warn the user when she requests TP/DAQ but we have something in UNKNOWN_MODE.

t-b:
We know warn the user when the wrong mode is used.

Close #2392
Close #2394


timjarsky:
@t-b 
image
very nice!

Do we have an issue for providing an error message when a user provides a non-existent labnotebook key?

timjarsky (approved):

t-b:
> Do we have an issue for providing an error message when a user provides a non-existent labnotebook key?

No, but this was easy to implement so I just did it here.

t-b:
Rebased to adapt to new SF functions.

timjarsky (approved):

t-b:
Fixed conflicts.

PR #2439, opened 2025-06-05
Add FFI routines for datafolder export

t-b:
FFI_GetSweepBrowserTitles() returns a text wave with all sweep browser titles. 

FFI_SavePSXDataFolderToHDF5 takes a full filepath to a non-existing file and a sweep browser title. It will then save all psx datafolders it can find in it.

### DIFF
--- a/Packages/MIES/MIES_ForeignFunctionInterface.ipf
+++ b/Packages/MIES/MIES_ForeignFunctionInterface.ipf
@@ -142,3 +142,64 @@ Function/WAVE FFI_QueryLogbookUniqueSetting(string device, variable logbookType,
 
 	return settings
 End
+
+/// @brief Save the given permanent Igor Pro `datafolder` in the HDF5 format into `filepath`
+Function FFI_SaveDataFolderToHDF5(string filepath, string datafolder)
+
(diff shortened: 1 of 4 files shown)


t-b:
Of course we can also change it to allow passing a folder, we figure a file name and return that. No problem.

Packages/MIES/MIES_ForeignFunctionInterface.ipf
```diff
@@ -142,3 +142,65 @@ Function/WAVE FFI_QueryLogbookUniqueSetting(string device, variable logbookType,
...
+End
+
+static Function FFI_SaveDataFolderToHDF5Impl(string filepath, WAVE/T datafolders)
+
+	variable ref = NaN
+	variable groupID
+	string entry, cleanName
+
+	ASSERT(!FileExists(filepath), "filepath points to an existing file")
+	ASSERT(!FolderExists(filepath), "filepath points to an existing folder")
```
timjarsky, line 159: @t-b, should the error message read: filepath needs a file name? In other words, it's not possible to reach this assertion unless the file name is absent from the path?

t-b: Thanks. Rephrased.

PR #2441, opened 2025-06-06
TP_RecordTP: Fix resistance plotting in Oscilloscope graph

### LINKED ISSUE #2440: Oscilloscope not showing all Rss/Rpeak values
From https://aiephys.slack.com/archives/D6WGZV2V8/p1749159729937369:

Tim:
> I may have found a bug related to the TP precision start times - the second headstage doesn't plot the Rss or Rpeak

> Image

t-b:
In f1e2802003 (TP: Remove TPStorage field %TimeInSeconds, 2025-04-29) we removed the TimeInSeconds column from the TPStorage wave.

But this column held in the very first row and column the reference time for DeltaTimeInSeconds. This is used as x axis for the resistance plots in the oscilloscope.

The current code does therefore result in DeltaTimeInSeconds valus of NaN if a headstage was turned on later than others. Which in turn breaks the resistance display.

We now fix that by reintroducing the reference time for count zero, but this time as a wave note entry. As we are also not writing the DeltaTimeInSeconds for inactive headstage, we have to search the first active one in SCOPE_GetTPTopAxisStart.

Close #2440


timjarsky:
looks good @t-b 
image

PR #2449, opened 2025-06-11
Store hide state of windows

### LINKED ISSUE #1767: JSON Configuration: Does not remember hidden subwindows with the databrowser/sweepbrowser
The problem is that we don't record the state if a subwindow is shown or not.

t-b:
Close #1767


timjarsky:
@t-b IVSCC users will find this helpful. Thank you.

timjarsky (approved):

PR #2453, opened 2025-06-25
AnalysisBrowser: Don't support backslash escape sequences in file/folder listbox

### LINKED ISSUE #2400: Don't use backslash escape sequences in the analysis browser file listbox
With these we can't display network shares properly.

We could use special 4, but when we need to provide a height.

```igorpro
•ShowTools/A arrow
•Edit/K=1 root:MIES:Analysis:AnaBrowserFolderList
•Make/T list = {"\\a\\b\\c"}
•newpanel 
•listbox list  listWave=list 
•listbox list special={4, 10, 0}
•listbox list special={4, 12, 0}
•listbox list special={4, 15, 0}
•listbox list special={4, 25, 0}
•listbox list special={4, 20, 0}
•listbox list special={4, 20, 0}, mode=1
•Make/o/T list = {"\\\\a\\\\b\\\\c"}
```

t-b:
When you add on windows UNC path like

\\server\share\folder

in the analysis browser this get's shown as

<??>rver<??>ared<??>older

as Igor Pro treats two backslashes as starting an escape sequence for annotated text.

So let's remove escape code support for this listbox to fix the display issue.

Close #2400


timjarsky:
@t-b looks good: 
image

timjarsky (approved):

Issue #1624, opened 2023-01-24
Sweepformula error message space in DataBrowser is too small


MichaelHuth:
Currently the DataBrowser offers only a short line to display sweepformula errors. With the current and future complexity of the sweepformula module this is not enough to give the user a descriptive error message.
This requires a change of the GUI.

PR #2456, opened 2025-07-03
DB: Replace SF error SetVar/status LED with Notebook

MichaelHuth:
- The Notebook is now capable of multiple lines, error messages can be copied and the error text can be formatted
- The status LED was replaced by colored error output text
- Increased DB version to 53

- [x] Unify logic that error msg and severity gets first saved in globals and separately evaluated, one separate evaluation is the transfer to the GUI after "Display" or "Check" was pressed.

close https://github.com/AllenInstitute/MIES/issues/1624

### DIFF
--- a/Packages/MIES/MIES_BrowserSettingsPanel.ipf
+++ b/Packages/MIES/MIES_BrowserSettingsPanel.ipf
@@ -12,6 +12,7 @@
 static StrConstant EXT_PANEL_SUBWINDOW    = "BrowserSettingsPanel"
 static StrConstant EXT_PANEL_SWEEPCONTROL = "SweepControl"
 static StrConstant EXT_PANEL_SF_FORMULA   = "sweepFormula_formula"
+static StrConstant EXT_PANEL_SF_ERROR     = "sweepFormula_error"
 static StrConstant EXT_PANEL_SF_JSON      = "sweepFormula_json"
 static StrConstant EXT_PANEL_SF_HELP      = "sweepFormula_help"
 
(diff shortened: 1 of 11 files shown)


t-b:
Review:
bd07180f02 (DB: Replace SF error SetVar/status LED with Notebook, 2025-07-03)

- [ ] The new error notebooks needs to be cleared in DB_ResetAndStoreCurrentDBPanel
- [ ] A function like `SF_SetError` is misnamed if it accepts a `severity` not always being error.

- [ ] Please reduce the height of the left external panel to match the height of the graph.
grafik

This screenshot was taken after executing `Mies Panels->Advanced->Clear package settings` and then opening the databrowser.

Notebook looks good and works for my tests.

- [ ] SF_GetErrorColorsFromSeverity: We tend to return a RGBColor struct.
- [ ] BSP_GetSFError bad naming. This is also for warnings.

cd2df2e377 (SF: Separate SF error state from DB GUI, 2025-07-04)

- [ ] GetSweepFormulaParseErrorSeverity is not named correctly anymore. This is also about runtime SF errors/warnings/...

Commit message: `allow` -> `allows`

- [ ]: Documentation for GetSweepFormulaParseErrorSeverity should mention the to-be-added anchor of the flag group.

The two commits are also not separated properly.

MichaelHuth:
> * [x]  The new error notebooks needs to be cleared in DB_ResetAndStoreCurrentDBPanel

This is done in line 239:
```
	SF_SetError("", SF_MSG_OK)
	SF_DisplayErrorInGUI(device)

```

t-b (dismissed):

timjarsky (approved):

t-b:
@MichaelHuth Please rebase and fix conflicts.

MichaelHuth:
There is still an issue with the historicdata TC "TestStimsetLoading".

t-b:
> We should chat about this because the TC tries to look at Sweep 26, but there is only Sweep 0 present from the loaded NWB file.

I'll fix that then.

t-b:
@MichaelHuth See https://github.com/AllenInstitute/MIES/pull/2467

Issue #2177, opened 2024-07-11
Create flowchart of all possible SF parser states and transitions


t-b:
We can use the approach from f7751e43 (DAQ: Add acquisition state handling, 2020-11-27).

So we would write a NxN matrix where we mark a transition from row `x` to column `y` by adding a 1 into `transitions[x][y]`. This can be done during automated testing. Once this information is gathered we can generate a dot graph for it and have a look at it.

If we fixed all obvious bugs which we might have found, we can check that only found valid transitions in the tests from now on.

t-b:
We need three 2D matrizes:
- state to state
- state to action
- action to action

Add DEBUGPRINT output which outputs full state/action/last action/last calculation/XXX info.

PR #2457, opened 2025-07-04
SF: Add logging of SF parser states when DEBUGGING_ENABLED is defined

MichaelHuth:
- the initial formula gets logged
- the state, laststate, lastcalculation, action gets logged
- error message gets logged
- subsequent parser logs are aggregated in root:SFParserStateLog as text wave. Log Segments are indicated by keyword "**STARTPARSER**" in the TOKEN row. This row stores numbers otherwise.

- easily expandable, e.g. if things like recursionlevel are also required.

close https://github.com/AllenInstitute/MIES/issues/2177

### DIFF
--- a/Packages/MIES/MIES_SweepFormula.ipf
+++ b/Packages/MIES/MIES_SweepFormula.ipf
@@ -331,6 +331,53 @@ static Function/S SF_StringifyAction(variable action)
 	endswitch
 End
 
+#ifdef DEBUGGING_ENABLED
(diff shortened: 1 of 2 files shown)


t-b:
@MichaelHuth Thanks. What are the necessary steps to get the parser flowgraph for e.g. running all Basic.pxp tests?

t-b (approved):
Nicely done, works here.

PR #2465, opened 2025-07-15
MIES_AnalysisBrowser.ipf: Show files with loaded file type

t-b:
It can happen that a user selects NWB but then adds a PXP to the Analysis Browser. This then results in nothing being shown.

This is not very intuitive, so let's show the files from the currently loaded type.

- [ ] Please ensure that the branch is named correctly. See


timjarsky:
@t-b, the fix works well. What do you think about allowing the loading of both file types at once?

t-b:
> What do you think about allowing the loading of both file types at once?

Yes, let's do that.

PR #2481, opened 2025-08-06
Fix colorgroups across with statements

### LINKED ISSUE #2388: Extend color groups across with statements
In 91b5ce590 (SweepFormula: Add support for color groups, 2024-09-10) we added colors groups to have different colors for sweep data from the same channel type and number.

But we did not think about `with` needing to work as well as

```
data()

with

data()
```

creates two black traces.

t-b:
Close #2388


t-b:
@timjarsky Can I get a copolit review?

PR #2488, opened 2025-08-13
BeforeUncompiledHook: Ensure that all preemptive threads are stopped

t-b:
When you do

killdataFolder root:packages

and then uncompile/recompile the code you get the message

Function module in use by preemptive threads. Kill the thread?

The reason is that ASYNC_Stop does not stop the async threads as it's tgID is gone. The only thing we can do is stopping/killing all threads before the IP popup tells us.


timjarsky:
@t-b `killdatafolder root:packages` did not induce a threadgroup error

PR #2489, opened 2025-08-15
TP API Code

timjarsky:
- Added public API-style wrapper functions to start and stop MIES test pulses from external code.
- Implemented TP_DeviceSelectable() to verify device availability before issuing commands.
- Implemented TP_IsRunningForDevice() to check current TP state per device.
- API_StartTestPulse() now ignores start requests if the test pulse is already running.
- API_StopTestPulse() now ignores stop requests if the test pulse is already stopped.
- Unified optional API_TestPulse() entry point to handle both start and stop requests.
- Added return codes to indicate success, ignored request, or device unavailable for better external handling.

@t-b can i get your help switching the printf messages to messages that will transmit over zeroMQ?


t-b (changes requested):
See comment

PR #2490, opened 2025-08-20
tools/run-ipt.sh: Don't require the repository root as CWD

t-b:
We are now independent of the current working directory (CWD). The fixes
are having an absolute path to the config.toml and using the full path
relative to the repository root.


timjarsky:
@t-b can you remind me how to test this?

timjarsky (approved):

PR #2491, opened 2025-08-20
TS_ThreadGroupPutDFR: Make it more resilient

t-b:
We have a user report that when the (RAM) memory lows the DuplicateDataFolder call can fail which then triggers the assertion.

Unfortunately this keeps the free datafolder as CDF which fails GUI control procedures.

By using try-catch for this case we can prevent that.


MichaelHuth (approved):
Ok, though might result in a burst of ASSERTs triggered.

PR #2492, opened 2025-08-20
DAP_ChangeHeadStageMode: Stop Auto TP on clamp mode switch

t-b:
We now force Auto TP QC to failed when switching clamp mode.


t-b:
> @t-b switching from I-clamp to V-clamp while auto-tp is running does uncheck autoTP, but the TP also stops.

I'll have a look.

timjarsky:
@t-b thanks for the fix. Everything looks good now.

timjarsky (approved):

PR #2495, opened 2025-08-26
PS_ApplyStoredWindowCoordinate: Make it work with non-100% scaling

### LINKED ISSUE #2493: configuration and/or package settings issue
On rig 5 and my test setup, I'm seeing issues related to data browser initialization.

On rig 5, the windows are enormous because huge values are being set in the Settings.json. 

Image
Image
Image

On my test setup, when attempting to size all the subwindows, the sizings are not saved after closing the DB or saving the PXP file.

This is preventing data acquisition on Rig 5

t-b:
When the Display Scaling is something like 125% we would increase the
subwindows on every save/restore cycle. The reason is that the scaling
factor of 72 in PointsToPixel/PixelToPoints is only correct for 100%
scaling. In order to support the panel expansion factor which can make the scaling per panel, we now use the scaling of the panel.

Close #2493

### DIFF
--- a/Packages/MIES/MIES_GuiUtilities.ipf
+++ b/Packages/MIES/MIES_GuiUtilities.ipf
@@ -2298,12 +2298,10 @@ Function ReflowNotebookText(string win)
 
 	variable width
 
-	GetWindow $win, wsizeDC
+	GetWindow $win, wsizeRM
 	width = V_right - V_left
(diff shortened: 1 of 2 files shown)


t-b:
I'll force merge so that is finally in main even if CI is not yet working again.

PR #2504, opened 2025-09-08
MEN_OpenAboutDialog: Add comment about screen resolution calculations

t-b:
In 89a5194d2a (OpenAboutDialog: Scale the panel and notebook actions correctly, 2020-08-09) we added a rescaling of the original panel sizes and contents. On a recent discussion with Michael Huth, it came up that this code looks fishy as it has hardcoded the 96 (aka 100% scaling).

But that makes sense here as the panel was designed with 100% and then get's resized.

This might not be the best possible code, but it is working.


t-b:
@MichaelHuth pin

PR #2520, opened 2025-09-29
treewide: Update minimum IP10 version

t-b:
The old beta version has already expired so we updated to the latest one.

Will merge once CI passes.


t-b:
@timjarsky CI is back!

PR #2551, opened 2025-10-22
AI_SendToAmp: Use correct scaling for published amplifier values

### LINKED ISSUE #2544: ZMQ PUB topic "amplifier:set value" returns incorrectly scaled values
For example:

    'HoldingPotential': {'unit': 'mV', 'value': -0.07}

should have either unit=V or value=-70.

Same is true for BiasCurrent and some others, but I have not checked all keys.

t-b:
In f944d8329d (MIES_Publish.ipf: Add AMPLIFIER_SET_VALUE, 2025-02-26) we added support for publishing changed amplifier values.

And we tried hard to use the correct units as we also sent the units, e.g. pA for the bias current. But we did send the scaled values (without prefixes in MCC application scaling).

This was not caught in the test CheckIfConfigurationRestoresMCCFilterGain as the amplifier value there uses a scaling of 1, see AI_GetMCCScale() for MCC_PRIMARYSIGNALLPF_FUNC.

So we now pass the unscaled value to PUB_AmplifierSettingChange and use a amplifier value in the test with a scaling of unequal to 1.

Close #2544


campagnola:
Working on my end, thanks!

timjarsky (approved):

PR #2563, opened 2025-11-05
updates to how rig config files are recognized

timjarsky:
To work with the [new method](https://github.com/AllenInstitute/icp_rig_configurations/tree/MIES_config/Main) of distributing configuration files on production rigs, the configuration code that recognizes the _rig file was adapted such that the file name no longer needs to be an exact match with the main da_ephys config file name plus the suffix "_rig.json". Instead, the rig file name needs only contain the main da_ephys config file name.


t-b (changes requested):
See comment

PR #2579, opened 2025-11-21
treewide: Don't treat abort codes like error codes

### LINKED ISSUE #2577: Bogus error handling in ASYNC_ThreadReadOut
From: https://aiephys.slack.com/archives/C07LLQEDKU0/p1762472132055319

We can't do much about investigation the original issue, but the error handling in the catch block is bogus.

t-b:
When code aborts and this is caught in a try-catch statement Igor Pro sets the V_AbortCode variable.

This gives a value, see the IgorAbortCodes doxygen group, which denotes what type of abort was issued.

But the abort codes are nothing which can be fed to GetErrMessage as these only take error codes from runtime errors.

Close #2577


timjarsky (approved):

MichaelHuth (approved):
Looks good to me.

PR #2597, opened 2025-12-10
Allow to add files to analysis browser which got returned by a python script

### LINKED ISSUE #2591: Allow to fetch experiments paths from LIMS
The following python script should be integrated into MIES (IP10 only feature) and reworked to be able to fetch experiments paths from LIMS (a web application run at the institute, but we interface with the database directly) given a cell ID.

[limspath_from_cellname.py](https://github.com/user-attachments/files/23970742/limspath_from_cellname.py)

### LINKED ISSUE #2779: support inclusive tag selection
seltag currently behaves in the following way:
```
A sweep/experiment is only selected by `seltag` if its assigned set of tags matches the given tags exactly; tags can be assigned to experiments through the tag control side panel of the Analysis Browser.

Neither missing tags nor additional tags on the experiment result in a match.
```
I would like it to add support for additional tags.

t-b:
- [ ] Test on CI machine
  - [ ] Move to shortcuts with create-shortcut.exe in tools/clean_mies_installation.sh as otherwise uv.exe can not be found
  - [x] Get it working, currently blocked due to group policy, use https://github.com/ftnext/cdcasasagi/issues/57
- [x] Add parse format explanation to the parse window as short paragraph with link to long help
- [x] Add table format (below C&P from libreoffice)
```

<html>
<head>
	
	
	<title></title>
	<meta name="generator" content="LibreOffice 26.2.1.2 (Windows)"/>
	<style type="text/css">
		body,div,table,thead,tbody,tfoot,tr,th,td,p { font-family:"Liberation Sans"; font-size:x-small }
		a.comment-indicator:hover + comment { background:#ffd; position:absolute; display:block; border:1px solid black; padding:0.5em;  } 
		a.comment-indicator { background:red; display:inline-block; border:1px solid black; width:0.5em; height:0.5em;  } 
		comment { display:none;  } 
	</style>
	
</head>

<body>

  |   |   |   |  
-- | -- | -- | -- | --
  | Cellname | Tag | Tag |  
  | cellA | tagXX | tagZZ |  
  | cellB | tagYY | tagAA |  


</body>

</html>

```
- [x] Chat about UI with Tim:
      Result: Add button to open a generic modal window where users can paste files, folders, cellids
- [x] Can we not hardcode the credentials?
      Result: Yes don't hardcode them.
  - [x] Store credentials in packages settings? Obfuscate with base64
- [x] Fix the scripts in all necessary regards
- [x] Add uv to tools
- [x] Use uv from tools, add CI test
- [ ] Fix commits
- [ ] Check coverage for tests
- [ ] Think about "add entries" button. It does currently not play nice with remove/refresh
      -> Should clear listbox_AB_Folders before.
- [ ] Assert out on missing requirements.txt
- [ ] If some cellnames return a result and some not, return a wave with the same number of entries as was input but with empty? paths
- [ ] Ignore NWB files with filename suffixes of `_spikes`
- [ ] Rename cellname topic to id_cell_specimen and also in other places
- [ ] Rework seltag to fix #2779
  - [ ] Adapt seltag documentation (rst and ihf)
- [ ] select bug from Tim ??

Close #2591 

Close #2779


timjarsky:
@t-b Feature request: when a user provides cell IDs, also support including a tag list for each ID.

t-b:
@timjarsky An early version is pushed. The old version was d90052314a (live debugging fixes, 2026-09-11).

There are some test failures which I'll investigate later.

t-b:
@timjarsky I've pushed a fix so that the analysisbrowser does not load automatically on reopening.

t-b:
Previous version: 863d32ba33 (AB_OpenAnalysisBrowser: Don't load the files from the previous session on startup, 2026-09-15)

PR #2606, opened 2026-01-21
AI_QueryGainsFromMCC: Change logic

t-b:
This is also includes some speed improvements as we could skip selecting the amplifier in more cases.


t-b:
@timjarsky Should be fixed.

timjarsky (dismissed):

PR #2639, opened 2026-02-26
Overwrite main config settings from rig file

t-b:
rig-overwrites_rig.json disables HS1 from the main file by doing `"1": null`.

### DIFF
--- a/Packages/tests/UTF_HelperFunctions.ipf
+++ b/Packages/tests/UTF_HelperFunctions.ipf
@@ -77,7 +77,7 @@ End
 ///
 /// Works for both main and rig files.
 ///
-/// Returns the full path to the rewritten JSON configuration file the corresponding jsonID.
+/// Returns the full path to the rewritten JSON configuration file and the corresponding jsonID.
 Function [variable jsonID, string fullPath] FixupJSONConfig_IGNORE(string path, string device)
 
 	string data, fName, rewrittenConfigPath, basename


timjarsky:
@t-b I tried configuration with a main file that has two headstages and a rig file that has one heastage. I got: `  Message: "Missing "Common configuration data/Headstage Association/1/Amplifier" entry for headstage 1 in configuration."`

Should it have configured in this case?[](url)
[twoHeastate.json](https://github.com/user-attachments/files/25616984/twoHeastate.json)
[twoHeastate_rig.json](https://github.com/user-attachments/files/25616985/twoHeastate_rig.json)

timjarsky:
> @timjarsky Can you attach the modified json configuration files (rig and main) as well? Thanks.

They should be in the pxp. Are they not there?

t-b:
@timjarsky The problem is that you still enable HS1. Now we could add some magic to not enable it in this case, or you tweak your rig file to only activate HS0. Is that a good way forward?

```
diff --git a/Packages/Settings/ITC1600_0_rig.json b/Packages/Settings/ITC1600_0_rig.json
index 3554a96edf..9704d35882 100644
--- a/Packages/Settings/ITC1600_0_rig.json
+++ b/Packages/Settings/ITC1600_0_rig.json
@@ -12,5 +12,21 @@
       },
       "1": null
     }
+  },
+  "ITC18USB_Dev_0": {
+      "Generic ControlGroup": {
+          "Check_DataAcqHS": {
+              "Values": [
+                  true,
+                  false,
+                  false,
+                  false,
+                  false,
+                  false,
+                  false,
+                  false
+              ]
+          }
+}
   }
 }
```

You need to adapt to ITC1600 obviously.

PR #2638, opened 2026-02-25
Various installer bugfixes

t-b:
Close #2635
Close #2611

PR #2640, opened 2026-02-27
Reenable IP10 testing

t-b:
- [x] Needs #2638 
- [ ] May need to bump IP10 version due to crash fix, see https://aiephys.slack.com/archives/C06P3MYEV6H/p1772638815299929?thread_ts=1770942009.219479&cid=C06P3MYEV6H.


t-b:
The above solution would double the amount of jobs as the non-executed test is still run but finishes early. I'd suggest to just change the versions manually if we want to adapt them.

timjarsky:
@t-b yes, randomization is nice, but we don't want to repeat testing.

PR #2652, opened 2026-03-05
Reset data acquisition start button when aborting DAQ during indexing/RA

### LINKED ISSUE #2651: Aborting a sweep in indexing should reset the start acquisition button
- Index over three stimsets
- The last one has on out of range ITCDatawave value
- The acquisition is stopped but the button still shows "Stop acquisition"

t-b:
While at it I also fixed the DAQ stop reason when aborting RA.

Close #2651

### DIFF
--- a/Packages/MIES/MIES_DAEphys.ipf
+++ b/Packages/MIES/MIES_DAEphys.ipf
@@ -3615,7 +3615,7 @@ End
 ///
 /// @param device device
 /// @param mode       One of @ref ToggleAcquisitionButtonConstants
-Function DAP_ToggleAcquisitionButton(string device, variable mode)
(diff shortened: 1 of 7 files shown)


timjarsky:
@t-b, what's the best way to test this? I tried sticking a debugger call into the code, but I don't think that's doing what I want.

t-b:
> @t-b, what's the best way to test this? I tried sticking a debugger call into the code, but I don't think that's doing what I want.

My test in https://github.com/AllenInstitute/MIES/pull/2652/changes#diff-fd327cb0cf36d247bea88eb498fe689cf90b012cf5cbad46f995a2619a55c93dR1155 has an analysis function which aborts in POST_DAQ_EVENT.

timjarsky (approved):

PR #2655, opened 2026-03-10
Fix slow cache

### LINKED ISSUE #2173: Solution for fast index lookup from string
We could use a faster solution for a generic index lookups based on strings.

We use that on key locations in MIES such as the Wave Cache and the Labnotebook.

Could be wrapped in an XOP with a suitable API, shared R/W mutex.

If the keys are stored sorted then the lookup can use binary search that uses only O(log N).

### LINKED ISSUE #2653: MIES get's really slow with > 100 sweeps on rig5
FTP: issue-2653

Slack: https://aiephys.slack.com/archives/C06P3MYEV6H/p1772748939321309?thread_ts=1772744001.060339&cid=C06P3MYEV6H

t-b:
- [x] Polish move to hashmap in cache
- [x] Add upgrade path
- [x] Fix tests
- [x] Only iterate over active channels in EP_CopyLBNEpochsToEpochsWave
- [x] BackupCacheWaves: Throw away entries with no hits before backing it up
- [x] Test on rig: https://aiephys.slack.com/archives/C07LLQEDKU0/p1773358413701659?thread_ts=1773355511.893559&cid=C07LLQEDKU0
- [x] Review from Max

Close #2653 
Close #2173

### DIFF
--- a/Packages/MIES/MIES_Hashmap.ipf
+++ b/Packages/MIES/MIES_Hashmap.ipf
@@ -735,54 +735,37 @@ End
 /// Complexity: O(n)
 threadsafe Function/WAVE HM_GetAllKeys(WAVE/WAVE hashmap)
 
-	variable idx, entriesWithHash, wvIndex
-
-	WAVE/Z results = HM_GetFilledEntries(hashmap)
-
-	if(!WaveExists(results))
-		return $""
-	endif
-
-	Make/FREE/WAVE/N=(DimSize(results, ROWS)) allKeysTmp
-
-	WAVE usedRows = HM_FetchUsedRows(hashmap)
-
-	for(idx : results)
-		WAVE/T keys = HM_FetchKeys(hashmap, idx)
-
-		entriesWithHash = usedRows[idx]
-
-		Duplicate/FREE/RMD=[0, entriesWithHash - 1]/T keys, filledKeys
-		allKeysTmp[wvIndex] = filledKeys
-		wvIndex++
-	endfor
-
-	Concatenate/NP=(ROWS)/FREE/T {allKeysTmp}, allKeys
-
-	return allKeys
-End
-
-/// @brief Return the indizes into hashmap with one or more key/value pairs
-///
-/// Complexity: O(n)
-threadsafe static Function/WAVE HM_GetFilledEntries(WAVE/WAVE hashmap)
-
 	variable numThreads, numPossibleEntries
 
 	numPossibleEntries = HM_GetSize(hashmap)
 
 	numThreads = GetNumberOfUsefulThreads({numPossibleEntries})
 
+	Make/FREE/WAVE/N=(numPossibleEntries) allKeysWR
+
 	WAVE usedRows = HM_FetchUsedRows(hashmap)
+	Multithread/NT=(numThreads) allKeysWR[] = HM_GetAllKeysPerRow(hashmap, p, usedRows[p])
 
-	// not using GetTemporaryWave here to avoid issues with recursion
-	Make/FREE/N=(numPossibleEntries) matches
+	Concatenate/NP=(ROWS)/FREE/T {allKeysWR}, allKeys
 
-	Multithread/NT=(numThreads) matches = usedRows[p] ? p : NaN
+	if(DimSize(allKeys, ROWS) == 0)
+		return $""
+	endif
 
-	WAVE/Z result = ZapNaNs(matches)
+	return allKeys
+End
 
-	return result
+threadsafe static Function/WAVE HM_GetAllKeysPerRow(WAVE/WAVE hashmap, variable index, variable entriesWithHash)
+
+	if(entriesWithHash == 0)
+		return NewFreeWave(IGOR_TYPE_TEXT_WREF_DFR, 0)
+	endif
+
+	WAVE/T keys = HM_FetchKeys(hashmap, index)
+
+	Duplicate/FREE/RMD=[0, entriesWithHash - 1]/T keys, filledKeys
+
+	return filledKeys
 End
 
 /// @brief Calculate the load factor from the hashmap and the filled entries
--- a/Packages/tests/Basic/UTF_Hashmap.ipf
+++ b/Packages/tests/Basic/UTF_Hashmap.ipf
@@ -94,6 +94,29 @@ static Function CreateHashmapWorks()
 	endtry
 End
 
+/// @brief Return the indizes into hashmap with one or more key/value pairs
+///
+/// Complexity: O(n)
+static Function/WAVE GetFilledEntries(WAVE/WAVE hashmap)
+
+	variable numThreads, numPossibleEntries
+
+	numPossibleEntries = MIES_HM#HM_GetSize(hashmap)
+
+	numThreads = GetNumberOfUsefulThreads({numPossibleEntries})
+
+	WAVE usedRows = MIES_HM#HM_FetchUsedRows(hashmap)
+
+	// not using GetTemporaryWave here to avoid issues with recursion
+	Make/FREE/N=(numPossibleEntries) matches
+
+	Multithread/NT=(numThreads) matches = usedRows[p] ? p : NaN
+
+	WAVE/Z result = ZapNaNs(matches)
+
+	return result
+End
+
 static Function CheckHashmapEntry(WAVE/WAVE hashmap, variable index, variable usedEntries, WAVE/T keys, WAVE/T values)
 
 	[WAVE usedRows, WAVE/T keysRef, WAVE valuesRef] = MIES_HM#HM_FetchWaves(hashmap, index)
@@ -117,7 +140,7 @@ static Function AddEntryWorks()
 
 	WAVE hashmap = HM_Create()
 
-	WAVE/Z result = MIES_HM#HM_GetFilledEntries(hashmap)
+	WAVE/Z result = GetFilledEntries(hashmap)
 	CHECK_WAVE(result, NULL_WAVE)
 
 	// adding first value
@@ -126,7 +149,7 @@ static Function AddEntryWorks()
 	ret   = HM_AddEntry(hashmap, key, str = value)
 	CHECK_EQUAL_VAR(ret, 1)
 
-	WAVE/Z result = MIES_HM#HM_GetFilledEntries(hashmap)
+	WAVE/Z result = GetFilledEntries(hashmap)
 	CHECK_WAVE(result, NUMERIC_WAVE)
 	CHECK_EQUAL_WAVES(result, {6100}, mode = WAVE_DATA)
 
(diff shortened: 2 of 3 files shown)


t-b:
@Garados007 Could I get a review from you for the hashmap implementation and tests? Thanks.

Packages/MIES/MIES_Hashmap.ipf
```diff
@@ -0,0 +1,629 @@
...
+	WAVE/Z filledEntries = HM_GetFilledEntries(hashmap)
+
+	loadFactor = HM_CalculateLoadFactor(hashmap, filledEntries)
+
+	if(loadFactor < HM_MAX_LOAD_FACTOR)
+		// all good
+		return 0
+	endif
+
+	newSize = 2^FindNextPower(DimSize(hashmap, ROWS), 2)
```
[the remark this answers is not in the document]
t-b: You are totally right.

Packages/tests/Basic/UTF_Hashmap.ipf
```diff
@@ -0,0 +1,447 @@
...
+	CHECK_EQUAL_VAR(WaveRef(hashmap, row = index, col = 0)[0], usedEntries)
+
+	// and it is the first one in keys
+	WAVE keysRef = WaveRef(hashmap, row = index, col = 1)
+	INFO("keys: @%s", s = keysRef)
+	CHECK_EQUAL_TEXTWAVES(keysRef, keys, mode = WAVE_DATA)
+
+	// and values
+	WAVE valuesRef = WaveRef(hashmap, row = index, col = 2)
+	INFO("keys: @%s", s = keysRef)
```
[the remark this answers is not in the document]
Garados007: I see it the same way as Copilot. Is it intended to log `keysRef` twice? One time for `keysRef` and one time for `valuesRef`?

Packages/MIES/MIES_Hashmap.ipf
```diff
@@ -0,0 +1,681 @@
...
+	Multithread/NT=(numThreads) hashmap[][1] = NewFreeWave(IGOR_TYPE_TEXT_WREF_DFR, 2)
+	Multithread/NT=(numThreads) hashmap[][2] = NewFreeWave(IGOR_TYPE_TEXT_WREF_DFR, 2)
+
+	return hashmap
+End
+
+/// @brief Add an entry into the hashmap
+///
+/// Complexity: Amortized O(1)
+threadsafe Function HM_AddEntry(WAVE/WAVE hashmap, string key, string value)
```
Garados007: Might be useful to return a status value, like 1 if the entry was added without any problem and 0 if a previous entry was overwritten. This can be useful if the consumer wants to add/update an entry and then perform an action depending on if the entry was already existing. With the status return value, there is no need to check if the key was existing before the add operation.

Packages/MIES/MIES_Hashmap.ipf
```diff
@@ -0,0 +1,681 @@
...
+	values[entriesWithHash - 1] = ""
+
+	usedRows[0] = entriesWithHash - 1
+
+	return 0
+End
+
+/// @brief Return all keys in the hashmap
+///
+/// Complexity: O(n)
```
Garados007: This is wrong and is `O(n²)`. See my previous comment as an answer to copilot's review.

Packages/MIES/MIES_Hashmap.ipf
```diff
@@ -0,0 +1,681 @@
...
+
+	if(loadFactor < HM_MAX_LOAD_FACTOR)
+		// all good
+		return 0
+	endif
+
+	newSize = 2 * DimSize(hashmap, ROWS)
+	WAVE/WAVE hashmapLarger = HM_Create(size = newSize)
+
+	for(srcIdx : filledEntries)
```
Garados007: At this point there is no big benefit in using `filledEntries`. You can assume that most entries are filled anyway if you need to rehash. You can just iterate the whole list and skip empty entries. If skipping empty entries takes to long, you can always tweak `HM_MAX_LOAD_FACTOR` to perform a rehash less often and less skips when performing a rehash.

Packages/MIES/MIES_Utilities_Numeric.ipf
```diff
@@ -116,7 +116,7 @@ threadsafe Function FindNextPower(variable a, variable p)
 		return 1
 	endif
 
-	return ceil(log(a) / log(p))
+	return ceil(log(a) / log(p) + 1e-15)
```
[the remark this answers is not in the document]
t-b: Good point. I also fixed the documentation.

Packages/MIES/MIES_Cache.ipf
```diff
@@ -821,10 +822,61 @@ Function CA_OutputCacheStatistics()
...
+	for(i = 0; i < numEntries; i += 1)
+		key   = allKeys[i]
+		index = CA_GetCacheIndex(keys, key)
+
+		if(stats_old[index][%Hits] == 0)
+			// remove entries with no hits
+			CA_DeleteCacheEntryImpl(keys, key, index, values_old, stats_old)
+		else
+			EnsureLargeEnoughWave(values, dimension = ROWS, indexShouldExist = newIndex)
+			EnsureLargeEnoughWave(stats, dimension = ROWS, indexShouldExist = newIndex, initialValue = NaN)
```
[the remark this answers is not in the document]
t-b: > I see it the same way as Copilot. Is it intended to log keysRef twice? One time for keysRef and one time for valuesRef?

Fixed.

Packages/MIES/MIES_Hashmap.ipf
t-b: From my understanding the fix you posted makes it `O(n)` again.

Garados007: Yes.

t-b:
@Garados007 

Thanks for the review. I've added inline replies and here is the rest.

> Couldn't you just store ints in your hashmap and remove any type conversion? Storing ints as strings waste
> (conversion) time and memory. Also becomes your implementation slower with larger int values (because of the
> larger strings that need to be copied).

For the cache I want to store int's as value. But the more expected cases will be storing strings. I could
make it configurable what the type of the values wave is, but that is a bit too much effort for now IMHO.

> You specify the time complexity of many operations in MIES_Hashmap, which I think is good, but you didn't
> specify the space/memory complexity, which can become very large in certain cases (up to a point that you
> create almost a full copy of the hashmap).

I don't think memory complexity matter as much as time complexity.

> I put special care in functions that are computed a lot and therefore needs to be fast. You can see this in my
> comments.

Sure!

> I didn't check your benchmark code for now. I will do this after I posted my review.

Jeep.

t-b:
git range-diff 5a63a5fc7c54bd15c5d5557b81c25e4cdab037ab...d4f6ec372fe61d1a284db7a55d8cc16490403ca9

Packages/MIES/MIES_Hashmap.ipf
t-b: I have a  local version which stores the total number of used entries so that the load factor calculation is O(1).

t-b:
@Garados007 Here are the enhancements I mentioned earlier. Please have a look.

$ git range-diff d4f6ec372fe61d1a284db7a55d8cc16490403ca9...99e1bdb1dbb33e8a2616e32b7da0cd15599fa175
```
 1:  f9eb0245bb !  1:  3fc555ad0b CA_Compactify: Add it and use it in BackupCacheWaves
    @@ Packages/MIES/MIES_Cache.ipf: Function CA_OutputCacheStatistics()
     +       return NaN
     +   endif
     +
    -+   WAVE/WAVE values_old = MakeWaveFree(GetCacheValueWave())
     +   WAVE      stats_old  = MakeWaveFree(GetCacheStatsWave())
    ++   WAVE/WAVE values_old = MakeWaveFree(GetCacheValueWave())
     +
    -+   WAVE/WAVE values = GetCacheValueWave()
     +   WAVE      stats  = GetCacheStatsWave()
    ++   WAVE/WAVE values = GetCacheValueWave()
     +
     +   numEntries = DimSize(allKeys, ROWS)
     +   EnsureLargeEnoughWave(values, dimension = ROWS, indexShouldExist = numEntries - 1)
```
 2:  be4414b610 =  2:  7124146409 GetNumberOfUsefulThreads: Make it usable from IgorStartOrNewHook()
 3:  0c2500b151 =  3:  c3dbd1716a FindNextPower/FindPreviousPower: Fix documentation
 4:  ac0bab316c =  4:  4218b2c956 HM_DJBHash: Make it faster
 5:  e9a5b3e5ed =  5:  25456962b9 treewide: Unify accessing temporary waves for indexing help
 6:  d4f6ec372f =  6:  1b6337da46 DecimateWithMethod: Postpone junkWave fetching
 -:  ---------- >  7:  03f89e4246 MIES_Hashmap.ipf: Revise storage layout
 -:  ---------- >  8:  4b83bee2c6 HM_GetKeyIndex: Add fastpath if we don't have any entries
 -:  ---------- >  9:  ddafe390ba HM_DJBHash: Mention that embedded nulls are supported
 -:  ---------- > 10:  6c4284ce2c UTF_DataGenerators.ipf: Factor GenerateAllPossibleNumericWaveTypes() out
 -:  ---------- > 11:  1c085fbbce HM_DeleteEntry: Only rotates keys if not the last one
 -:  ---------- > 12:  af3337487d MIES_Hashmap.ipf: Add support for non-complex numeric values
 -:  ---------- > 13:  bea2b20409 MIES_Cache.ipf: Switch the cache to numeric value waves
 -:  ---------- > 14:  99e1bdb1db HM_GetAllKeys: Reimplement without HM_GetFilledEntries

I've also created https://github.com/AllenInstitute/MIES/issues/2666 for the other cases in MIES where we could use the hashmap.

Packages/MIES/MIES_Hashmap.ipf
```diff
@@ -558,70 +735,47 @@ End
...
-	WAVE/Z result = ZapNaNs(matches)
+	Duplicate/FREE/RMD=[0, entriesWithHash - 1]/T keys, filledKeys
 
-	return result
+	return filledKeys
 End
 
 /// @brief Calculate the load factor from the hashmap and the filled entries
 ///
 /// Complexity: O(n)
```
Garados007: The complexity is no longer `O(n)` but `O(1)`.

Packages/MIES/MIES_Hashmap.ipf
```diff
@@ -0,0 +1,839 @@
...
+	WAVE totalEntries = HM_FetchStats(hashmap)
+
+	return totalEntries[HM_TOTAL_ENTRIES_ROW] / HM_GetSize(hashmap)
+End
+
+/// @brief Rehashes if required and returns a modified hashmap pass-by-reference
+///
+/// The load factor (number of available entries vs filled entries) is determined.
+/// And if that is above #HM_MAX_LOAD_FACTOR we create a new hashmap with the doubled size and
+/// readd all existing entries.
```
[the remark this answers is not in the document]
Garados007: Oh, I misunderstood you sentence. 😂

t-b:
LGTM: https://aiephys.slack.com/archives/C07LLQEDKU0/p1774485178665349?thread_ts=1774484930.034029&cid=C07LLQEDKU0

PR #2671, opened 2026-03-18
CHI_CheckInstallation: Check that we actually have installed with shortcuts on windows

t-b:
A common error when manually installing is that the files are copied and not linked. So let's check that as well.


timjarsky:
@t-b, do we want this to check for all shortcuts or MIES shortcuts? It's currently checking for any/all shortcuts.

t-b:
> @t-b, do we want this to check for all shortcuts or MIES shortcuts? It's currently checking for any/all shortcuts.

My thinking was that checking for any shortcuts is easier to implement.

t-b:
@timjarsky Should I change it? Or keep as is?

Issue #2695, opened 2026-04-27
Opening files with large hashmaps is really slow


t-b:
Opening experiments with large caches (> 100k entries) is really slow on my machine. I suspect this is due to the large number of free text waves the hashmap creates. It has 2 free text waves per entry, one for the list of keys and one for the values.

We could avoid that by creating one global wave for keys and values, disadvantage being that we then waste saome memory if a few hash entries have a lot of duplicates.

Required changes:
- HM_CreateHashMap
- Explicit upgrade path via HM_UpgradeHashmap
- Upgrade hashmaps in GetCacheKeyHashMap and GetLogbookKeyHashmap, add wave versioning

@timjarsky Did you also notice that?

t-b:
Using a 2D wave has also the advantage that this takes less memory.

```igorpro
•make/n=2/t txt
•print waveinfo(txt, 0)
  NUMTYPE:0;DUNITS:;XUNITS:;MODIFIED:1;FULLSCALE:0,0,0;MODTIME:3860315657;PATH:;LOCK:0;HGETSTATE:0;SIZEINBYTES:664;
•make/n=0/t/O txt
•print waveinfo(txt, 0)
  NUMTYPE:0;DUNITS:;XUNITS:;MODIFIED:1;FULLSCALE:0,0,0;MODTIME:3860315683;PATH:;LOCK:0;HGETSTATE:0;SIZEINBYTES:648;
•make/n=(2, 150e3)/t/O txt
•print waveinfo(txt, 0)
  NUMTYPE:0;DUNITS:;XUNITS:;MODIFIED:1;FULLSCALE:0,0,0;MODTIME:3860315727;PATH:;LOCK:0;HGETSTATE:0;SIZEINBYTES:2400648;
```

PR #2696, opened 2026-04-27
More agressive cache compactification

t-b:
Related to #2695 

Will merge once CI passes.


t-b:
@timjarsky I don't think this really needs testing from your side. So if you are okay, I would just merge.

PR #2704, opened 2026-05-08
Bugfix: Change amp storage row label for button_DataAcq_WCAuto to empty string

### LINKED ISSUE #2702: Duplicate GUI control association to row label "WholeCellCap" in AmpStorageWave for VC
After a Copilot comment in #2559 about duplicated DimLabels in a wave created for FFI I digged a bit deeper and noticed the following:

In AI_AmpStorageControlToRowLabel each GUI control is mapped to a row label for the AmpStorageWave and the row labels are sorted by clamp mode.

There `button_DataAcq_WCAuto` -> `WholeCellCap` is in the block for IC whereas it actually is a VC control.

I checked this through: `AI_GetFunctionConstantForClampMode` and then mapping the function constant to the controls through `AI_MapFunctionConstantToControl`.

The issue is now that for VC also `setvar_DataAcq_WCC` maps to `WholeCellCap`, thus to the same row label.

I would propose to rename the row label for `button_DataAcq_WCAuto` to something else unique.
Then create a new version of the AmpStorageWave in GetAmplifierParamStorageWave with the new row.

I currently do not overview if there has been some fallout through the duplication, where e.g. the wrong value from the button (usually 0) was written to the AmpStorageWave and possibly written to the LBN?

MichaelHuth:
The button can not return a usable value, but was causing a duplicate row label before.

close #2702

If this fix is working, I could think of a possible test like:

```
	WAVE aiFuncs = AI_GetFunctionConstantForClampMode(clampMode)
	numFuncs = DimSize(aiFuncs, ROWS)
	Make/FREE/T/N=(numFuncs) ctrlNames, ampLabels

	ctrlNames[] = AI_MapFunctionConstantToControl(aiFuncs[p], clampMode)
	ampLabels[] = AI_AmpStorageControlToRowLabel(ctrlNames[p])
        // Do that for all clamp modes and concatenate the retrieved labels
	// Then check if ampLabels have no duplicates as well as check if each label exist as dimlabels in the amp storage wave
```

### DIFF
--- a/Packages/MIES/MIES_AmplifierInteraction.ipf
+++ b/Packages/MIES/MIES_AmplifierInteraction.ipf
@@ -1246,6 +1246,18 @@ static Function/S AI_AmpStorageControlToRowLabel(string ctrl)
 		case "setvar_DataAcq_PipetteOffset_VC":
 			return "PipetteOffsetVC"
 			break
+		case "check_DataAcq_Amp_Chain":
+			return "RSCompChaining"
+			break
+		case "button_DataAcq_WCAuto":
+			return "WholeCellCap"
+			break
+		case "button_DataAcq_FastComp_VC": // fallthrough
+		case "button_DataAcq_SlowComp_VC": // fallthrough
+		case "button_DataAcq_AutoPipOffset_VC":
+			// no row exists
+			return ""
+			break
 		// I-Clamp controls
 		case "setvar_DataAcq_Hold_IC":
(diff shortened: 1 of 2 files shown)


t-b:
@MichaelHuth LGTM. A test would be good.

t-b:
@MichaelHuth I've added a test.

t-b (approved):

PR #2709, opened 2026-05-12
Switch internal hashmap layout to be more efficient

### LINKED ISSUE #2695: Opening files with large hashmaps is really slow

t-b:
- [x] Add upgrade paths for cache keys and labnotebook keys
- [x] Look into coverage for MIES_Hashmap.ipf
- [x] Polish commits

Close #2695

### DIFF
--- a/Packages/MIES/MIES_Hashmap.ipf
+++ b/Packages/MIES/MIES_Hashmap.ipf
@@ -455,21 +455,21 @@ threadsafe static Function HM_GetSize(WAVE/WAVE hashmap)
 	return DimSize(hashmap[HM_HASHMAP_ROW], ROWS)
 End
 
-threadsafe static Function/WAVE HM_FetchKeys(WAVE/WAVE hashmap, variable idx)
+threadsafe static Function/WAVE HM_FetchKeys(WAVE/WAVE hashmap, variable bucketIndex)
 
-	return WaveRef(hashmap[HM_HASHMAP_ROW], row = idx, col = HM_KEYS_COLUMN)
+	return WaveRef(hashmap[HM_HASHMAP_ROW], row = bucketIndex, col = HM_KEYS_COLUMN)
 End
 
-threadsafe static Function/WAVE HM_FetchValues(WAVE/WAVE hashmap, variable idx)
+threadsafe static Function/WAVE HM_FetchValues(WAVE/WAVE hashmap, variable bucketIndex)
 
-	return WaveRef(hashmap[HM_HASHMAP_ROW], row = idx, col = HM_VALUES_COLUMN)
+	return WaveRef(hashmap[HM_HASHMAP_ROW], row = bucketIndex, col = HM_VALUES_COLUMN)
 End
 
-threadsafe static Function [WAVE usedRows, WAVE/T keys, WAVE values] HM_FetchWaves(WAVE/WAVE hashmap, variable idx)
+threadsafe static Function [WAVE usedRows, WAVE/T keys, WAVE values] HM_FetchWaves(WAVE/WAVE hashmap, variable bucketIndex)
 
 	WAVE   usedRows = HM_FetchUsedRows(hashmap)
-	WAVE/T keys     = HM_FetchKeys(hashmap, idx)
-	WAVE   values   = HM_FetchValues(hashmap, idx)
+	WAVE/T keys     = HM_FetchKeys(hashmap, bucketIndex)
+	WAVE   values   = HM_FetchValues(hashmap, bucketIndex)
 
 	return [usedRows, keys, values]
 End
@@ -579,12 +579,12 @@ threadsafe Function HM_Clear(WAVE/WAVE hashmap)
 	totalEntries[HM_TOTAL_ENTRIES_ROW] = 0
 End
 
-threadsafe static Function HM_ClearKeysAndValues(WAVE/WAVE hashmap, variable idx)
+threadsafe static Function HM_ClearKeysAndValues(WAVE/WAVE hashmap, variable bucketIndex)
 
-	WAVE/T keys = HM_FetchKeys(hashmap, idx)
+	WAVE/T keys = HM_FetchKeys(hashmap, bucketIndex)
 	keys[] = ""
 
-	WAVE values = HM_FetchValues(hashmap, idx)
+	WAVE values = HM_FetchValues(hashmap, bucketIndex)
 
 	if(IsTextWave(values))
 		WAVE/T valuesText = values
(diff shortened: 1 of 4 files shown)


Packages/MIES/MIES_Hashmap.ipf
```diff
@@ -528,12 +599,13 @@ threadsafe static Function/WAVE HM_CreateHashmap(variable size, variable valueTy
...
-	numThreads = GetNumberOfUsefulThreads({size})
+	Make/FREE/WAVE/N=(2) hashmap_impl
 
-	Make/FREE/WAVE/N=(size, 2) hashmap_impl
+	Make/FREE/N=(size, 2)/T keys
+	hashmap_impl[HM_KEYS_COLUMN] = keys
 
-	Multithread/NT=(numThreads) hashmap_impl[][HM_KEYS_COLUMN] = NewFreeWave(IGOR_TYPE_TEXT_WREF_DFR, 2)
-	Multithread/NT=(numThreads) hashmap_impl[][HM_VALUES_COLUMN] = NewFreeWave(valueType, 2)
+	Make/FREE/Y=(valueType)/N=(size, 2) values
```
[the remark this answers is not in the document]
t-b: Wrong, see https://docs.wavemetrics.com/igorpro/commands/make#wave-data-types. And also the tests pass.

Packages/MIES/MIES_Hashmap.ipf
```diff
@@ -452,26 +448,102 @@ End
...
+
+	for(i = 0; i < size; i += 1)
+		numEntries = usedCols[i]
+
+		if(numEntries == 0)
+			continue
+		endif
+
+		if(numEntries > DimSize(keys, COLS))
+			newSize = 2^FindNextPower(numEntries, HM_INITIAL_KEYS_VALUES_COL_SIZE)
```
MichaelHuth, line 523: I think the max newsize can be determined before the loop to do only a single redimension.

Packages/MIES/MIES_Hashmap.ipf
```diff
@@ -711,15 +774,14 @@ End
...
+	[WAVE usedCols, WAVE/T keys, WAVE values] = HM_FetchWaves(hashmap)
 
-	entriesWithHash = usedRows[idx]
+	bucketIndex     = HM_HashKey(hashmap, key)
+	entriesWithHash = usedCols[bucketIndex]
 
-	keyIndex = HM_GetKeyIndex(keys, key, entriesWithHash)
+	keyIndex = HM_GetKeyIndex(keys, bucketIndex, key, entriesWithHash)
 
 	if(!(keyIndex >= 0))
```
MichaelHuth: Why not use `IsNaN(keyIndex)` here?

Packages/MIES/MIES_Hashmap.ipf
```diff
@@ -728,29 +790,29 @@ threadsafe Function HM_DeleteEntry(WAVE/WAVE hashmap, string key)
...
-		// move all keys and values in the range [keyIndex + 1, entriesWithHash - 1]
+		// move all keys and values in the column range [keyIndex + 1, entriesWithHash - 1]
 		// one element to the left
-		keys[keyIndex, entriesWithHash - 2] = keys[p + 1]
+		keys[bucketIndex][keyIndex, entriesWithHash - 2] = keys[bucketIndex][q + 1]
 
 		if(isStr)
 			WAVE/T valuesText = values
-			valuesText[keyIndex, entriesWithHash - 2] = valuesText[p + 1]
+			valuesText[bucketIndex][keyIndex, entriesWithHash - 2] = valuesText[bucketIndex][q + 1]
```
MichaelHuth, line 799: Maybe at some day the built-in `Rotate` operation is capable of doing that wave type independent.

Packages/MIES/MIES_Hashmap.ipf
MichaelHuth, line 895: The loadFactor variable could be optimized out here.

MichaelHuth, line 946: iff -> if

Also for the numEntries argument add the hint that it is helpful when the entries input wave is a NOTE_INDEX managed wave.

Packages/MIES/MIES_Hashmap.ipf
```diff
@@ -514,26 +586,23 @@ End
...
-/// Columns:
-/// - 0: 1D text wave with all keys for this hash
-/// - 1: 1D wave with all values for this hash, type depends on valueType parameter
+/// - 0: 2D text wave with all keys for this hash, rows: buckets, cols: numbers of key/value pairs
+/// - 1: 2D wave with all values for this hash, type depends on valueType parameter, rows: buckets, cols: number of key/value
 ///
 /// Complexity: O(n)
 ///
 /// @param size size of the hashmap, needs to be a power of two
 /// @param valueType wave type of the values, defaults to text wave and can be one of @ref IgorTypes
```
MichaelHuth: Is only partially correct.
valueType allows Text waves and numeric wave types, but not DFREF and WAVEREF wave types, while WaveType with mode 0 also returns 0 for DFREF and WAVEREF types. Thus, it is not exactly the same as WaveType would return. This can already be seen by the default, which is set to IGOR_TYPE_TEXT_WREF_DFR, but applies only as Text wave type with /Y=(valueType).
So referencing the IGOR_TYPE_ return constants for WaveType for the /Y=(valueType) flag is ok, but needs a mention that WAVEREF and DFREF type is not supported.

On a side note: NewFreeWave introduces the special types 0x100 and 0x200 for DF/WR, which Make does not support, which makes NewFreeWave the only? function to create a wave of any type through a type code.

MichaelHuth (dismissed):
I think it is good, I have seen some minor things.

If you think none of these are really relevant, feel free to merge.

Packages/MIES/MIES_Hashmap.ipf
t-b: Yes this is feasible, but it is very rare to require the keys/values to redimension, and much rarer to do it multiple times. So I'd lean to keep it as is.

Packages/MIES/MIES_Hashmap.ipf
t-b: I think my original idea was speed, but this is not a good argument here. Fixed.

Packages/MIES/MIES_Hashmap.ipf
t-b: Yes let's hope for IP11 to have that. It is quite sad to have to do that manually.

Packages/MIES/MIES_Hashmap.ipf
t-b: I think it is more readable without having a function call in the if condition.

t-b: > iff -> if

This means if-and-only-if, short iff, see https://en.wikipedia.org/wiki/If_and_only_if.

> Also for the numEntries argument add the hint that it is helpful when the entries input wave is a NOTE_INDEX managed wave.

Good point. Done.

Packages/MIES/MIES_Hashmap.ipf
t-b: Good point. Done.

t-b:
> I think it is good, I have seen some minor things.
> 
> If you think none of these are really relevant, feel free to merge.

THanks for the review. I've added a couple of minor fixes based on your review. Will merge once CI passed.

PR #2718, opened 2026-05-29
Enhance Wave Reference Wave Browser

### LINKED ISSUE #2672: Enhance wave ref browser
- [x] Make it work with 2D wave ref waves
- [x] Feedback from external testing (AL):

> ... if you enable Table->Show Column Info Tags, you can then hover over a point in a WAVE ref wave and the tooltip will show you information about the referenced wave. 
> Michael could use DisplayHelpTopic "TablePopup Menu" to add a contextual menu item that does something similar to the BrowseWref button in his package. He could also use DisplayHelpTopic "DataBrowserObjectsPopup Menu" to add such a menu item to the DB's wave list. IMO that's better than using a button in the DB.
> The table tooltip does work from within the debugger as well (if you've enabled them outside of the debugger).

MichaelHuth:
- add context menu on wave reference waves in data browser
- add context menu for columns in tables to open WRB
- add preview of the first 10 points in the sub wave
- add attached subwindow that shows the current waveref wave in a table

The table allows to navigate on multidimensional waves and to browse further. Due to Igor Pro limitations the workflow is like:

Press enter to open the last wave reference wave where a tooltip was shown.

close #2672

### DIFF
--- a/Packages/MIES/MIES_WaveDataFolderGetters.ipf
+++ b/Packages/MIES/MIES_WaveDataFolderGetters.ipf
@@ -9253,6 +9253,33 @@ Function/WAVE GetWaverefBRowserReferenceWave()
 		Make/WAVE/N=(0) dfr:$name/WAVE=wv
 	endif
 
+	SetWaveVersion(wv, version)
+
+	return wv
+End
+
+Function/WAVE GetWaverefBrowserLastTooltipLocation()
+
+	variable version = 1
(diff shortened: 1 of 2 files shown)


MichaelHuth:
@copilot Review this PR

t-b:
Works here. I have only skimmed over the code.

t-b (approved):

Issue #2724, opened 2026-06-03
FFI_GetWithOptionToSetPressure returns before pressure has updated


campagnola:
I am finding that when I call FFI_GetWithOptionToSetPressure, it does set the pressure correctly, but the return value is the old pressure value rather than the new one. I assume it is simply returning before it has completed updating the pressure?

campagnola:
MIES version 2.9_20250502-890-ge4dbeba90

t-b:
@MichaelHuth Can you look into this? Thanks.

MichaelHuth:
What I see is the following:

I think pressure is only a property that is set through a analog output, but never read back from some hardware sensor.

When used within limits I could not see that the returned value is different from the set value.

When the set value is outside the limits the current behavior is:


In mode "user" and mode "regulator" the limits are > -9.9 psi and < 9.9 psi. If the set value exceed the limits then an ASSERTion is thrown.


In mode "athmosphere" and "default" (or anything else) the limits are >= -10 psi and <= 10 psi. (Limited by the pressure input GUI control) When the set value exceeds the limits then the set value is capped at the limit, no ASSERTion is thrown. **This is a case where the set value can be unequal to the readback value**.


@campagnola Does that fit to the case where you see the discrepancy?


My hacked together test code:

```igorpro
Function test()

	string source
	string device = "IPA_E_100170" // needs this device as locked device
	variable result, value
	[source, result] = ReadPressureSourceAndPressure(device, 0)
	print/D "Read Old:", source, result
//	source = "athmosphere"
	source = "user"
//	source = "regulator"
//	source = "default_something"
//	value = abs(enoise(3))
	value = 100
	print/D "Set:", source, value
	[source, result] = FFI_GetWithOptionToSetPressure(device, 0, source, value)
	print/D "Readback:", source, result
End
```

campagnola:
No, in my case the function always returns the previous pressure value rather than the new one. I'd suggest testing via zmq or acq4, since that's how we are calling it. Minimally:
```
from acq4.util.mies import MIES
m = MIES.getBridge()
m.setPressureAndSource(source='regulator', pressure=1)
```
Code for that function here: https://github.com/acq4/acq4/blob/main/acq4/util/mies.py#L112

t-b:
I presume this was closed as an error.

MichaelHuth:
Ok I got acq4 working (the install batch did not work on my machine).

Here is the code I ran:

```python
[...]

PR #2726, opened 2026-06-05
Fixes for Pressure Control Popup and psi limits error message

MichaelHuth:
Two things I found while looking into #2724

(this is not a fix for the issue)

### DIFF
--- a/Packages/MIES/MIES_Configuration.ipf
+++ b/Packages/MIES/MIES_Configuration.ipf
@@ -592,6 +592,7 @@ End
 Function CONF_PrimeDeviceLists(string device)
 
 	variable hardwareType
+	string emptyDevList = DAP_GetEmptyDeviceList()
 
 	SVAR globalITCDeviceList = $GetITCDeviceList()
 	SVAR globalNIDeviceList  = $GetNIDeviceList()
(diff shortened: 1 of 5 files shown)


t-b (dismissed):

timjarsky:
@MichaelHuth, can you confirm that these changes won't cause the DAC's voltage range to be exceeded?

MichaelHuth:
I had to update this once more because the fix was wrong. It makes no sense that I need to add a ";" when joining parts of a list. I fixed it now in `DAP_GetITCDeviceList`.

@timjarsky These changes do not touch any limits enforced by MIES. I just improved the error message for manual pressure mode when the value exceeds the limits. There the limits are MIN_REGULATOR_PRESSURE and MAX_REGULATOR_PRESSURE. For other modes the user interface input enforces a range of -10 to 10. So depending on mode there are different limits.

MichaelHuth:
@t-b The CI errors indicated that the issue is a bit deeper. I fixed some more stuff. I reassign to myself, until it runs through again.

t-b:
Review:

318b7546ff (PressureControl UI: Fix building of popupmenu for devices, 2026-06-05)

ButtonProc_Hrdwr_P_UpdtDAClist:
```
+            if(IsEmpty(filteredList))
+                filteredList = emptyDevList
+            endif
```

I think this needs to be `if(!cmpstr(filteredList, emptyDevList))` because filteredList is not empty anymore
by default. 

4f71266154 (PressureControl: Correct error message for regulator psi limits, 2026-06-05)

Looks good.

t-b (changes requested):
See comment.

MichaelHuth:
> I think this needs to be `if(!cmpstr(filteredList, emptyDevList))` because filteredList is not empty anymore by default.

I removed the three lines that were actually dead code now. filteredList starts as emptyDevList and then valid devices get added to the list. There is no need to have any extra treatment after this loop.

t-b (approved):

PR #2738, opened 2026-07-07
Add tolerance option to GetRowIndex

t-b:


t-b:
**Review summary** — 8 finder agents (3 correctness angles, 3 cleanup angles, altitude, conventions) surfaced candidates across the `main...HEAD` diff (treewide `TXOP` magic-number → named-constant rewrite + `GetRowIndex` tolerance parameter). After 1-vote verification, 5 findings survived, ranked most-severe first:

1. **`tools/check-code.sh:193`** (CONFIRMED) — the new anti-magic-number lint rule only pattern-matches `FindValue.../TXOP=`, missing the identical bitmask passed via `RemoveTextWaveEntry1D(options=...)` / `GetRowIndex(textOp=...)`. Proven by a live escape: `MIES_Cache.ipf:868` already has `textOp = 2` unconverted and unflagged.
2. **`MIES_Utilities_Algorithm.ipf:328`** (PLAUSIBLE) — `GetRowIndex`'s new `tol` param is silently ignored by the Igor Pro 9 fallback path when searching for `Inf`/`-Inf` (exact-equality loop, no tolerance). Version-dependent behavior with no test coverage; the underlying Igor-9 limitation predates this PR but the gap wasn't addressed when tol was added.
3. **`MIES_OverlaySweeps.ipf:158`** (CONFIRMED) — two `RemoveTextWaveEntry1D` calls still use raw `options = 1` instead of `TXOP_CASE_SENSE`, missed by the "treewide" conversion in the very file that got other call sites converted.
4. **`UTF_Utils_WaveHandling.ipf:3320`** (CONFIRMED) — the rewritten RTWE1D tests dropped all coverage of `TXOP_WHOLE_ELEM` combined with another bit; that exact combination is used in production (`GetDecimalMultiplierValue`) but is now untested anywhere.
5. **`UTF_Utils_WaveHandling.ipf:3323`** (PLAUSIBLE) — the rewritten test's expected results rely on an unverified assumption about Igor's undocumented "whole word" TXOP semantics; this is the only call site in the repo exercising that bit, so it's worth confirming by actually running the suite before merge.

One candidate (redundant `git grep` invocation in check-code.sh) was REFUTED — it matches the file's existing one-check-per-grep convention and is negligible next to Igor Pro compile/test time.

Packages/tests/Basic/UTF_Utils_WaveHandling.ipf
```diff
@@ -3310,24 +3310,21 @@ End
...
 
-	variable TXOP_CASE_INSENSITIVE = 2
-	variable TXOP_WHOLEWORD        = 4
-	variable TXOP_COMBO            = TXOP_CASE_INSENSITIVE | TXOP_WHOLEWORD
-
-	Make/FREE/T wv = {"Testing", "test", "TEST", "tests"}
-	CHECK_EQUAL_VAR(RemoveTextWaveEntry1D(wv, "test", options = TXOP_COMBO, all = 1), 0)
-	CHECK_EQUAL_VAR(DimSize(wv, ROWS), 2)
+	Make/FREE/T wv = {"Testing", "test", "TEST", "tests", "test"}
+	CHECK_EQUAL_VAR(RemoveTextWaveEntry1D(wv, "test", options = TXOP_CASE_SENSE | TXOP_WHOLE_WORD, all = 1), 0)
```
MichaelHuth: I think if it is decided to use TXOP_WHOLE_WORD here it should be distinguishable from using TXOP_WHOLE_ELEM instead here.
So for TXOP_WHOLE_WORD the elements should be like `{"Testing 123", "test 123", "TEST 123", "tests 123", "test 234"}`. 

on a second thought, this TC is just for checking the `all = 1` argument. So just leaving it with `TXOP_CASE_SENSE` as in the previous TC and having then `test` and `tests` removed would be more to the point for this TC.

Packages/MIES/MIES_Utilities_Algorithm.ipf
```diff
@@ -286,6 +287,12 @@ threadsafe Function GetRowIndex(WAVE wv, [variable val, string str, WAVE/Z refWa
 		textOp = TXOP_WHOLE_ELEM
 	endif
 
+	if(ParamIsDefault(tol))
+		tol = 0
+	else
+		ASSERT_TS(IsFinite(tol) && tol >= 0, "Expected finite and non-negative tolerance")
```
MichaelHuth: We have `IsNullOrPositiveAndFinite` in our Checkers functions.

Packages/tests/Basic/UTF_Utils_Algorithm.ipf
```diff
@@ -260,6 +276,16 @@ static Function TestGetRowIndex()
...
 	CHECK_EQUAL_VAR(GetRowIndex(doubleWave, str = strToSearch, reverseSearch = 1), NaN)
 
+	// but a tolerance can be set
+	Make/FREE/D doubleWave = {1}
+	tol           = 1e-12
+	valueToSearch = 1 + 1e-13
+	strToSearch   = num2str(valueToSearch, "%.15g")
+	CHECK_EQUAL_VAR(GetRowIndex(doubleWave, val = valueToSearch, tol = tol), 0)
+	CHECK_EQUAL_VAR(GetRowIndex(doubleWave, str = strToSearch, tol = tol), 0)
+	CHECK_EQUAL_VAR(GetRowIndex(doubleWave, val = valueToSearch, tol = tol, reverseSearch = 1), 0)
```
MichaelHuth: It would be good to have also here for reverseSearch the match at a different index, otherwise the call could just ignore this option and the test does not catch it.

MichaelHuth (changes requested):
See comments.

t-b:
@MichaelHuth Thanks for the review. I think I've fixed the outlined points.

MichaelHuth (approved):

PR #2759, opened 2026-08-04
Add PerformSubsystemEntry guards when a different MIES module is called

### LINKED ISSUE #2610: Check for lingering RTEs on subsystem entry
A recent issue with an lingering RTE [1] highlighted again that bugs like this are very difficult to diagnose and fix. While we wait for IP to get better in this regard (issue 5595) we could also try to make it easier to localize.

One idea would be that we add `PerformSubsystemEntry()` (which does the same as `AssertOnAndClearRTError`) calls when we enter a new subsystem:

- Logbook reading/writing
- Amplifier
- SweepFormula
- NWB
- PressureControl
- Publish
- Wavebuilder
- Databrowser/SweepBrowser
- AnalysisBrowser
- Cache
- Async Framework
- Configuration

In that way we should catch lingering RTEs earlier and also have them more localized.

[1]: https://github.com/AllenInstitute/MIES/issues/2568#issuecomment-3608720379

MichaelHuth:
This should catch lingering RTEs earlier.

close #2610

### DIFF
--- a/Packages/MIES/MIES_AmplifierInteraction.ipf
+++ b/Packages/MIES/MIES_AmplifierInteraction.ipf
@@ -122,7 +122,7 @@ End
 /// @param clampMode  clamp mode (pass `NaN` for doesn't matter)
 /// @param func       MCC function, one of @ref AI_SendToAmpConstants
 /// @param accessType One of @ref MCCAccessType
-Function AI_GetMCCScale(variable clampMode, variable func, variable accessType)
+static Function AI_GetMCCScale(variable clampMode, variable func, variable accessType)
 
 	AI_AssertOnInvalidAccessType(accessType)
 
@@ -601,7 +601,7 @@ static Function AI_UpdateAmpView(string device, variable headStage, [variable fu
 	endfor
 End
 
-Function AI_SetMIESHeadstage(string device, [variable headstage, variable increment])
+static Function AI_SetMIESHeadstage(string device, [variable headstage, variable increment])
 
 	if(ParamIsDefault(headstage) && ParamIsDefault(increment))
 		return NaN
@@ -1106,7 +1106,7 @@ threadsafe Function/S AI_MapFunctionConstantToName(variable func, variable clamp
 End
 
 /// @brief Map human readable names to functions constants from @ref AI_SendToAmpConstants
-threadsafe Function AI_MapNameToFunctionConstant(string name)
+threadsafe static Function AI_MapNameToFunctionConstant(string name)
 
 	strswitch(name)
 		// begin AmpStorageWave row labels
@@ -2184,7 +2184,7 @@ static Function [WAVE telegraphServers, WAVE ampMCC] AI_FindConnectedAmpsNoCache
 	return [telegraphServers, ampMCC]
 End
 
-Function [STRUCT AxonTelegraph_DataStruct tds] AI_GetTelegraphStruct(variable axonSerial, variable channel)
+static Function [STRUCT AxonTelegraph_DataStruct tds] AI_GetTelegraphStruct(variable axonSerial, variable channel)
 
 	variable i, err
 	string errMsg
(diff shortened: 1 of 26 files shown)


t-b:
There was an issue with CI therefore I rebased.

t-b:
Review:

eb5d6b4195 (ProgramFlow: Move MIES state string generation to own functions, 2026-08-06)

Makes sense.

0f319a1ea9 (ProgramFlow: Add optional verbose mode to AssertOnAndClearRTError, 2026-08-06)

Okay.

3a517568a8 (ProgramFlow: Add PerformSubsystemEntry as guard for calls between MIES modules, 2026-08-06)

- Typo: outptu

- Adding functions without callers is suboptimal as it leaves the reader to guess how they are used.
- The two functions PerformSubsystemEntry and PerformSubsystemEntry_TS should be very similiar, but they are not. How about instead of adding a verbose option to AssertOnAndClearRTError, you could pass in additional info? In that way, the only different betwewen PerformSubsystemEntry/PerformSubsystemEntry_TS would be that  additionally fed-in info.

cd1ddd508a (Changed local functions to static in some modules, 2026-08-04)

> Functions that were only called from testing were moved to
  UTF_HelperFunctions.ipf

IIRC we discussed that offline and decided not to do that or?

- Typo: fstatic
- No comments like `Unused function?`, these are (mostly) used with the ZeroMQ XOP
- `+///        Note: Keep function global as it is called from
  Packages/doc/ipf/MIES_Regenerate-Screenshots.ipf in CI` superfluous comment, everyone who can use `git grep` knows
  that.

5ad4ff3176 (Add PerformSubsystemEntry guards when public funcs of another module are called, 2026-08-04)

The commit message lacks the motivation. And it should also mention that the call is cheap.

Without TS suffix:

```
@@ -624,6 +628,9 @@ static Function AI_SetMIESHeadstage(string device, [variable headstage, variable
 Function AI_ZeroAmps(string device, [variable headStage])

     variable i
+
+    PerformSubsystemEntry_TS()
+
     // Ensure that data in BaselineSSAvg is up to date by verifying that TP is active
     if(IsDeviceActiveWithBGTask(device, "TestPulse") || IsDeviceActiveWithBGTask(device, "TestPulseMD"))

```

Rest looks good to me.

t-b (changes requested):
See comment

MichaelHuth:
> > Functions that were only called from testing were moved to
> > UTF_HelperFunctions.ipf
> 
> IIRC we discussed that offline and decided not to do that or?

Yes. This and the commit does not do that. I missed to remove that sentence in the commit message.

MichaelHuth:
> The two functions PerformSubsystemEntry and PerformSubsystemEntry_TS should be very similiar, but they are not. How about instead of adding a verbose option to AssertOnAndClearRTError, you could pass in additional info? In that way, the only different betwewen PerformSubsystemEntry/PerformSubsystemEntry_TS would be that additionally fed-in info.

I thought of that as well, but is not feasible because of the calling asymmetry for threadsafe functions. I can call threadsafe functions from the main thread but not the other way around. So when I call a function like AssertOnAndClearRTError which is threadsafe, I need to move the call to gather the MIES state before that as I need to distinguish between GetVerboseMIESState_TS and GetVerboseMIESState. This is possible, but has the major disadvantage that it is then not lightweight anymore because the call moves before determining if there is a pending RTE.

As this is not feasible, I end up with the same logic as AssertOnAndClearRTError already has.
Generally if I discard the verbose option and call GetVerboseMIESState_TS always in AssertOnAndClearRTError in the error case then it is already exactly what we need as PerformSubsystemEntry_TS.
The code for PerformSubsystemEntry is then the same but with GetVerboseMIESState, BUG instead.

What I noticed that is actually missing is a recursion protection. That is when GetVerboseMIESState / GetVerboseMIESState_TS call functions that call PerformSubsystemEntry, PerformSubsystemEntry_TS itself (like GetIgorInfo by accessing the Cache) then new errors in the recursion can lead to a stack overflow.

t-b (approved):

